PR #13, opened 2015-04-25
Clang complains about missing override declartions

ff2000:
```
src/gui/shell.h:21:8: fatal error: 'sizeHint' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
```

So make clang happy.

### DIFF
--- a/src/gui/shell.h
+++ b/src/gui/shell.h
@@ -18,9 +18,9 @@ public:
 	Shell(NeovimConnector *nvim, QWidget *parent=0);
 	~Shell();
 	QSize sizeIncrement() const;
-	QSize sizeHint() const;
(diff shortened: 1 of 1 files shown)


equalsraf:
I can't replicate with clang 3.5.0 but this sounds right in a more pedantic environment.

@ff2000 does replacing `override` with Q_DECL_OVERRIDE also works for you?

And just for completeness what OS/Clang version are you using?

ff2000:
I'm on Gentoo linux, clang-2.6.0.
And yes, Q_DECL_OVERRIDE also works. I just updated the PR.

equalsraf:
Merged thanks @ff2000.

Issue #8, opened 2015-03-31
Changing Neovim font


equalsraf:
A font changing dialog would be nice, but how do we trigger without risking shadowing a neovim mapping.

Otherwise if neovim/neovim#2203 is implemented we can wait for a notification for a predefined variable.

aikeru:
If we're stuck with just one font, could you switch it to be a powerline font?
https://github.com/powerline/fonts/tree/master/DejaVuSansMono

Or even better, the font with the file icons, too, for those that want them:
https://github.com/ryanoasis/font-nerd-icons/tree/master/patched-fonts/DejaVuSansMono

That would be really sweet.

equalsraf:
@aikeru, PR #16 seems ready to be merged, and should allow you to change your fonts.

But requires the following nvimrc snippet.
```
" Neovim-qt Guifont command, to change the font
command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "<args>")
" Set font on start
let g:Guifont="DejaVu Sans Mono for Powerline:h13"
```

yngwin:
neovim-qt should not ship a copy of dejavu, but should simply use the default monospace font as defined by fontconfig, in case no Guifont is set.

equalsraf:
@yngwin yup the bundled DejaVu will be removed when #16 is merged. It just defaults to Monospace.

vityafx:
I still don't understand what should I do to change font in neovim-qt, can you guys please enlight me?

When I do

    :Guifont Monaco:h12

Neovim says:

> E116: Invalid arguments for function rpcnotify

equalsraf:
Hi @vityafx.

I'm assuming you are using the branch from #16.

> > E116: Invalid arguments for function rpcnotify

It seems Neovim is not happy about the Guifont command, what do you have in your .nvimrc? Mine looks like this.

    command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "\<args\>") | let g:Guifont="<args>"

vityafx:
Yes, had a problems with it while pasting from here.
I've changed it to your one and it does not throw any error anymore. But the font still does not change.

equalsraf:
Strange did you build neovim-qt from the branch in #16? Also does `:messages` show any messages about the font?

vityafx:
`:messages` say:

> Unknown font: <args>

equalsraf:
[...]

PR #16, opened 2015-05-07
set GUI font

equalsraf:
For #8.

You need to add this command to your nvimrc for this to work
```
:command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "<args>") | let g:Guifont="<args>"
```

The command does two things

1. It sends an rpc notification "Gui" "SetFont" "Font"
2. It sets the global variable g:Guifont (this is necessary for cases when Guifont is called but the is GUI is not attached, i.e. nvimrc)

To see the current font

```
:Guifont
DejaVu Sans Mono:h11
```

To set the font
```
:Guifont Monaco:h12
```


equalsraf:
ping @coolwanglu, @carlosdcastillo, @qvacua, @rogual, @stefan991, @oakes (did I miss anyone?)

Neovim notifications seem like a good way to replace the old `set guifont` it would be cool if we can agree on semantics and share the same VimL bits across all GUIs. Maybe start a PR on neovim/neovim for the VimL. Or maybe one of you already has this working?

PS: I've used the `set guifont` Windows semantics for font names, but just because I prefer those to the space escaping in Unix.

rogual:
I'm looking at using `rpcnotify` in NeoVim.app (OSX) and it seems like the way to go; `rpcrequest` supports returning values and flagging errors but I couldn't figure out if there's a way to use it for a GUI or not; its synchronous nature seems to cause it to hang the program.

In my experimental branch I'm using the `Mac` prefix, i.e. `MacSetFont`; a cross-GUI `Gui` prefix sounds like a great idea to me.

equalsraf:
For errors I'm just printing with `vim_report_error("Invalid font height")` or `vim_out_write(...)`

coolwanglu:
I'm not sure if this should be made central at neovim. Some believe that it's the specific GUI's response to manage this.
My concern of a centralized font settings is that, the format may vary among platforms, e.g. neovim-e uses the format of CSS, which should be different from other native platforms.

equalsraf:
@coolwanglu I did not mean the actual font format, just the commands the user calls to set the font and the semantics for font names i.e. the equivalent to the old `set guifont=`

coolwanglu:
@equalsraf yes, then you will need a format for font family and size for that option?
Well anyway in `options.txt` I saw that it's already platform dependent.

I'm not sure if we need a new command for that. It sounds simpler and more powerful to me that if GUI clients would be able to subscribe to some options.

equalsraf:
> clients would be able to subscribe to some options

I don't think option subscription is working yet, but that would work too, just need to know which option and what is the format for the value - for set guifont there were some semantics differences between Unix/Windows/Mac, but we can choose one of the already existing ones.

Mariappan:
hi @equalsraf can you please rebase this with the master branch.
Will use this branch build, with the latest tcp connection close fix also :stuck_out_tongue:

equalsraf:
Rebased.

Mariappan:
Thanks @equalsraf 
I am facing a small issue with the font rendering. The below picture explains the prob.
First window is the vim-qt and second one is neovim-qt

nvim-qtfonterror
In the status line, you can see the difference clearly

Mariappan:
One more update, with monospace font I am facing this problem

dani-h:
Can't seem to get this working. No error message either, the font just doesn't change

equalsraf:
@dani-h this only works in the tb-setguifont branch. What fonts did you try with? Does `:Guifont` print anything at all?

equalsraf:
@Mariappan where did you get that Consolas font, can you send me a link so I can test?

equalsraf:
@dani-h added a fix that may help, some error cases were not printing any error message.

dani-h:
Yeah the error message helps: `Font needs to be fixed pitch` for Monaco:h12. Works perfectly for GVim and in the terminal.

Also something slightly annoying is that you can't paste text into the bottom command prompt. <C-v> and <Insert> simply write those characters into the prompt instead of pasting text.

equalsraf:
> Yeah the error message helps: Font needs to be fixed pitch for Monaco:h12. Works perfectly for GVim and in the terminal.

I'm a bit more pedantic about fonts than GVim :D. GVim does allow you do use non Monospace fonts, some "Monospace" fonts are just a little buggy concerning metrics - but I prefer to throw in an early error to let the user know something is wrong with the font.

Where did you get your Monaco? This one works for me https://github.com/cstrap/monaco-font

equalsraf:
> Also something slightly annoying is that you can't paste text into the bottom command prompt. and simply write those characters into the prompt instead of pasting text.

You mean pasting into the vim command line, like [\<C-r\>"](http://stackoverflow.com/questions/3997078/how-to-paste-yanked-text-into-vim-command-line)

dani-h:
Yeah, I usually just use shift-insert

equalsraf:
@dani-h Shift-Insert is handled by your terminal, not by Vim, if you try it in GVim you will get \<S-Insert\> as output too.

dani-h:
> I'm a bit more pedantic about fonts than GVim :D. GVim does allow you do use non Monospace fonts, some "Monospace" fonts are just a little buggy concerning metrics - but I prefer to throw in an early error to let the user know something is wrong with the font.

Where did you get your Monaco? This one works for me https://github.com/cstrap/monaco-font

I'm pretty sure I used that installation script in that repo (for Ubuntu) in the past and I installed it again for good measure, the error message was different this time: `Unknown font`

dani-h:
> @dani-h Shift-Insert is handled by your terminal, not by Vim, if you try it in GVim you will get <S-Insert> as output too.

Nope, \<S-Insert> for in Gvim, tested again now

equalsraf:
> Nope, <S-Insert> for in Gvim, tested again now

Strange maybe the mapping is set somewhere but my GVim does not have it (also [here](http://superuser.com/questions/322947/gvim-shift-insert-dump-s-insert-instead-of-the-clipboard-text)) it should be just a mapping like

    cmap <S-Insert>  <C-R>+

But this requires something like `has("gui_running")` and I dont think it works ATM.

> I'm pretty sure I used that installation script in that repo (for Ubuntu) in the past and I installed it again for good measure, the error message was different this time: Unknown font

So it does not find the font at all now. Do other applications see it?

dani-h:
> So it does not find the font at all now. Do other applications see it?

Yes. It's my default font for the terminal

dani-h:
Is it possible to define and call the Guifont command in a .vimrc file? It seems like when I try to do that the font doesn't change at startup. So each time I open nvim-qt I have to change the font.

`
  command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "<args>")
  Guifont Monaco:h10

`

equalsraf:
> Is it possible to define and call the Guifont command in a .vimrc file?

Because when nvimrc is processed the GUI is not attached yet. This is what I use in my nvimrc right now.

    " Neovim-qt Guifont command
    command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "<args>")
    let g:Guifont="DejaVu Sans Mono:h13"

The global variable is not ideal but until Neovim supports options for this and notifications I think this is the only way to do it.

dani-h:
Hah you are really fast to reply :+1: It works, thanks.

The font in nvim-qt isn't as clear as in the terminal, do you have any clue why? Seems a little blurry

equalsraf:
Maybe its related to #39? Are you using an HiDPI display?

dani-h:
I'm not sure but I don't think I'm using a HiDPI display. I'll double check later

It seems like I have the same issue with gnome-terminal when rendering the Monaco font. It seems a little wider than it's supposed to be.

Here's a comparison of lxterminal and neovim-qt
neovim-qt-bug

equalsraf:
Mmmm the lack of italic variant is also causing some issues.

bug1

equalsraf:
@dani-h I'll probably break Monaco for you later today :S.

Any idea if its possible to get Monaco with bold and italic?

Mariappan:
@equalsraf I am using the font from the below path:
https://github.com/runsisi/consolas-font-for-powerline/blob/master/Powerline%20Consolas.ttf

mkusher:
First one is neovim-qt and second one is gvim
snapshot422
snapshot421

equalsraf:
@mkusher what font/fontsize are you using in those screenshots?

mkusher:
@equalsraf I'm using patched Liberation Mono with https://github.com/ryanoasis/vim-devicons

equalsraf:
@mkusher, confirmed here. Those characters seem to be double width but are rendered  in a single cell. So either we are (mistakenly) identifying those as single width, or the font is broken. 

I'm also seeing some issues in my terminal (konsole) when I place the cursor on top of the gear icon - the chars spill outside of their cell and the text is pushed to the right. Any idea if this happens in other terminals too?

FIY ryanoasis/vim-devicons#11 mentions the ability to generate single width glyphs

mkusher:
@equalsraf gVim just highlight char part inside the cursor
snapshot425
but it looks like termite have managed to do highlighting the right way
snapshot427
snapshot426

Mariappan:
hi @equalsraf I am not able to open nvimqt (windows 7) from the latest app veyor.
I am getting the following error.
image

raghur:
I have the same issue as @Mariappan when trying the build from appveyor (#327). build no. 325 works well though (though you can't change Guifont on that yet)

equalsraf:
Quick look, apparently I tripped over a NULL pointer in there

```
Unhandled exception: page fault on read access to 0x00000000 in 32-bit code (0x100a7500).
Register dump:
 CS:0023 SS:002b DS:002b ES:002b FS:0063 GS:006b
 EIP:100a7500 ESP:0033fcc8 EBP:0033fcdc EFLAGS:00010206(  R- --  I   - -P- )
 EAX:0033fd28 EBX:005d6368 ECX:00000000 EDX:005d7338
 ESI:005d6318 EDI:005d63f0
Stack dump:
0x0033fcc8:  0040fea3 005d63f0 005d6318 005d6318
0x0033fcd8:  00000009 0033fcf4 0040fb78 005d6318
0x0033fce8:  0040feec 005d63f0 6707f520 0033fd40
0x0033fcf8:  0040b1ca 0033fd28 00000006 1866f014
0x0033fd08:  005efba0 005d4d88 00136300 0033fd44
0x0033fd18:  6504167f 005f0838 6504168c 005c6e30
Backtrace:
=>0 0x100a7500 ?height@QFontMetrics@@QBEHXZ() in qt5gui (0x0033fcdc)
  1 0x0040fb78 NeovimQt::Shell::neovimHeight+0x7() [c:\projects\neovim-qt\src\gui\shell.cpp:150] in nvim-qt (0x0033fcf4)
  2 0x0040fb78 NeovimQt::Shell::neovimHeight+0x7() [c:\projects\neovim-qt\src\gui\shell.cpp:150] in nvim-qt (0x0033fcf4)
  3 0x0040fb78 NeovimQt::Shell::neovimHeight+0x7() [c:\projects\neovim-qt\src\gui\shell.cpp:150] in nvim-qt (0x0033fcf4)
  4 0x0040fb78 NeovimQt::Shell::neovimHeight+0x7() [c:\projects\neovim-qt\src\gui\shell.cpp:150] in nvim-qt (0x0033fcf4)
  5 0x0040fb78 NeovimQt::Shell::neovimHeight+0x7() [c:\projects\neovim-qt\src\gui\shell.cpp:150] in nvim-qt (0x0033fcf4)
  6 0x0040fb78 NeovimQt::Shell::neovimHeight+0x7() [c:\projects\neovim-qt\src\gui\shell.cpp:150] in nvim-qt (0x0033fcf4)
```

equalsraf:
Pushed a fix it should be up by now, let me know how it goes.

Mariappan:
hi @equalsraf thanks for the quick resolution. But after updating none of my fonts are loading.
Fixedsys
monofur for Powerline
Powerline Consolas
Consolas

What font are you using?? Any powerline fonts??

Mariappan:
The default font loaded is CourierNew:h8

I tried to increase the font size. That also is not working

equalsraf:
I cant test on Windows right now. It is refusing to load the font because it can't find italic or bold variants for that font. Its a bit pedantic but the font must

- be fixed with
- have italic/bold/bold+italic variants
- the metrics for all variants must match

raghur:
the metrics and variants requirements seem to be eliminating pretty much all the fonts I have. Input, Hack and Pragmata aren't working and in each case I get "<font> has non fixed metrics for its variants"

equalsraf:
> the metrics and variants requirements seem to be eliminating pretty much all the fonts I have. Input, Hack and Pragmata aren't working and in each case I get " has non fixed metrics for its variants"

Do these fonts have bold/italics installed or just the regular font?

raghur:
For both Pragmata and Input I have bold/italics installed. For Hack I only have regular and bold. I'd much prefer if the font was used but a warning was shown that the metrics are not fixed.

equalsraf:
@raghur good point, I've pushed a commit.

- Apparently QFontMetrics in Windows is reporting slightly weird maxWidth metrics for all fonts anyway, the averageWidth is actually ok.
- The error message might be too annoying, I don't think it can block anything, but let me know what you think

aikeru:
If this branch is up-to-date but is yet to be merged due to testing effort, is there any chance I could get a link for a binary? :) That or would you be willing to share one somehow? Thanks for making this awesome project!

equalsraf:
@aikeru if you are in windows you can get the binaries from appveyor https://ci.appveyor.com/project/equalsraf/neovim-qt/branch/tb-setguifont - press the "Artifacts" tab and there is a zip there that you can download. There are also nupkg build available.

expipiplus1:
This is working well for me, although when one tries to use a missing font (for example by running `Guifont foo` I get the error message `Unknown font : <args>` instead of `Unknown font: foo`

equalsraf:
@expipiplus1 that is strange, what do you have in nvimrc for Guifont?

expipiplus1:
A little bit more info:
I'm running HEAD of neovim and HEAD of neovim-qt on Windows.
The line I have in my .nvimrc is `    command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "\<args\>") | let g:Guifont="<args>"`

equalsraf:
Do you actually have the slashes in there? Should be `<args>` instead of `\<args\>`, or maybe it was typo from the markdown? Mine looks like this

    command -nargs=? Guifont call rpcnotify(0, 'Gui', 'SetFont', "<args>") | let g:Guifont="<args>"

expipiplus1:
Ah, that's it. I removed the backslashes and everything's fine.
I copied that line from one of the font threads on neovim-qt.

Issue #4, opened 2014-12-31
Window resize is slightly broken


equalsraf:
(in the tb-remote-ui branch) resizing the GUI may have some unexpected effects e.g.

1. We drag the window to resize and afterwards the window is resized back
1. We resize the window, but the actual Neovim work area remains the same size, with a wide margin around it

There are multiple reasons for this issue

1. try_resize_ui does not guarantee a resize and SHOULD NOT be called simultaneously - Neovim will ignore paralell calls
1. When the user is dragging the window it will cause multiple calls to try_resize_ui, emphasizing the previous point
1. resizing can be expensive in drawing operations (i.e. takes time)

There are ways around these issues, but I would prefer to wait until the Neovim API becomes a bit more stable and the UI functions are place in the API metadata object (so the bindings generate signals for those calls as well). Preliminary suggestions

1. Collapse multiple calls to try_resize_ui and only make a single call at a time - this is similar to what was done for vim-qt
1. Delay repainting due to window resizing.

equalsraf:
Have an ongoing fix for this, but I'm having some issues with this https://github.com/equalsraf/neovim-qt/issues/4

equalsraf:
[QTBUG-45086](https://bugreports.qt.io/browse/QTBUG-45806)

equalsraf:
#28 was a big help but caused an issue when using i3 as window manager. Because we try to attach the Neovim window as 2/3 of the available screen space - but the tilled window manager does not allow the window to be resized - this causes a margin to appear around the Neovim widget.

![bug1](https://cloud.githubusercontent.com/assets/715238/8633120/cca6b67a-27b1-11e5-9ebb-e8592e336923.png)

The margin background is QTBUG-45086. But the actual size issue can be fixed. We need to treat that default size a bit more carefully, e.g.

```diff

--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -147,8 +147,7 @@ void Shell::neovimIsReady()
        connect(m_nvim->neovimObject(), &Neovim::neovimNotification,
                        this, &Shell::handleNeovimNotification);
        // FIXME: this API will change
-       QRect screenRect = QApplication::desktop()->availableGeometry(this);
-       m_nvim->attachUi(screenRect.width()*0.66/neovimCellWidth(), screenRect.height()*0.66/neovimRowHeight());
+       m_nvim->attachUi(width()/neovimCellWidth(), height()/neovimRowHeight());
 
[...]

PR #28, opened 2015-06-25
Fix window resize issues

### LINKED ISSUE #21: Window startup size
When i startup the window is quite small, about 210x130 px. Is there a way to configure the startup size?

The greates problem is that i cannot see the error messages, just some gibberish pixels:

This is also the startup size.

equalsraf:
Fix #21 and #4.

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -34,6 +34,8 @@ void MainWindow::init(NeovimConnector *c)
 			this, &MainWindow::neovimError);
 	connect(m_errorWidget, &ErrorWidget::reconnectNeovim,
 			this, &MainWindow::reconnectNeovim);
+	connect(m_shell, &Shell::neovimResized,
+			this, &MainWindow::neovimWidgetResized);
 	m_shell->setFocus(Qt::OtherFocusReason);
 
 	if (m_nvim->errorCause()) {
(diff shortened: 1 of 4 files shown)


prabirshrestha:
While this is definitely lot better then what it used to be. I find it too big now.

image

I think this is due to the position. The height does seem to be the height of the screen including taskbar which is usually what we don't want since most of us might not have set the taskbar to autohide.

image

Width also seems to be the width of the screen if so I think the best position is 0,0. Or just start it in maximized mode and then remember the size and position when I re-open it later.

equalsraf:
Agreed, the size difference should be the width/height of the window decorations (not the taskbar) but the window size is defined at attachment, so we can:

- as you said start maximized, but this is probably unintuitive for most people
- pick an arbitrary default size that sounds reasonable e.g. 2/3 width and height of the current screen
- if we find a way to measure window decoration sizes we can keep the current behavior, but this is trickier to get right ([like this](https://github.com/equalsraf/vim-qt/blob/master/src/gui_qt.cpp#L619)) because menubars/toolbars and internal widget styles take up space as well

equalsraf:
Went with the easier option :smile:, just defined the window size as 2/3 of the screen size.

prabirshrestha:
:+1: 2/3 seems reasonable. I don't like maximized state or small size like what default gvim does.

PR #30, opened 2015-07-27
Use new Neovim mode change API events

equalsraf:
The Neovim API changed, now uses a single event for mode change
notifications. Remove the old functions and replace them with the
new one.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -401,10 +401,13 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs, QPa
 		this->unsetCursor();
 	} else if (name == "mouse_off"){
 		this->setCursor(Qt::BlankCursor);
-	} else if (name == "normal_mode"){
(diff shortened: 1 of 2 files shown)


src/gui/shell.cpp
```diff
@@ -426,14 +429,14 @@ void Shell::setNeovimCursor(quint64 row, quint64 col)
...
+void Shell::handleModeChange(const QString& mode)
 {
-	m_insertMode = false;
-}
-
-void Shell::handleInsertMode(QPainter& painter)
-{
-	m_insertMode = true;
+	// TODO: Implement visual aids for other modes
+	if (mode == "insert") {
```
justinmk: should this be `mode == "insert" || mode == "replace"` ?

equalsraf: Not yet, I've just changed the minimum amount to keep showing the thin cursor in insert mode. For replace mode (at least in gvim) the mode cursor looks different (like a thick underline).

Issue #36, opened 2015-08-24
Closing nvim-qt closes the headless nvim server running in TCP socket


Mariappan:
Hi
I have started neovim in headless mode in remote server and connect from local laptop using tcp/ip msgpack.

Everything is working fine. But when i close the neovim-qt (which i will do accidently most of the time), it closes the nvim server which is running in the server. I am expecting only the connection get closed so that I can connect to the same nvim server after sometime.

Can we change the behavior of close button to just disconnect the session without actually closing the nvim running in headless mode.

equalsraf:
Maybe for TCP sessions, but certainly not for embedded Neovim processes.

The `:qa` command is issued when a [closeEvent](https://github.com/equalsraf/neovim-qt/blob/master/src/gui/mainwindow.cpp#L92) takes place. The type of connection is available through the [m_ctype](https://github.com/equalsraf/neovim-qt/blob/master/src/neovimconnector.h#L37) attribute (currently private).

My question here is, what would the default behavior?

1. If Neovim is launched via `spawn()` the current behavior remains unchanged
1. If its a TCP connection then skip the call to `:qa` as you suggest
1. If its a local socket/pipe (same as the TCP case for consistency)

Not ideal if the user actually expects the window close button to close a remote Neovim, but I currently don't have a way to customize this behavior. If the user really needs to exit he can call `:q`

A simple approach would be to expose `m_ctype` and add an if statement to [Shell::closeEvent](https://github.com/equalsraf/neovim-qt/blob/master/src/gui/shell.cpp#L637)

Mariappan:
> If Neovim is launched via spawn() the current behavior remains unchanged
> If its a TCP connection then skip the call to :qa as you suggest
> If its a local socket/pipe (same as the TCP case for consistency)

Thats what i want exactly. You summarized it crystal clearly.

> Not ideal if the user actually expects the window close button to close a remote Neovim, but I currently don't have a way to customize this behavior.

I dont anticipate this will be a ideal case, since if the user starts nvim in socket or tcp connection (probably in server), he wants it to be alive just like any other server. If user wants to edit only local file, he/she could open with embed option which is much more convenient than opening the GUI specifying nvim listen address.

Last thing, thanks a lot for this wonderful GUI :+1: . Already started using it in my development setup with nvim in headless mode in server and nvim-qt acting as both ssh connection and tmux session :smile: .

Mariappan:
Closing this Issue as the code changes are merged.

PR #38, opened 2015-09-04
GUI: Don't call :qa on close for UNIX/TCP sockets

equalsraf:
When the Neovim shell is closed (e.g. user closes window) the GUI would ignore
the event and call :qa, triggering the Neovim process to exit. This makes sense
when Neovim was launched by the GUI, since we need to cleanup the process.
However when connecting to an external Neovim process (TCP/UNIX socket) this
might not be what the user intends.

For TCP/UNIX sockets the "qa" command is no longer sent. This means that if the
user closes the GUI window the GUI will terminate the connection and close, but
the Neovim process will remain runnning. In cases where the user wants to exit
the Neovim process he would need to call :q himself.

This breaks the previous behaviour, but there was no way to override it. From a
user perspective the situation might improve later for UNIX sockets if the GUI
can discover available sockets.

Issue #36

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -632,7 +632,10 @@ void Shell::changeEvent( QEvent *ev)
 
 void Shell::closeEvent(QCloseEvent *ev)
 {
-	if (m_attached) {
+	if (m_attached &&
+		m_nvim->connectionType() == NeovimConnector::SpawnedConnection) {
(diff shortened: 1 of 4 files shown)


equalsraf:
Fixed the semaphore builds.

@Mariappan this should do what you wanted.

Mariappan:
Thanks @equalsraf . Downloaded appveyor binary.
Working as expected :+1: :smile:

PR #40, opened 2015-09-11
Support mouse events

equalsraf:


equalsraf:
Don't have a reliable way to write UI unit tests yet. Otherwise this seems to work ok.

Issue #43, opened 2015-09-21
Add support for SpellBad and SpellCap highlighting


mkusher:
I mean thing like:
![snapshot424](https://cloud.githubusercontent.com/assets/523021/10005823/2863e9c6-60c2-11e5-9bb6-0a00b68976ba.png)

equalsraf:
@mkusher I got started on the undercurl drawing. Its not clear to me at this point what Neovim is doing for the SpellCap case. I assumed the color would be different but I'll need to have a closer look.

mkusher:
@equalsraf you're pretty fast :))

equalsraf:
This should be fixed with #121

PR #44, opened 2015-09-21
WIP: Support undercurl attribute

equalsraf:
When drawing text the undercurl attribute was being represented as an
underline, and the underline attribute was ignored - this commit
corrects the use of underline and adds initial support for undercurl.

The undercurl is drawn over the text as a red dotted line, at the
bottom of the text cell.

 For #43.

Caveats:
- currently the color is hardcoded
- its not actually an undercurl :D
- Quick reference to [old issues](https://bitbucket.org/equalsraf/vim-qt/issues/59/undercurl-underline-or-even-underscore-not) we want to avoid.


equalsraf:
Merged the temporary fix into master, closing this until the upstream issue is fixed #43 will remain open to track this issue.

PR #45, opened 2015-09-22
Support mouse scroll events

equalsraf:
Vertical scroll seems to work here, but I can't test horizontal scroll. It would also be nice to test with mice that have unusual resolutions to make sure I'm doing this right (might be too sensitive right now)

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -676,6 +676,34 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
 	}
 }
 
+void Shell::wheelEvent(QWheelEvent *ev)
+{
+	int horiz, vert;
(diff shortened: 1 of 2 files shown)


equalsraf:
Quick test shows that horizontal scroll events are being processed, but they are are ignored.

equalsraf:
Easy way to test in X11

    xte 'sleep 2' 'mouseclick 6'

The horizontal scroll events are being handled and passed to nvim, But nothing happens - the same is happening in gVim, so maybe its just an error on my part (nowrap is on). For example if I set a mapping to

    nmap <ScrollWheelRight> l

it actually works, so maybe there is no default behaviour for the horizontal scroll (or maybe I'm misreading the docs). In any case it seems good on our end.

Issue #50, opened 2015-09-30
Can neovim-qt connect to nvim using stdio?


tarruda:
Hi

I'm planning to make some changes in the "builtin UI" infrastructure for the next month(after 0.1 release). My goal is to make every UI an external program that is spawned by nvim at startup so projects like neovim-qt are basically on the same layer as the builtin TUI.

The idea is to have nvim spawn the UI program and communicate to it via stdin/stdout, so the UI program doesn't have to worry about connecting to neovim.

This also lets one use the nvim binary to spawn any UI, all nvim command line options will be readily available to neovim-qt. For example, here's an idea of how neovim-qt could be used to edit files from stdin(the exact mechanism used to select which UI will be spawned is still going to be determined):

```
echo "file contents"  | nvim -u NONE --ui=neovim-qt -
```

I'm also preparing to implement nvim "remote mode" where nvim will be able to connect to other instances, so no UIs have to worry about connectivity beyond stdin/stdout.

Can you prepare neovim-qt to work in this scenario? I think this will only simplify things since it will no longer be necessary to worry about sockets or named pipes.

equalsraf:
Hi! Got a PR going for this, seems to work with

    :rpcstart('nvim-qt', ['--embed'])

Is that what you had in mind?

> This also lets one use the nvim binary to spawn any UI, all nvim command line options will be readily available to neovim-qt.

Yes, also makes sense to avoid some of the issues we have seen: e.g. Neovim starts and blocks on some error waiting for a keypress, then the UI connects and nothing happens.

An issue that comes to mind is this - if Neovim fails to start (on an early error) or if it can't find or launch the UI process. it will be impossible to report errors to the user.

> mechanism used to select which UI will be spawned is still going to be determined):

It might be enough to have `nvim --ui` and pick an option from nvimrc with the UI binary path - this way Neovim doesn't need to know any specifics about the UI. In Vim there was a `-g` option to run the GUI instead of the terminal version, so we can even be backwards compatible and use that.

tarruda:
> Is that what you had in mind?

Yes, but neovim will do it automatically at startup unless --embed or --headless is passed

> An issue that comes to mind is this - if Neovim fails to start (on an early error) or if it can't find or launch the UI process. it will be impossible to report errors to the user.

This shouldn't happen in standard situations, we'll always distribute and install an UI with neovim. In any case we can always print a warning to stdout and start in headless mode as a fallback.

[...]

PR #51, opened 2015-09-30
Communicate with Neovim using stdin/stdout

equalsraf:
For #50. Right now you can

    :call rpcstart('nvim-qt', ['--embed'])

I have not decided on the internals yet, but this seems to be a working start - there are probably lots of pitfalls I'm missing about reading from stdin.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,11 @@ foreach(CFGNAME ${CMAKE_CONFIGURATION_TYPES})
   set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY_${CFGNAME} ${CMAKE_BINARY_DIR}/lib)
 endforeach()
 
+if(MSVC)
+  # Allow use of deprecated function names in MSVC (read/write)
+  add_definitions(-D_CRT_NONSTDC_NO_DEPRECATE)
+endif()
+
 add_subdirectory(src)
 add_subdirectory(doc)
 enable_testing()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,5 +1,9 @@
 
 set(NEOVIM_QT_SOURCES util.cpp neovimconnector.cpp neovimconnectorhelper.cpp function.cpp msgpackrequest.cpp auto/neovim.cpp msgpackiodevice.cpp)
+if(WIN32)
+  list(APPEND NEOVIM_QT_SOURCES stdinreader.cpp)
+endif()
+
 add_library(neovim-qt STATIC ${NEOVIM_QT_SOURCES})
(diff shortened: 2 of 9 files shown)


equalsraf:
@tarruda quick note when doing this via rpcstart in Windows [UV_PROCESS_WINDOWS_HIDE](https://github.com/neovim/neovim/blob/dc9652e68de163290abee880a74bf1727c715a1e/src/nvim/event/libuv_process.c#L22) causes the Window to be invisible. The flag makes sense when running commands in background, just not for this particular scenario.

equalsraf:
Broken on Windows, it seems QAbstractSocket::setSocketDescriptor does not work.

tarruda:
> @tarruda quick note when doing this via rpcstart in Windows UV_PROCESS_WINDOWS_HIDE causes the Window to be invisible. The flag makes sense when running commands in background, just not for this particular scenario

:+1: 

> Broken on Windows, it seems QAbstractSocket::setSocketDescriptor does not work.

It seems `select()` only works with sockets on Windows, for other types of handles you need to use IOCP. Perhaps Qt uses `select()` on the arguments passed to `setSocketDescriptor`?

equalsraf:
> It seems select() only works with sockets on Windows, for other types of handles you need to use IOCP. Perhaps Qt uses select() on the arguments passed to setSocketDescriptor?

Seems to be the case, the Qt classes are unable to do asynchronous IO over stdin in Windows. Most of what I've seen so far uses either a background thread to read stdin or a timed poll.

suy:
In case it's useful, [there was recently a thread about a similar issue with QProcess](http://lists.qt-project.org/pipermail/interest/2015-April/016312.html). Some maybe useful internal details were mentioned.

equalsraf:
Thanks @suy, related bug [QTBUG-15457](https://bugreports.qt.io/browse/QTBUG-15457).

tarruda:
@equalsraf one workaround is to integrate Qt event loop with libuv and use it to poll for stdin. AFAIK libuv uses IOCP in windows and should have no problem with this scenario. I don't know how well libuv can integrate with Qt, but [this SO answer](http://stackoverflow.com/a/25825311) seems to contain some ideas.

equalsraf:
@tarruda I have a working PoC that just reads stdin in a separate thread and passes data around. The right way *TM* would be to integrate the event loop, or [implementing IOCP for Qt](https://bugreports.qt.io/browse/QTBUG-8978), there is also an [issue to support the libuv loop](https://bugreports.qt.io/browse/QTBUG-19944)

equalsraf:
Merged as is, not as battle tested as I would like but this should get us started, and this PR is useful for #61.

Issue #64, opened 2015-11-06
v:windowid is not defined


jalvesaq:
The value of `v:windowid` is not properly defined on Windows. It should be the number corresponding to Neovim-Qt window handle. I am not sure if `GetActiveWindow()` is the correct function to get the window handle in this case, but once you get it you can cast it to `long` (32 bit) or `long long` (64 bit).

On Linux, nvim running in a terminal emulator does not set the value of `v:windowid`, but we can use the value of `$WINDOWID`.

equalsraf:
I can't fiddle with v:windowid from the GUI since it is read only (this deserves an issue request in neovim/neovim), but I think I can provide an alternative workaround for windows, by setting the $WINDOWID inside the Neovim process when the GUI attaches.

jalvesaq:
Thanks! `$WINDOWID` is going to solve my problem.

equalsraf:
`g:GuiWindowId`now holds the window id.

justinmk:
> I can't fiddle with v:windowid from the GUI since it is read only (this deserves an issue request in neovim/neovim)

Oh, didn't realize that--let's fix that.

PR #65, opened 2015-11-08
Windows: Set $WINDOWID in Neovim process

equalsraf:
From issue #64 - in Neovim the v:windowid var is not available, while in Unix
systems a workaround the issue is to use the $WINDOWID variable, in Windows
there is no alternative. This commit sets the $WINDOWID environment var inside
the attached Neovim process using a call to vim_command(let $WINDOWID="...").

Still need to clarify some bits:

- [ ] unset the variable on detach
- [ ] This has not been tested for correctness, the Qt docs warn about corner cases such as the winId changing - and there were some recent bugs over the Qt API, maybe we need to use the native Win32 API. Also casts might be needed.
- [ ] Should this be set when connecting to remote instances?

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -247,6 +247,23 @@ void Shell::init()
 
 	// Subscribe to GUI events
 	m_nvim->neovimObject()->vim_subscribe("Gui");
+
+	updateWindowId();
+}
+
+// In Win32, set $WINDOWID environment variable
+// in other systems this does nothing
+void Shell::updateWindowId()
+{
+#ifdef _WIN32
+	if (m_attached &&
+		m_nvim->connectionType() == NeovimConnector::SpawnedConnection) {
+		WId window_id = effectiveWinId();
+		m_nvim->neovimObject()->vim_command(
+				m_nvim->encode(QString("let $WINDOWID=\"%1\"").arg(window_id))
+				);
+	}
+#endif
 }
 
 void Shell::neovimError(NeovimConnector::NeovimError err)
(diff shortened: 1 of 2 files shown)


jalvesaq:
I tested the AppVeyor build "artifact", and it works as expected. I used the following C code to get Neovim window handle in an application running as Neovim job:

```c
#ifdef _WIN64
        NvimHwnd = (HWND)atoll(getenv("WINDOWID"));
#else
        NvimHwnd = (HWND)atol(getenv("WINDOWID"));
#endif
```

Regarding your uncertainty about when to (un)set the variable, here is my two cents: The variable should only be set locally to the window because it will be used by other applications to act on the window. I am using the variable to raise the Neovim window because Neovim did not implement the `foreground()` function yet.

equalsraf:
Ok thanks @jalvesaq.

> The variable should only be set locally to the window because it will be used by other applications to act on the window.

Right now its set inside the Neovim process - there is no other way around without changing Neovim. This could be a problematic if for example Neovim starts a new Neovim instance.

If the only use case is the foreground() call that one is actually easier to handle - I've pushed a commit that allows you to emulate foreground() with

    call rpcnotify(0, 'Gui', 'Foreground')

jalvesaq:
I also need $WINDOWID to change the size and position of Neovim window after R startup. So, please, keep it.

The "Foreground" did not work. I tested with the following command:

```vim
:sleep 3 |  call rpcnotify(0, 'Gui', 'Foreground')
```

Then, I changed the focus to another window. After the `sleep`, the icon of Neovim window starts blinking in the taskbar, but the window is not raised.

equalsraf:
@jalvesaq ok thanks for checking - will keep the windowid in place and leave the foreground to another day.

jalvesaq:
OK, thanks!
Perhaps, instead of (src/gui/shell.cpp, lines 595 and 596)
```c
activateWindow();
raise();
```
you could mix Windows API and Qt API and try
```c
SetForegroundWindow(effectiveWinId());
```

jalvesaq:
I now tested Vim and it also does not work with similar command:

```vim
sleep 3 | call foreground()
```
The result is exactly the same of that of Neovim: the icon blinks in the task bar, but the window is not raised.

However, the same `foreground()` function does raise the GVim window after a command is pasted into R window by the [Vim-R-plugin]. I don't know why it works in one case but not in the other.

[Vim-R-plugin]: https://github.com/jcfaria/Vim-R-plugin

equalsraf:
Closing in favor of #114

PR #66, opened 2015-12-03
Fix two regressions

Marenz:
I bisected this bug down to commit https://github.com/equalsraf/neovim-qt/commit/38644b9

Here is a .gif demonstrating the problem. It is a bit multisided. The gif demonstrate the following things in order: 

* Without mouse/keyboard input you just see white. It only renders something once it gets input.
* Moving the window back and forth is fine until it gets to the big screen, then it freezes
* But giving it a lot of input also causes it to freeze (last part)

*Edit: I removed the embedded gif and only post a link as it seems to crash some browsers / create huge memory usage*
https://cloud.githubusercontent.com/assets/424752/11551006/5787c8ee-9975-11e5-8137-e58a932d1bd2.gif

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -213,18 +213,29 @@ void Shell::neovimIsReady()
 		return;
 	}
 
+	init();
+
 	// Check g:Guifont for a font from user settings
 	MsgpackRequest *r = m_nvim->neovimObject()->vim_get_var("Guifont");
 	connect(r, &MsgpackRequest::finished,
 			this, &Shell::neovimFontVarOk);
 	connect(r, &MsgpackRequest::error,
-			this, &Shell::init);
+			this, &Shell::attach);
 }
 
 void Shell::neovimFontVarOk(quint32, Function::FunctionId, const QVariant& ret)
 {
 	setGuiFont(m_nvim->decode(ret.toByteArray()));
-	init();
+	attach();
+}
+
+void Shell::attach()
+{
+	// FIXME: Don't set this here, wait for return from ui_attach instead
+	setAttached(true);
+	// FIXME: this API will change
+	QRect screenRect = QApplication::desktop()->availableGeometry(this);
+	m_nvim->attachUi(screenRect.width()*0.66/neovimCellWidth(), screenRect.height()*0.66/neovimRowHeight());
 }
 
 /**
@@ -233,14 +244,8 @@ void Shell::neovimFontVarOk(quint32, Function::FunctionId, const QVariant& ret)
  */
 void Shell::init()
 {
-	// FIXME: Don't set this here, wait for return from ui_attach instead
-	setAttached(true);
-
 	connect(m_nvim->neovimObject(), &Neovim::neovimNotification,
 			this, &Shell::handleNeovimNotification);
-	// FIXME: this API will change
-	QRect screenRect = QApplication::desktop()->availableGeometry(this);
-	m_nvim->attachUi(screenRect.width()*0.66/neovimCellWidth(), screenRect.height()*0.66/neovimRowHeight());
 
 	connect(m_nvim->neovimObject(), &Neovim::on_ui_try_resize,
 			this, &Shell::neovimResizeFinished);
--- a/src/gui/shell.h
+++ b/src/gui/shell.h
@@ -44,6 +44,7 @@ protected slots:
 	void mouseClickReset();
 	void mouseClickIncrement(Qt::MouseButton bt);
         void init();
+	void attach();
 
 protected:
 	void tooltip(const QString& text);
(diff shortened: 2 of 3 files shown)


Marenz:
Ah, damn, I just noticed that 2d393d9 doesn't fix the problem at all. Have to look at that again. Tomorrow though.

equalsraf:
IIIRC, I really made a mess of things there, thanks for tracking this down.

By removing the `msgpack_unpacker_reserve_buffer` the buffer can never grow, once it fills up we just keep reading 0 bytes. Looking at the old version of `::dataAvailable()` I suspect it was also wrong in that it should check the buffer before attempting to read.

Does this work?

```diff
index 109a62d..246d71b 100644
--- a/src/msgpackiodevice.cpp
+++ b/src/msgpackiodevice.cpp
@@ -176,16 +176,28 @@ void MsgpackIODevice::dataAvailableFd(int fd)
  */
 void MsgpackIODevice::dataAvailable()
 {
-       qint64 bytes = m_dev->read(msgpack_unpacker_buffer(&m_uk), msgpack_unpacker_buffer_capacity(&m_uk));
-       if ( bytes > 0 ) {
-               msgpack_unpacker_buffer_consumed(&m_uk, bytes);
-               msgpack_unpacked result;
-               msgpack_unpacked_init(&result);
-               while(msgpack_unpacker_next(&m_uk, &result)) {
-                       dispatch(result.data);
+       qint64 bytes = 1;
+       while(bytes > 0) {
+               if (msgpack_unpacker_buffer_capacity(&m_uk) == 0) {
+                       if (!msgpack_unpacker_reserve_buffer(&m_uk, 8192)) {
+                               // FIXME: Error out
+                               qWarning() << "Could not allocate memory in unpack buffer";
+                               return;
+                       }
+               }
+
+               bytes = m_dev->read(msgpack_unpacker_buffer(&m_uk), msgpack_unpacker_buffer_capacity(&m_uk));
+
+               if ( bytes > 0 ) {
+                       msgpack_unpacker_buffer_consumed(&m_uk, bytes);
+                       msgpack_unpacked result;
+                       msgpack_unpacked_init(&result);
+                       while(msgpack_unpacker_next(&m_uk, &result)) {
+                               dispatch(result.data);
+                       }
+               } else if (bytes == -1) {
+                       setError(InvalidDevice, tr("Error when reading from device"));
                }
-       } else if (bytes == -1) {
-               setError(InvalidDevice, tr("Error when reading from device"));
        }
 }

```

BTW the best way to trigger the event here seems to be resizing the window multiple times, likely because this causes lots of msgpack notifications.

Marenz:
Yes, your patch works.

Any ideas about the second regression?

Marenz:
Your patch seems to be the same as my commit? Maybe you didn't see that this turned into a PR? ^^

equalsraf:
I did - I was fiddling with the code for a bit, and assumed the buffer check/size was different - but its the same.

> Any ideas about the second regression?

Cannot replicate the second one. Best way to debug is to either print some debug messages or hook a debugger to handleNeovimNotification. The notifications can only arrive after attachUi is called anyway, but maybe `setAttached(true)` being set too early is causing the empty image to be painted before it has any content - perhaps initializing the m_image to be white in the constructor fixes the issue.

Marenz:
Interesting. Filling m_image with a color kind of fixes it. I get a dirty black square for a second and then I see the expected content. It seems to be always black, no matter what color I specify.

equalsraf:
Maybe something like this?

```diff
diff --git a/src/gui/shell.cpp b/src/gui/shell.cpp
index 80ba1d9..155f1a6 100644
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -26,6 +26,7 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
        m_fm = new QFontMetrics(m_font);
 
        m_image = QImage(neovimSize(), QImage::Format_ARGB32_Premultiplied);
+       m_image.fill(m_background);
 
        setAttribute(Qt::WA_KeyCompression, false);
        setAttribute(Qt::WA_OpaquePaintEvent, true);
@@ -285,6 +286,7 @@ void Shell::handleResize(uint64_t cols, uint64_t rows)
 
        if (needs_resize) {
                QImage new_image = QImage(neovimSize(), QImage::Format_ARGB32_Premultiplied);
+               new_image.fill(m_background);
                {
                        // copy the old contents into the new canvas
                        QPainter painter(&new_image);
```

equalsraf:
@Marenz any objection if I push your first commit into master?

Marenz:
No go ahead. Sorry, had no time yet to try more of your suggestion

Marenz:
I seem to be unable to reproduce the other problem anymore.. not even with the original commit.

Marenz:
What I do get is a black square but it disappears after maybe 500ms. With your suggested diff above it becomes a white square.

equalsraf:
> I seem to be unable to reproduce the other problem anymore.. not even with the original commit.

The input issue? It could be an issue with Neovim upstream, did you update your neovim since then?

Marenz:
Hmm I might, can't say for sure anymore.

equalsraf:
Closing this, since the remaining commits no longer apply.

Issue #33, opened 2015-08-18
Allow mapping key sequences including key press and key release separately


mildred:
As we are not in a terminal any more, it should be possible to map key sequences using fine events like key press and key release.

If you are wondering about the use case, I have a vim extension that allows switching buffers in most recently used order. The goal is to mimic the Ctrl-Tab and Ctrl-Shift-Tab of others editors. Allowing to register an event on the Ctrl key released would allow me to register more appropriately when the buffer is used and have more reliable ordering of the buffers. Something similar to [this question on Stack Overflow: Distinguishing between the hold and release of a modifier key between two keystrokes in Vim](http://stackoverflow.com/questions/16564293/distinguishing-between-the-hold-and-release-of-a-modifier-key-between-two-keystr).

equalsraf:
While the GUI can catch those events, they are still sent to Neovim for processing and Neovim does not internally support semantics for hold/release events (just like Vim, the format has not changed). The best way to get this going is to request support in neovim/neovim.

equalsraf:
Closing, if needed reopen when upstream moves in on this.

sassanh:
@equalsraf key presses and key releases can be mapped to <Plug>(neovim-qt) like pressing `a` triggers `<Plug>(neovim-qt-pressed)a` and releasing `a` triggers `<Plug>(neovim-qt-released)a`. This way we can have this feature in neovim-qt without waiting for upstream.

sassanh:
Also there's a feature most terminals provide (I know iTerm and Konsole provide it) to let you map keys to key sequences in terminal layer, using it I was able to map `<c-i>` to something other than `tab` so that I can use it separately, using the method above maybe we can have a `<Plug>(neovim-qt-pressed-and-released)` (or whatever) to provide raw key events to end user so that he's able to use `<c-i>` and `<tab>` separately.

justinmk:
> able to use `<c-i>` and `<tab>` separately.

That should _already_ be possible with any GUI such as nvim-qt. The GUI should send `<c-i>` and `<tab>`, and each of those can be mapped individually.

justinmk:
Hmm, I just checked: `call nvim_input('<tab>')` invokes a mapping to `:nn <c-i>`. That should be fixed in Nvim.

sassanh:
[...]

PR #67, opened 2015-12-08
Added modifier keys to InputConv

mrtyormaa:
This pull request is to handle the modifier keys for neovim. Ref [issue-33](https://github.com/equalsraf/neovim-qt/issues/33)
There is a pending pull request in neovim (Ref [neovim-3802](https://github.com/neovim/neovim/pull/3802)), which enables it to handle the various modifier keys such as Ctrl, Alt, Meta and Shift. 
As QT is able to handle all the keyPress and keyRelease events, we just have to send the correct key-codes to neovim for it to work.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -45,6 +45,10 @@ InputConv::InputConv() {
 	specialKeys.insert(Qt::Key_Backslash, "Bslash");
 	specialKeys.insert(Qt::Key_Less, "lt");
 	specialKeys.insert(Qt::Key_Space, "Space");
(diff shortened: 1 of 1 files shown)


equalsraf:
What happened? I thought this was still waiting upstream.

mrtyormaa:
Yeah, the way the Modifier keys are being handled in neovim is slightly different than I had anticipated. When I send the Modifier Key commands as "Ctrl", "Shift" etc, they are being printed in the console rather than being handled, which is strange. Hence, I am closing it so that I can take another look into it and make the final changes appropriately. 

Do you have any thoughts on the same?

PR #70, opened 2015-12-13
Show the window only after nvim is initted with correct geometry

nowrep:
This is kind of hacky, but I think it is better if the window is displayed only after nvim is loaded, otherwise there is a small white window at the start only to be resized a moment later.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -2,6 +2,7 @@
 #include <QFontDatabase>
 #include <QtGlobal>
 #include <QFile>
+#include <QTimer>
 #include "neovimconnector.h"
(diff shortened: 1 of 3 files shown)


equalsraf:
How about some minor improvements 22aebb8e8eaf7b5799360416ff297da09c388469, we could likely get way with not having the timeout but it should not hurt either.

PS: I'm not using QTimer::singleShot simply because of compatibility (it was only added in 5.4), that was why the CI was failing.

nowrep:
+1 looks good. I would keep the timeout just to be sure, because if nvim hangs `neovimError` will be emitted after 10s (if i'm correct) and that's too long.

PR #71, opened 2015-12-13
Use custom layout instead of dock widgets

nowrep:
Adding dock widget to main window will add a separator between the
dock widget and the central widget.
This resulted in an extra spacing at the top of main window.

No more spacing at the top, error widget still works and looks the same.

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -1,15 +1,23 @@
 #include "mainwindow.h"
 
 #include <QCloseEvent>
+#include <QVBoxLayout>
 
 namespace NeovimQt {
 
 MainWindow::MainWindow(NeovimConnector *c, QWidget *parent)
-:QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0)
+:QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0), m_layout(0)
 {
 	m_errorWidget = new ErrorWidget();
-	newDockWidgetFor(m_errorWidget);
 	m_errorWidget->setVisible(false);
+
+	QWidget *w = new QWidget(this);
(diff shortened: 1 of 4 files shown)


equalsraf:
Cheers @nowrep. I'm having an issue with this, the window becomes invisible after init is called.

Cannot pinpoint the issue yet, but maybe the new widget or layout are being reparented and deleted causing the window to disappear.

nowrep:
Oh sorry, I was testing this on top of PR #66 and there it works fine.

equalsraf:
Ah ok, will merge that one first then.

nowrep:
But don't merge this as it is now, I didn't realize `MainWindow::init()` may be called more than once.

nowrep:
Fixed the issue now. The other way to fix it would be to call `QApplication::processEvents`, but I think this is cleaner.

equalsraf:
It shows now, but I'm seeing issues when resizing the window, sometimes the shell widget is not resized with the main window, e.g. the shell widget is larger than the window.

neovimqt

nowrep:
Hmm, weird .. works fine for me. What should I see on the screenshot? (cutoff tilde at the bottom?)

The only issue I'm seeing is sometimes it renders garbage on the resized part of screen, but I'm not sure if it is specific to this patch (will need to test it before).

In any case, I'll see if I can fix it somehow.

equalsraf:
> Hmm, weird .. works fine for me. What should I see on the screenshot? (cutoff tilde at the bottom?)

In my case it cuts off a big chunk of the widget contents.

> The only issue I'm seeing is sometimes it renders garbage on the resized part of screen, but I'm not sure if it is specific to this patch (will need to test it with master).

Not caused by this change, but likely a consequence of using adjustSize() - https://github.com/equalsraf/neovim-qt/issues/4#issuecomment-96081917

PR #75, opened 2015-12-14
Use custom layout instead of dock widgets

nowrep:
Sorry, accidentally removed the branch.

The only change is `adjustSize()` -> `resize(sizeHint())`.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -6,6 +6,7 @@
 #include <QDesktopWidget>
 #include <QApplication>
 #include <QKeyEvent>
+#include <QLayout>
 #include "msgpackrequest.h"
 #include "input.h"
 #include "konsole_wcwidth.h"
@@ -297,6 +298,12 @@ void Shell::handleResize(uint64_t cols, uint64_t rows)
 		}
 		m_image.swap(new_image);
 		updateGeometry();
+		// Layout recalculation is done in next event cycle, but we start
+		// rendering right after emitting the neovimResized() signal.
+		if (parentWidget() && parentWidget()->layout()) {
+			parentWidget()->layout()->invalidate();
+			parentWidget()->layout()->activate();
+		}
 		emit neovimResized(neovimSize());
 	}
 }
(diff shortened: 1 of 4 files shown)


src/gui/shell.cpp
```diff
@@ -292,6 +298,12 @@ void Shell::handleResize(uint64_t cols, uint64_t rows)
 		}
 		m_image.swap(new_image);
 		updateGeometry();
+		// Layout recalculation is done in next even cycle, but we start
+		// rendering right after emitting the neovimResized() signal.
```
equalsraf: even cycle -> event cycle?

src/gui/shell.cpp
```diff
@@ -292,6 +298,12 @@ void Shell::handleResize(uint64_t cols, uint64_t rows)
 		}
 		m_image.swap(new_image);
 		updateGeometry();
+		// Layout recalculation is done in next event cycle, but we start
+		// rendering right after emitting the neovimResized() signal.
+		if (parentWidget() && parentWidget()->layout()) {
+			parentWidget()->layout()->invalidate();
+			parentWidget()->layout()->activate();
+		}
```
equalsraf: Doesn't updateGeometry cause this to happen anyway?

nowrep: Yes, in next event cycle, so `sizeHint()` is wrong in slot connected to `neovimResize()` that is emitted on the next line.

equalsraf: Ah, then emitting the signal from a QTimer(0) should work too.

nowrep: Or using `Qt::QueuedConnection` (that was my first idea to fix it) from mainwindow, but then there comes the rendering bugs.

equalsraf: Yeah, I'm having similar issues - one of these days I'll have to rewrite the shell widget to avoid those anyway. We can leave it as it is for now.

equalsraf:
This is fixed by #109.

PR #81, opened 2015-12-18
Replace spaces with tabs in shell.h

Marenz:


equalsraf:
Thanks @Marenz, merging.

Issue #86, opened 2016-01-23
mac homebrew formula?


rkitover:
Hi, I noticed there is no mac homebrew formula for this project, I was thinking we could do something like [this](https://github.com/neovim/homebrew-neovim).

I am happy to write the formula for you guys if you are interested.

equalsraf:
Go for it :D

I don't have a Mac where I can test though, I suspect there will be some issues with the keyboard Ctrl/Cmd keys.

rkitover:
The cmd key is not really used in vim, the ctrl key is indeed borked.

The other issue is that the default font on mac is terrible, here is a screenshot:

![screenshot](https://www.dropbox.com/s/eynldi1k11e21rs/Screenshot%202016-01-24%2002.59.03.png?dl=1)

I'll try to send you a PR for these two things over the next few days.

equalsraf:
> The cmd key is not really used in vim, the ctrl key is indeed borked.

According to the [Qt docs](http://doc.qt.io/qt-5/qt.html#Key-enum) the Qt::Key_Control is Cmd and Qt::Key_Meta is Ctrl. So its likely Cmd is being accepted as Ctrl key.

> The other issue is that the default font on mac is terrible, here is a screenshot:

We just set Monospace 10pt by default in shell.cpp.

rkitover:
Just checked and cmd key does not function as a ctrl, but then cmd is a system key on macs so likely not mappable by applications, except for combos.

rkitover:
I noticed you have no tags, the way homebrew formulae that get their sources from git work is that they reference a specific tag, and are constantly updated for new tags. The reason for this is so that when you do a `brew upgrade` you won't always rebuild the same thing needlessly and for stability reasons of course.

Would you be willing to push version tags occasionally? This would of course also help packagers for other systems.

equalsraf:
Sure, just tagged v0.1.0

gwerbin:
For those following this issue: I just wrote a formula

https://github.com/equalsraf/neovim-qt/issues/135

equalsraf:
Closing for #135

PR #87, opened 2016-01-27
fix ctrl on mac and better default fonts #86

rkitover:
This makes neovim-qt fully functional on Macs from my limited testing.

I would appreciate a thorough review because I don't actually know very much about C++, and of course I will be happy to implement any suggested improvements.

Details are in commit message.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -57,21 +57,25 @@ InputConv::InputConv() {
  */
 QString InputConv::modPrefix(Qt::KeyboardModifiers mod)
 {
-	// FIXME: Mac and Meta check the Qt docs
 	QString modprefix;
-	if ( mod & Qt::ControlModifier
+#ifdef Q_OS_MAC
+	if ( mod & CmdModifier ) {
+		modprefix += "D-"; // like MacVim does
+	}
+#endif
+	if ( mod & ControlModifier
 #ifdef Q_OS_WIN32
-		&& !(mod & Qt::AltModifier)
+		&& !(mod & AltModifier)
 #endif
 	   ) {
 		modprefix += "C-";
 	}
-	if ( mod & Qt::ShiftModifier ) {
+	if ( mod & ShiftModifier ) {
 		modprefix += "S-";
 	}
-	if ( mod & Qt::AltModifier
+	if ( mod & AltModifier
 #ifdef Q_OS_WIN32
-		&& !(mod & Qt::ControlModifier)
+		&& !(mod & ControlModifier)
 #endif
 	   ) {
 		modprefix += "A-";
@@ -148,37 +152,51 @@ QString InputConv::convertKey(const QString& text, int k, Qt::KeyboardModifiers 
 {
 	if (specialKeys.contains(k)) {
 		return QString("<%1%2>").arg(modPrefix(mod)).arg(specialKeys.value(k));
-	} else if (text.isEmpty()) {
-		// This is a special key we can't handle
-		return QString();
 	}
 
+	QChar c;
 	// Escape < and backslash
 	if (text == "<") {
 		return QString("<%1%2>").arg(modPrefix(mod)).arg("lt");
-	}
-	if (text == "\\") {
+	} else if (text == "\\") {
 		return QString("<%1%2>").arg(modPrefix(mod)).arg("Bslash");
+	} else if (text.isEmpty()) {
+		// on macs, text is empty for ctrl+key and cmd+key combos (with or without alt)
+		if (mod & ControlModifier || mod & CmdModifier) {
+			// ignore ctrl, alt and cmd key combos by themselves
+			QList<Qt::Key> keys = { Key_Control, Key_Alt, Key_Cmd };
+			if (keys.contains((Qt::Key)k)) {
+				return QString();
+			} else {
+				// key code will be the value of the char (hopefully)
+				c = QChar(k);
+			}
+		} else {
+			// This is a special key we can't handle
+			return QString();
+		}
+	} else {
+		// Key event compression is disabled, text has one char
+		c = text.at(0);
 	}
 
-	QChar c = text.at(0);
 	// Remove SHIFT
 	if (c.unicode() < 0x100 && !c.isLetterOrNumber() && c.isPrint()) {
-		mod &= ~Qt::ShiftModifier;
+		mod &= ~ShiftModifier;
 	}
 
 	// Remove CTRL empty characters at the start of the ASCII range
 	if (c.unicode() < 0x20) {
-		mod &= ~Qt::ControlModifier;
+		mod &= ~ControlModifier;
 	}
 
 	// Format with prefix if necessary
 	QString prefix = modPrefix(mod);
 	if (!prefix.isEmpty()) {
-		return QString("<%1%2>").arg(prefix).arg(text);
+		return QString("<%1%2>").arg(prefix).arg(c);
 	}
 
-	return text;
+	return QString(c);
 }
 
 } // Namespace
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -21,7 +21,7 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 	m_resizing(false), m_logo(QPixmap(":/neovim.png")),
 	m_neovimBusy(false)
 {
-	m_font = createFont("Monospace");
+	m_font = createFont(DEFAULT_FONT);
 	m_font.setPointSize(10);
 	m_fm = new QFontMetrics(m_font);
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,9 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 # konsole_wcwidth.h
 include_directories(third-party)
 
+# homebrew/macrpots/fink etc.
+include(MacOSXPaths)
+
 # Qt
 set(CMAKE_AUTOMOC ON)
 find_package(Qt5Widgets REQUIRED)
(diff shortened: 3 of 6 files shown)


equalsraf:
Is Monaco the right choice for a default on a Mac? I did a quick test on 10.10

- the shipped Monaco font lacks italic variants
- Courier/Courier New/Menlo are also fixed width and are shipped with all variants

src/gui/shell.cpp
```diff
@@ -21,7 +21,17 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 	m_resizing(false), m_logo(QPixmap(":/neovim.png")),
 	m_neovimBusy(false)
 {
-	m_font = createFont("Monospace");
+	m_font = createFont(
+#if defined(Q_OS_MAC)
+		"Monaco"
+#elif defined(Q_OS_WIN32)
```
equalsraf: Its better to leave the Windows and Android cases for another PR.

rkitover:
Give me a few minutes I will change it, I also have some more stuff I did last night.

equalsraf:
No rush I still need to have a deeper look, and I'm working on some checks too.

equalsraf:
If you want you can cherry-pick 5f7a552e125a29486563136099a7699daa52e33a to add unit tests.

rkitover:
I split the input changes and the font fix into two separate commits.

I also realized this morning that my input code didn't handle ctrl+alt combos, I fixed this as well hopefully.

I added some cmake code to add search paths on OS X in the third commit, for supporting homebrew, macports and fink, and automatically finding the homebrew qt5 and zlib.

Some issues with the build:

- GTest is not included in homebrew, the message says you should include it in your project, yet I can run `make test`, is GTest being used at all?
- msgpack is included in homebrw, should the build use the user's msgpack if it is installed or always download it?
- I get a test failure:
```
The following tests FAILED:
          5 - tst_msgpackiodevice (Failed)
```
I don't know anything about C++ tests, how would I see where it failed and debug?

rkitover:
Well, without this code I have to run cmake with

```
cmake -DCMAKE_PREFIX_PATH=/usr/local/opt/qt5
```

for qt5 to be found, because homebrew qt5 is keg-only and not linked to /usr/local, neither is zlib.

equalsraf:
> GTest is not included in homebrew, the message says you should include it in your project, yet I can run make test, is GTest being used at all?

No, this is printed by msgpack, for tests we only use QtTest.

> msgpack is included in homebrw, should the build use the user's msgpack if it is installed or always download it?

By default it downloads msgpack. You can use your own with cmake -DUSE_SYSTEM_MSGPACK=ON ..

I'm not sure what is the usual policy in brew recipes concerning linked libraries. I have not issue with using the one already provided in brew.

> I get a test failure:

I saw this too. e676a5ccb5980cd469c986f481473209b3d1d92c fixes this issue. Can you cherry pick it?

src/gui/input.cpp
```diff
@@ -144,13 +143,26 @@ QString InputConv::convertMouse(Qt::MouseButton bt, QEvent::Type type, Qt::Keybo
...
+				return QString();
+			} else {
+				// key code will be the value of the char (hopefully)
+				text = QString(QChar(k));
+			}
+		} else {
+			// This is a special key we can't handle
+			return QString();
+		}
 	}
```
equalsraf: Reversing the logic here helps make it more readable, and avoids the extra string var

```diff
diff --git a/src/gui/input.cpp b/src/gui/input.cpp
index 7d91c97..b2ece2c 100644
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -143,12 +143,18 @@ QString InputConv::convertMouse(Qt::MouseButton bt, QEvent::Type type, Qt::Keybo
  *
  * see QKeyEvent
  */
-QString InputConv::convertKey(const QString& in_text, int k, Qt::KeyboardModifiers mod)
+QString InputConv::convertKey(const QString& text, int k, Qt::KeyboardModifiers mod)
 {
-       QString text = in_text;
-
        if (specialKeys.contains(k)) {
                return QString("<%1%2>").arg(modPrefix(mod)).arg(specialKeys.value(k));
+       }
+
+       QChar c;
+       // Escape < and backslash
+       if (text == "<") {
+               return QString("<%1%2>").arg(modPrefix(mod)).arg("lt");
+       } else if (text == "\\") {
+               return QString("<%1%2>").arg(modPrefix(mod)).arg("Bslash");
        } else if (text.isEmpty()) {
                // on macs, text is empty for ctrl+key combos
                if (mod & ControlModifier) {
@@ -157,23 +163,17 @@ QString InputConv::convertKey(const QString& in_text, int k, Qt::KeyboardModifie
                                return QString();
                        } else {
                                // key code will be the value of the char (hopefully)
-                               text = QString(QChar(k));
+                               c = QChar(k);
                        }
                } else {
                        // This is a special key we can't handle
                        return QString();
                }
+       } else {
+               // Key event compression is disabled, text has one char
+               c = text.at(0);
        }
 
-       // Escape < and backslash
-       if (text == "<") {
-               return QString("<%1%2>").arg(modPrefix(mod)).arg("lt");
-       }
-       if (text == "\\") {
-               return QString("<%1%2>").arg(modPrefix(mod)).arg("Bslash");
-       }
-
-       QChar c = text.at(0);
        // Remove SHIFT
```

rkitover:
Yes tests pass now.

I really want to fix this in cmake itself in the platform files at some point if I get around to it, in the meantime I will improve this and make it an included file like you said.

As for msgpack, would you be interested in having the cmake logic amended to use the system msgpack if it is found and of recent enough version, and only then fall back to downloading it?

I will update the PR in a few hours.

equalsraf:
> As for msgpack, would you be interested in having the cmake logic amended to use the system msgpack if it is found and of recent enough version, and only then fall back to downloading it?

No, just leave it as is for now.

rkitover:
Ok I updated these commits, sorry it took so long I wasn't feeling well yesterday.

* added your changes to the input code and tested, needed to make one minor change
* moved cmake path code into include file, with a fix to not generate link warnings for paths that don't exist

I was thinking, if you are interested, that I could add support for building an .app bundle. This is not a minor project though and is complicated, it would need to be linked statically and include a copy of nvim in the bundle. In a separate PR of course. But in any case, the first priority is the homebrew formula.

Minor nit: do you think 10 is a bit too small for the default font point size? I'd use 14 or at least 12.

equalsraf:
> Ok I updated these commits, sorry it took so long I wasn't feeling well yesterday.

No problem. I'm pretty busy myself. I won't have access to any Mac OSX machines before monday though.

> I was thinking, if you are interested, that I could add support for building an .app bundle. This is not a minor project though and is complicated, it would need to be linked statically and include a copy of nvim in the bundle. In a separate PR of course.

You'll probably want to delay this. Its likely Neovim will change the way it handles GUIs (at which point Neovim calls the GUI), its probably best to avoid working on this until the Neovim stabilizes on that front - https://github.com/neovim/neovim/issues/3901.

> Minor nit: do you think 10 is a bit too small for the default font point size? I'd use 14 or at least 12.

To me sure, but my eyes don't work that well :D. This seems to be the default in gvim, macvim, so lets leave it. Most likely people will change fonts anyway.

Slightly related - I'm not sure what is the support for Cmd in macvim, but based on the docs it seems we should be able to support the Cmd key as D-.

rkitover:
I will look into how macvim handles cmd and windows gvim handles the win key and make any necessary adjustments.

rkitover:
I added support for mapping cmd combos to D- like MacVim does, unfortunately neovim does not recognize this right now, perhaps this needs to be fixed in neovim itself?

You can try this with a simple `nmap` like so:

```vim
nmap <D-1> iGOT CMD-1!<Esc>
```

this works in MacVim but not here yet, even though the key string is correct.

Here is what I know about the Win/Cmd key:

* as you pointed out, you can make some mappings with the cmd key in MacVim, this is described in `:h key-notation`
* on Windows some apps can make some mappings with the Win key, like AutoHotKey etc., but according to `:h key-notation` at least, windows gvim has no support for this
* on linux desktops, the Win key is usually mapped to Super, and as far as I can tell, only the Window Manager is allowed to make mappings with the Super key, and again no mention of this in `:h key-notation`

rkitover:
Looks like the build is failing on MSVC because of a missing cast, fixing...

rkitover:
Ah good, fixed now. Reworded commit messages as well.

equalsraf:
> I added support for mapping cmd combos to D- like MacVim does, unfortunately neovim does not recognize this right now, perhaps this needs to be fixed in neovim itself?

Indeed see https://github.com/neovim/neovim/issues/2190

I'll have a further look at this tomorrow, but otherwise it looks good.

equalsraf:
@rkitover I would suggest these changes ba80b5b4cb56f99b68a38f14081d4db9bdf219a9, otherwise it looks ready to merge.

rkitover:
Applied, except I put DEFAULT_FONT in shell.h .

Any luck with Travis? I looked at your job a couple of days ago and it seems you got further but had some sort of weird failure with one of the deps neovim downloads.

equalsraf:
> Any luck with Travis?

No luck so far it consistently throws the same download error.

rkitover:
Awesome, thank you for working with me on this!

About the Travis build, I will try to debug a bit in my fork and check with the folks on the travis channel on freenode as well.

So now I can get started on the homebrew formula, I should be done in a few days probably.

The repo has to be called `homebrew-neovim-qt`, after I'm done you can copy it to your namespace and it will be installable as `brew install equalsraf/neovim-qt` or as `brew install neovim-qt/neovim-qt` if at some point you decide to make an org. It can depend directly on the neovim tap and install it as a dep (very cool, just found that out recently.) Should be a pretty straight-forward formula.

If you are curious, I got involved in all of this because of one of my projects, I don't actually use neovim right now:

rkitover/vimpager#133

vimpager has a feature where it launches gvim instead of terminal vim, I want to add support for neovim-qt at some point as well, and this will probably take a couple more PRs before that's possible.

Issue #89, opened 2016-01-30
Can't install on Arch Linux


fuelen:
I can't install package from AUR with new version of msgpack-c.
See my comment (2016-01-27) here https://aur.archlinux.org/packages/neovim-qt-git/

PR #91, opened 2016-01-31
Update bundled msgpack to 1.4.0

equalsraf:
In msgpack 1.4.0 the library name is msgpackc and the msgpack target
is msgpackc-static. This should also fix building against msgpack
1.4.

For #89, @fuelen I didn''t get a chance to test on Arch, but this is working in my system with msgpack 1.4, can you test it?


fuelen:
It works :) Thank you!

Issue #96, opened 2016-02-09
Focus events are not triggered


mschulkind:
Events like `FocusGained` are not triggered when the window regains focus. 

This is a show stopper for me personally because this means that `autoread` does not work, as it is normally triggered by focus events.

equalsraf:
@mschulkind, thanks for pointing those out, completely missed them. The changes in #97 should add support for both events.

mschulkind:
now that's service :)

equalsraf:
Let me know if it works ok. I can't merge right away, but it looks simple enough to be merged later.

PR #97, opened 2016-02-09
GUI: Support FocusGained and FocusLost

equalsraf:
Forward widget focus events as <FocusGained> and <FocusLost>. For #96. Should work with `:au FocusGained ...`


Shougo:
:+1:

PR #101, opened 2016-03-01
set NVIM_TUI_ENABLE_TRUE_COLOR to 1

mikew:
I noticed vim-airline neovim-qt had [the same issues as Neovim.app](https://github.com/rogual/neovim-dot-app/issues/215), and the fix is the same.


equalsraf:
@mikew thanks.

equalsraf:
Closing this after the change upstream https://github.com/neovim/neovim/wiki/Following-HEAD#20160511 this can be set using an option.

Issue #83, opened 2015-12-20
Can't start maximized


cdvv7788:
I am using neovim with qt-neovim (linux arch). On gvim i could launch a maximized window with:

    if has("gui_running")
        set lines 999 columns 999
    endif

That is not working now tho. I tried removing the condition with no success.
Is there any way i can achieve that?

cdvv7788:
After restarting somehow it is reading the command, however it doesn't restrict the max size to my screen size, it creates a giant one.

equalsraf:
if has("gui_running") is not reliable right now (see neovim/neovim#3646)

> however it doesn't restrict the max size to my screen size, it creates a giant one.

This is strange, should not happen, will have a look.

cdvv7788:
Yeah, i read on some other issues that has("gui_running") is not reliable, so i tried without it too.

equalsraf:
The problem is vimrc will likely be processed before the gui even starts, so its all a matter of timing.

cdvv7788:
Any workaround for this? Anything i can help with on this issue?

equalsraf:
@cdvv7788 not forgotten, I've got something in #109

equalsraf:
#109 was merged, --maximized should start a maximized window. From what we have seen so far using set lines/columns is not a good way to do this, but we can finally start looking at other ways to do this.

khalidchawtany:
leaving a comment to be notified of anything new on toggle full screen command :)

equalsraf:
Starting to look at this on #114  ([script](https://github.com/equalsraf/neovim-qt/pull/114/files#diff-b996c9a7410476c5ddf91c4b00121b01R1))
Any idea what are the right semantics for this type of stuff

- are these supposed to be functions like `foreground()` or actual commands?
- Do we want toggle or explicit commands `GuiMaximize` vs `GuiMaximizeToggle`
- Have no idea how we are supposed to share state, ideally set g:GuiIsMaximized (because that sounds easy enough :D)

equalsraf:
Pushed something into #114 

If testing directly from a source build use this for now
```
bin/nvim-qt -- --cmd "set rtp+=../runtime/ | runtime autoload/nvim_gui.vim"
```

init.vim looks like this

```vim
" We can check first if the following commands are available, but the last one is just a var anyway
if exists('g:Gui')
    Guifont DejaVu Sans Mono:h15

    " Starts in fullscreen, Maximized should work too (untested)
[...]

PR #103, opened 2016-03-02
Enable Guifont force assignment

lambdalisue:
Multibyte users sometime require to use "Invalid font" so allow to specify that kind of font by option
With this change, the following will be available

    command -nargs=? -bang Guifont
        \ call rpcnotify(0, 'Gui', 'SetFont', <q-args>, <q-bang> ==# '!') |
        \ let g:Guifont = "<args>"

    " The following will show exception
    Guifont An invalid font:h10
    " The following won't
    Guifont! An invalid font:h10

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -80,7 +80,7 @@ QFont Shell::createFont(const QString& family)
 /**
  * Set the GUI font, or display the current font
  */
-bool Shell::setGuiFont(const QString& fdesc)
+bool Shell::setGuiFont(const QString& fdesc, bool force)
 {
 	if (fdesc.isEmpty()) {
 		QFontInfo fi(m_font);
@@ -114,15 +114,17 @@ bool Shell::setGuiFont(const QString& fdesc)
 		m_nvim->neovimObject()->vim_report_error(m_nvim->encode(errmsg));
 		return false;
 	}
-	if ( !fi.fixedPitch() ) {
-		QString errmsg = QString("%1 is not a fixed pitch font").arg(f.family());
-		m_nvim->neovimObject()->vim_report_error(m_nvim->encode(errmsg));
-		return false;
-	}
+	if (!force) {
+		if ( !fi.fixedPitch() ) {
+			QString errmsg = QString("%1 is not a fixed pitch font").arg(f.family());
+			m_nvim->neovimObject()->vim_report_error(m_nvim->encode(errmsg));
+			return false;
+		}
 
-	if (isBadMonospace(f)) {
-		QString errmsg = QString("Warning: Font \"%1\" reports bad fixed pitch metrics").arg(f.family());
-		m_nvim->neovimObject()->vim_report_error(m_nvim->encode(errmsg));
+		if (isBadMonospace(f)) {
+			QString errmsg = QString("Warning: Font \"%1\" reports bad fixed pitch metrics").arg(f.family());
+			m_nvim->neovimObject()->vim_report_error(m_nvim->encode(errmsg));
+		}
 	}
 
 	m_font = f;
@@ -573,9 +575,15 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 {
 	if (name == "Gui" && args.size() > 0) {
 		QString guiEvName = m_nvim->decode(args.at(0).toByteArray());
-		if (guiEvName == "SetFont" && args.size() == 2) {
-			QString fdesc = m_nvim->decode(args.at(1).toByteArray());
-			setGuiFont(fdesc);
+		if (guiEvName == "SetFont") {
+			if (args.size() == 2) {
+				QString fdesc = m_nvim->decode(args.at(1).toByteArray());
+				setGuiFont(fdesc);
+			} else if (args.size() == 3) {
+				QString fdesc = m_nvim->decode(args.at(1).toByteArray());
+				bool force = args.at(2) == 1;
+				setGuiFont(fdesc, force);
+			}
 		}
 		return;
 	} else if (name != "redraw") {
--- a/src/gui/shell.h
+++ b/src/gui/shell.h
@@ -39,7 +39,7 @@ signals:
 public slots:
 	void handleNeovimNotification(const QByteArray &name, const QVariantList& args);
 	void resizeNeovim(const QSize&);
-	bool setGuiFont(const QString& fdesc);
+	bool setGuiFont(const QString& fdesc, bool force=false);
 
 protected slots:
 	void neovimIsReady();


lambdalisue:
It seems executing `Guifont` in `init.vim` does not effect. It seems timing-issue mentioned at [this comment](https://github.com/equalsraf/neovim-qt/issues/83#issuecomment-166158006).

The following is quick workaround, in case someone desire to use `Guifont` for startup.

```vim
" Workaround for GUIEnter {{{
if exists('#GUIEnter')
  echohl WarningMsg
  echomsg 'GUIEnter autocmd exists. Time to remove workaround.'
  echohl None
endif
if !exists('g:_nvim_qt_GUIEnter')
  augroup nvim_qt_GUIEnter_workaround
    autocmd! *
    autocmd FocusGained *
          \ doautocmd <nomodeline> User GUIEnter |
          \ let g:_nvim_qt_GUIEnter = 1 |
          \ autocmd! nvim_qt_GUIEnter_workaround
  augroup END
endif
" }}}

augroup nvim_gui_fontset
  autocmd! *
  autocmd User GUIEnter Guifont! Ricty for Powerline:h9
  "autocmd GUIEnter * Guifont! Ricty for Powerline:h9
augroup END
```

equalsraf:
> Multibyte users sometime require to use "Invalid font" so allow to specify that kind of font by option

Can you give some example of fonts/chars where this happens? I'm asking because double column chars (like chinese or japanese) should still work.

> The following is a quick workaround, in case someone desire to use Guifont for startup.

I wonder if it would be simpler to just use `dictwatcheradd` to watch for `nvim_qt*`

lambdalisue:
> Can you give some example of fonts/chars where this happens? I'm asking because double column chars (like chinese or japanese) should still work.

[`Ricty`](https://github.com/yascentur/Ricty). This font works in Gnome-Terminal, GVim, iTerm2, ... else but in neovim-qt.

> I wonder if it would be simpler to just use dictwatcheradd to watch for nvim_qt*

I'm sorry but I'm not really familiar with neovim's implementation. Could you show me an example if you have time?

equalsraf:
Neovim has `dictwatcheradd` to call a function whenever entries in a dictionary are changed. For example lets say you want to call a function whenever a **global** variable starting with Gui is changed. for example (untested)

```viml

function! SendGuiEvent(dict, key, change)
	if a:key == 'Guifont'
		call rpcnotify(0, 'Gui', 'SetFont', g:Guifont)
	endif
endfunction

call dictwatcheradd(g:, 'Gui*', 'SendGuiEvent')

" The old Guifont command, now just sets g:Guifont
command! -nargs=? Guifont let g:Guifont="<args>"
```

i.e. `let g:Guifont=` will SendGuiEvent and rpcnotify. It does not solve the timing issue, but maybe it makes things simpler. It also allows us to use `let g:Guifont` directly instead of the command.

An alternative to using the **bang!** to force the font is to add an attribute i.e. `An invalid font:h10:forcefont`, or some other name.

equalsraf:
> Ricty. This font works in Gnome-Terminal, GVim, iTerm2, ... else but in neovim-qt.

For Ricty I get `Ricty is not a fixed pitch font`, so it does not seem to be a monospace font. This is not caused by the wide asian characters, looking at the font metrics it would seem this is clearly not monospace (e.g. for 10pt)

- The average font width is 12px
- The reported maximum font width is 13px
- The width of two single column chars is 14px (??? this doesn't seem right)

BTW, if you are building neovim-qt from source ,there is a binary fontinfo that can show you more information about a font, just call `fontinfo Ricty`

lambdalisue:
> Neovim has dictwatcheradd to call a function whenever entries in a dictionary are changed. For example lets say you want to call a function whenever a global variable starting with Gui is changed. for example (untested)

I see. Thanks. Btw do you have any plan to support `set guifont=XXX` which is supported in GVim.

> This is not caused by the wide asian characters, looking at the font metrics it would seem this is clearly not monospace (e.g. for 10pt)

Sorry. I thought `Ricty is not a fixed pitch font` because of wide asian characters which Ricty contains. But anyway I think there should be some way to force.

equalsraf:
> I see. Thanks. Btw do you have any plan to support set guifont=XXX which is supported in GVim.

The only way to support set guifont is by changing Neovim upstream.

> Sorry. I thought Ricty is not a fixed pitch font because of wide asian characters which Ricty contains. But anyway I think there should be some way to force.

This would take some effort though. GVim does it by assuming a very large font width, for example when I load Ricty all characters have a lot of empty space around them.

lambdalisue:
> GVim does it by assuming a very large font width, for example when I load Ricty all characters have a lot of empty space around them.

Hum... Not for my GVim and patched neovim-qt.

selection_005

equalsraf:
Can you send me the ttf fonts?

lambdalisue:
Hum... I'm sorry but no. Ricty has a License problem so I cannot distribute compiled one. I'll try with RictyDimnish and tell you a result.

lambdalisue:
selection_006

With **Ricty Diminished:h9**

lambdalisue:
selection_007

With `$ gvim -u NONE`

jasonwilliams200OK:
> Ricty has a License problem so I cannot distribute compiled one

probably make sense to find some other font having same issue with permissive license for steps to repro: http://sscce.org/

lambdalisue:
> probably make sense to find some other font having same issue with permissive license for steps to 

Ricty Diminished has same problem without this patch.

lambdalisue:
> GVim does it by assuming a very large font width, for example when I load Ricty all characters have a lot of empty space around them.

@equalsraf I just intrested in but could you show me a screenshot?

equalsraf:
Ricty 11pt In gVim

gvim

lambdalisue:
Because you specify `pt`. Try `:set guifont=Ricty\ Diminished\ 11`

lambdalisue:
Gvim seems to have different approach for this problem. I'll check and will modify this pr!

lambdalisue:
https://github.com/vim/vim/blob/master/src/gui_gtk_x11.c#L5292-L5327

equalsraf:
> Because you specify pt. Try :set guifont=Ricty\ Diminished\ 11

Indeed now it looks ok.

> Gvim seems to have different approach for this problem. I'll check and will modify this pr!

When forcing a font with a non fixed width some issues will always occur, there is no way around this. The strategy in gVim is to find the width of a double column char and divide it by two. Ours is to just consider the width of the `W` char for a single column.

lambdalisue:
So... I don't really remember but what can I do next?

It seems the conversation went a wrong way.
The most important point is **I would like to use a non fixed width font like GVim**. 
So is `Guifont!` approach is not correct way to ignore the validation or you don't want to accept such exception?

equalsraf:
> It seems the conversation went a wrong way

Hardly, the reason this didn't get merged at the time was because there were several other PRs that changed the font related code (the shell widget and the runtime scripts), and conflicted with this one, the later of which only got merged into master recently.

I have no problem with adding an exception mechanism. This just needs to be changed after all the changes that took place, in particular

- Right now the c++ code that sets the font moved over to shellwidget.cpp
- the VimL for the GuiFont command is now in src/gui/runtime

lambdalisue:
I made new PR to follow the HEAD.

PR #107, opened 2016-03-12
GUI: Allow font style customization with :Guifont

jpleau:
I personally work with Monaco Bold, this change makes neovim-qt look exactly like my gvim setup. I added the extra styles in case anyone would like those. 

You can use them like in gvim with Guifont, for example to get Monospace with Bold+Italic:

`:Guifont Monospace:h12:b:i`

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -81,6 +81,8 @@ bool Shell::setGuiFont(const QString& fdesc)
 	}
 
 	int pointSize = font().pointSize();
+	int weight = -1;
+	bool italic = false;
 	foreach(QString attr, attrs) {
 		if (attr.size() >= 2 && attr[0] == 'h') {
 			bool ok = false;
@@ -90,9 +92,13 @@ bool Shell::setGuiFont(const QString& fdesc)
 				return false;
 			}
 			pointSize = height;
+		} else if (attr == "b") {
+			weight = QFont::Bold;
+		} else if (attr == "i") {
+			italic = true;
 		}
 	}
-	setShellFont(attrs.at(0), pointSize);
+	setShellFont(attrs.at(0), pointSize, weight, italic);
 
 	if (m_attached) {
 		resizeNeovim(size());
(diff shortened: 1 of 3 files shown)


equalsraf:
I think we will hold off on the strike through and underline styles for now.

The one issue I'm seeing here is that the shell does not get updated immediately.

jpleau:
I removed the extra styles (strikeout / underline).

To make sure the shell gets updated when the font style is changed, do I have to redraw all the text? Depending on if the QPainter's font was bold or not when text was drawn the first time, it might need to be redrawn to pickup the new font-style

equalsraf:
Can you rebase this, to check if it works for the new widget.

equalsraf:
> jpleau @equalsraf: I got your message, I notice that you no longer build a QFont in Shell::setGuiFont() directly but build it later in setShellFont(). How would you prefer I pass the options for the fonts settings?
> jpleau Worth doing a custom struct for that?

@jpleau answering here insteadto keep history

Completely forgot I moved the method into the ShellWidget, I think adding a couple arguments with default values is enough, i.e. change setShellFont() to mimic the signature from the QFont constructor

    QFont(const QString &family, int pointSize = -1, int weight = -1, bool italic = false)

jpleau:
That is the exact code I have used in my branch :)

I will use this one, and set weight to QFont::Bold for bold font.

I will reply here instead of on freenode, sorry about that

equalsraf:
:+1: 

> I will reply here instead of on freenode, sorry about that

No problem, it was just late and I didn't want to loose track of this.

jpleau:
Rebased with current master

fuelen:
Hi. How can I enable use this in init.vim?
I want to use Terminus font.
When I write `:GuiFont Terminus:h10` - all is ok.
But if I put this (GuiFont Terminus:h10) into init.vim then font isn't changes.

equalsraf:
in **ginit.vim** not init.vim

fuelen:
thank you, it works

Issue #108, opened 2016-03-12
How should window resizing behave?


equalsraf:
More on this in a PR that should pop up in a bit, we need to discuss how resizing should behave to address #106, #83, #84(indirectly, resizes in layouts), and #4.

Resizes can arrive from multiple sources
1. Neovim itself may trigger a resize, e.g. `set lines`
2. User interaction i.e. user resizes the window
3. Window manager or widget layout changes, something in the window changed e.g. the layout of other widgets, that caused the Shell widget to be resized as a consequence

Neovim resizing works as follows 
1. The GUI calls neovim for a resize - see `ui_try_resize()`
2. Neovim (eventually) triggers the correct event (followed up by a redraw), but there may be other events in the pipeline
3. (I think) if the GUI triggers multiple resizes simultaneously (without waiting for completion) the following requests are dropped, very common when resizing a window

A few cases that need to be considered
1. The shell widget might be a window (maximized or not)
2. In some systems there is the notion of full screen mode (MacVim, Vim-Qt) it usually behaves very differently from maximized window
3. The shell widget might be part of another widget
4. Not all GUIs behave the same, check out vim-gtk, vim-qt, MacVim

The important bit to understand here is that a resize can trigger another resize and sometimes this is racy e.g.
- Some users drop a set lines/cols in init.vim, should that work for the GUI?
- If the windows is maximized and Neovim triggers a resize (e.g. plugin or set lines) should the window leave the maximized state and be resized - I feel this breaks user expectations - for e.g. vim-qt resized vim back to its previous size
- If Neovim is actually an embeded widget then resizes are likely not possible, but races can occur - Neovim resizes, the layout resize back 
- When resizing a window (depending on the window manager) the resize is progressive, i.e. one bit at a time, this is terribly inefficient.
- For a non-maximized window should we allow resizes larger than the current screen/workspace (e.g. set lines=2000)?
- Not all window managers respond to these events in the same way, e.g. KWin might allow you to resize the window, i3 will not

Ping #106, #83, #4. TLDR; need to agree on what is the expected behaviour for resizes, and need to write unit tests that check the correct behaviour (upcoming PR should help).

kierun:
Just adding another fly in the ointment: Some window managers ([awesome](http://awesome.naquadah.org/) for one) allow you to maximise a window vertically and/or horizontally and back again. This is a feature I use a lot as I tend to have many instances of (neo)vim open at once.

[...]

Issue #39, opened 2015-09-09
Fonts are blurred on HiDPI displays


johanhelsing:
Example:
![screenshot from 2015-09-09 10-15-46](https://cloud.githubusercontent.com/assets/2620557/9756597/c080cad0-56db-11e5-818b-f5268ebb4b2c.png)

equalsraf:
Hi @johanhelsing. I don't have a HDPI display to test with, but lets try and narrow it down. Does this also happen with other Qt apps? What version of Qt are you running and in what system?

Does setting the environment variable QT_DEVICE_PIXEL_RATIO=2 help at all? or QT_DEVICE_PIXEL_RATIO=AUTO.

There are bug reports in Qt apps for similar issues around HDPI ([here](http://www.itopen.it/qgis-and-qt-getting-ready-for-hidpi-screens/) or in [QtCreator1](https://bugreports.qt.io/browse/QTCREATORBUG-13578) and [here](https://bugreports.qt.io/browse/QTCREATORBUG-14283)). I don't think we are doing anything wrong on our side (e.g. we set font size in points no in pixels).

johanhelsing:
Hi, I just tested. I already had QT_DEVICE_PIXEL_RATIO set to 2... So I tried with 1. And the issue went away!
![screenshot from 2015-09-12 15-39-34](https://cloud.githubusercontent.com/assets/2620557/9831787/87eb21ee-5964-11e5-8ba1-8b28184dc5a2.png)

Looks exactly the same, but the font is not blurred anymore. Not sure why this happens. I develop Qt apps myself and I don't have that problem.

How are you rendering the glyphs/text? Could you point me to the right point in the source code, and I could try some experiments.

Don't have the time to look at the issues you linked to right now.

Btw, you shouldn't you be able to test and see the issues for yourself with that environment variable? Does it look good on your machine?

equalsraf:
> Btw, shouldn't you be able to test and see the issues for yourself with that environment variable? Does it look good on your machine?

Good call, confirmed QT_DEVICE_PIXEL_RATIO=2 breaks it for me. Text is rendered in [handlePut()](https://github.com/equalsraf/neovim-qt/blob/cf911aecdb920341b4290c4221db6dbb9e2f3578/src/gui/shell.cpp#L252).

![bug1](https://cloud.githubusercontent.com/assets/715238/9833772/f5f0f956-599c-11e5-97d9-8d74596812e0.png)

[...]

PR #109, opened 2016-03-12
Reimplement the shell widget

### LINKED ISSUE #3: Cursor for fullwidth characters
(in the tb-remote-ui branch) the text cursor is painted over existing characters. However for fullwidth characters, since we assume the cursor width is one cell it only covers half of the char.

snapshot-neovim-qt1

### LINKED ISSUE #84: White line on the top of window
I use `subtle` tiling WM. First screen - `vim-qt`, Second - `neovim-qt`

There is white line on the top of second screen. When I click right mouse button on it, then I can see small popup with inactive checkbox.
Also `nvim-qt` starts not fullscreen (left, top and bottom). Only reloading subtle's config makes `neovim-qt` fullscreen (but other programs starts fullscreen without reloading).

Arch Linux. latest neovim and neovim-qt

equalsraf:
This is a pretty big PR and no doubt will break some things. The code for this has been sitting in my hard drive for several months now and I've lacked the time to work it through, but we really need to push this to address a lot of outstanding issues.

- The new widget implementation is independent of the Neovim specific bits
- No more QImage, we only paint when needed
- The widget now has a copy of the contents (i.e. chars and formatting) that are displayed, this means we can redraw total or partial parts of the widget

### Challenges

1. Is performance worse? To me when testing it seems comparable, but I can't say for sure, specially on Windows (and I didn't get a chance to test on Mac). The store could likely benefit from manual optimizations.
2. Likely there are some regressions in rendering issues (off by a pixel, or something I've forgot to add)
3. Resizing is broken (but differently that before) but that is ok, we need to have a serious discussion on how resizing should behave - and with the previous implementations there were things we couldn't do. See #108.
4. I'm not very happy about the cell/shellcontents API

## Conditions

- [x] Fix #3
- [x] Hopefully this also addresses #39. **But I don't have a way to test with old Qt versions or a real HiDPI**.
- [x] Fix regression on #70, the timeout is being triggered, which is why it "seems" to work, we can connect to the neovimResized signal added in the last commit
- [x] Revert minimum size policy, we can't reduce the window size like this
- [x] Add --maximized not exactly the same as #83, but at least it seems to work without the previous issues
- [x]  Fix #84, or at least it should get rid of glitches seen in #75
- [x] BUG: when using the wildmenu, there is some sort of scrollling? bug


equalsraf:
In Windows --maximized/--fullscreen is clearly racy, the window gets maximized but the actual Neovim size is smaller than what it should be.

johanhelsing:
What kind of old qt versions so you wish to test with?

equalsraf:
@johanhelsing not that important anymore since the behaviour last night seemed consistent in 5.5, but if you still have 5.4 around can you give it a try?

johanhelsing:
Might give it a try when I get home, for the record, it looks nice with 5.5 except for a small white border at the top.

screenshot from 2016-03-15 10-55-20

equalsraf:
:D @johanhelsing awesome. The while line is a margin from the docking area (#84), but maybe after this is done #75 will work too.

equalsraf:
Is it just me or the wildmenu seems slow?

johanhelsing:
Not sure if it's performance, but there's definitely some flickering going on when using wildmenu. Also, the cursor seems to jump around in places it shouldn't be.

It's also visible in some plugins, i.e. ctrl-p.

equalsraf:
> Not sure if it's performance, but there's definitely some flickering going on when using wildmenu.

It almost seems that its failing to update the contents sometimes.

> Also, the cursor seems to jump around in places it shouldn't be.
.
Looking at the paint event it is possible we are painting the cursor in places where it is not really needed.

fuelen:
[out-1.ogv.zip](https://github.com/equalsraf/neovim-qt/files/174708/out-1.ogv.zip)
Just see video in archive (subtle WM)

equalsraf:
@fuelen not completely unexpected, saw a similar issue with i3. Just pushed a fix.

fuelen:
It's look nice for now :)

src/gui/shell.cpp
```diff
@@ -802,15 +615,31 @@ void Shell::wheelEvent(QWheelEvent *ev)
...
-			(cols != m_cols || rows != m_rows) ) {
-		m_nvim->neovimObject()->ui_try_resize(cols, rows);
+/// Resize remote Neovim.
+///
+/// Neovim ignores concurrent resizes. If you call this method while
+/// a resize is already in progress, the later call is delayed until
+/// the previous one is finished.
+void Shell::resizeNeovim(uint64_t n_cols, uint64_t n_rows)
+{
+	if (!m_nvim || (n_cols == columns() && n_rows == rows())) {
```
ivegotasthma: /home/ico/public/git/neovim-qt/src/gui/shell.cpp:636:25: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
      if (!m_nvim || (n_cols == columns() && n_rows == rows())) {
                             ^
    /home/ico/public/git/neovim-qt/src/gui/shell.cpp:636:48: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
      if (!m_nvim || (n_cols == columns() && n_rows == rows())) {
                                                    ^

Conversion warnings on Arch with gcc 5.3.0.

equalsraf:
Perhaps related to the cursor issues https://github.com/equalsraf/neovim-qt/issues/112#issuecomment-197567229, thanks @fuelen, seems likely we are painting the cursor when we shouldn't.

fuelen:
@equalsraf, oh sorry, that comment should be in this pull request

equalsraf:
@fuelen pushed a fix for the cursor issue, can you check if it still happens?

johanhelsing:
@equalsraf the cursor issue is fixed for me :)

equalsraf:
@johanhelsing awesome, thanks.

Is anyone seeing performance issues on Linux, when scrolling (maybe with :set cul)

fuelen:
@equalsraf and not fixed for me. 
And yes - navigation is very slow and jumping with :set cul

equalsraf:
> And yes - navigation is very slow and jumping with :set cul

Pushed a fix for the performance issue.

equalsraf:
Can't reproduce the cursor issue right now, @fuelen in the video what settings are you using for listchars/tabs/indentation?

fuelen:
@equalsraf https://github.com/fuelen/nvimfiles

equalsraf:
Thanks, :set cul was hiding the issue, should be fixed now.

fuelen:
yes, fixed

equalsraf:
Awesome, I still have to check this a bit further, but otherwise this should be ready to merge.

Issue #94, opened 2016-02-05
Condition for checking when in nvim vs. nvim-qt


james2doyle:
Is there a variable or if statement I can do to tell when in nvim vs. nvim-qt?

equalsraf:
Not at the moment. I've been thinking about it a bit, we could set a variable on attachment and unset it on detach. Ultimately this should be fixed when the issues pointed out in #50 are fixed upstream (neovim/neovim#3646).

equalsraf:
In a nutshell there are limitations to what can be done, and I'm not able to cover all the cases, usually one of two things happens when launching the GUI

1. nvim-qt *spawns* the nvim process - in practice nvim first reads init.vim and then accepts a connection from the GUI
2. nvim-qt *attaches* to an already running nvim process

In general checking for the GUI presence when parsing init.vim is not very reliable, because for all intents and purposes it might not be *attached* yet (event if nvim was launched by the GUI). Some people use autocommands to defer GUI configurations (see #95) but these are not exactly ideal yet. One of the possibilities is that Neovim will load something like ginit.vim when running a GUI.

Onward to what can be done right now:

**Option 1** The GUI could set global variables on attach and detach. But this would not helo with init.vim. And in general I dislike this, because its to fail to unset the variables on detach (connection errors and crashes do happen).
**Option 2** The GUI sets an environment variable when spawning the nvim process. Works for init.vim, but it does not work at all when attaching to a running nvim.
**Option 3** Source a predetermined file after the GUI attaches successfully. Maybe the GUI order loading of ginit.vim. There are a couple of corner cases here too, like what happens if we attach multiple GUIs, or when attach and detach multiple times?

Personally I'm slightly more fond of **Option 3** because it seems it could be mostly compatible with future Neovim behaviour. **Option 2** is by far the easiest, it likely solves what most people want (an easy check in `init.vim`) but it will likely be broken later on. The detach behaviour is still unclear to me.

ping @jpleau because he asked the same question in gitter.

lambdalisue:
> In general checking for the GUI presence when parsing init.vim is not very reliable, because for all intents and purposes it might not be attached yet

But I think `has('gui_running')` or whatever is useful to make a compatible vimrc. For example, I'm using Vim plugin manager which has *lazy* implementation and I do want to source plugins for GUI automatically for neovim-qt but neovim. In this case, neovim-qt does not necessary to be attached at that moment.

So generally I think **Option 2** is required at least.
[...]

PR #114, opened 2016-04-04
Add GUI helper plugin

### LINKED ISSUE #64: v:windowid is not defined

### LINKED ISSUE #83: Can't start maximized

equalsraf:
Basically load a VimL helper plugin to help simplify some issues, avoid writing more C++, and make it easier to add new commands or global variables.

This partially helps with things like #94, ~~`exists('g:Gui')`~~ does not guarantee the GUI is attached but it guarantees that whatever commands the helper plugin provides are available,  So far the only thing they do is setting global variables and calling `rpcnotify(0, ...)` which should be harmless if the GUI is not attached.

This is also required for #64/#65, but we still need to define an API for that, likewise for #83

- [x] Partially fix #83 
- [x] Fixes #64
- [x] Move helper plugin to a separate repo, easier to setup when using a remote GUI
- [x] Check scope for option function in the script
- [x] Unit tests
- [x] Load ginit.vim on attachment

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -11,13 +11,17 @@ add_subdirectory(shellwidget)
 
 include(GNUInstallDirs)
 set(RUNTIME_PATH )
+add_library(neovim-qt-gui shell.cpp input.cpp errorwidget.cpp mainwindow.cpp
+	${CMAKE_SOURCE_DIR}/third-party/konsole_wcwidth.cpp
+	${NEOVIM_RCC_SOURCES})
+target_link_libraries(neovim-qt-gui qshellwidget neovim-qt)
+
 set_property(SOURCE main.cpp PROPERTY COMPILE_DEFINITIONS
 	NVIM_QT_RUNTIME="${CMAKE_INSTALL_FULL_DATADIR}/nvim-qt/runtime")
-add_executable(nvim-qt WIN32 main.cpp shell.cpp input.cpp errorwidget.cpp mainwindow.cpp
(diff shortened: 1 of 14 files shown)


equalsraf:
Starting to look good. Featurewise I'm stopping here. API changes to the plugin can go into https://github.com/equalsraf/neovim-gui-shim

equalsraf:
Main gripe, errors in ginit.vim don't show up.

equalsraf:
Leaving the issues with ginit.vim to a future PR. Will merge later, the highlights:

- ginit.vim is loaded on attachment, there are still a few issues (errors wont be visible). You can set GUI preferences there.
- you can remove the Guifont command definition from your init.vim
- g:GuiWindowId for #64
- `call GuiWindowMaximized(1)` and similar for fullscreen #83
- if the Gui* commands and functions are not available check your installation and settings, the nvim-qt runtime folder should be in &rtp

See :h neovim-gui-shim for more details

PR #118, opened 2016-04-18
In Unix map Windows key as D-

equalsraf:
Neovim now supports the D- modifer, which we already in Mac OS X
for the Command key. This commit enables a similar behaviour for the
Windows key. Qt represents the Windows key as being the Meta key.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -58,7 +58,7 @@ InputConv::InputConv() {
 QString InputConv::modPrefix(Qt::KeyboardModifiers mod)
 {
 	QString modprefix;
-#ifdef Q_OS_MAC
+#if defined(Q_OS_MAC) || defined(Q_OS_UNIX)
 	if ( mod & CmdModifier ) {
 		modprefix += "D-"; // like MacVim does
 	}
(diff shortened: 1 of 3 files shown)


teto:
I pulled this PR and it works only if I leave the mapping pressed for some time, if I press and release it, it doesn't trigger.

equalsraf:
Also pressing the Windows key in insert mode inserts `\<D-"\>'

justinmk:
> Also pressing the Windows key in insert mode inserts `<D-">`

Does it happen if you `inoremap` the chord?

    inoremap `<D-">` foo

teto:
In my case whatever letter I bind e.g., <D-a>, <D-b>:
`noremap <D-a> :echom 'hello world'<CR>`
in insert mode, just pressing Super is enough to trigger the binding,
no need to press a or b at the same time


2016-04-19 16:11 GMT+02:00 Justin M. Keyes <notifications@github.com>:

equalsraf:
When the windows key was pressed by itself it was not being ignored as it should, causing the \<D-"> , this should be fixed now.

teto:
It does fix it thanks !

PR #121, opened 2016-05-04
GUI: Add support for underline/curl colors

AdnoC:
Added support for the 'special' color field used for underlines/undercurl.
When the 'special' color isn't set, uses foreground color for underline.

The use of foreground color when special isn't set is consistent with MacVim's behavior.

There is a pull request (https://github.com/neovim/neovim/pull/4633) for neovim to add special colors back in.


equalsraf:
:+1:  @AdnoC thanks for fixing this here too.

PR #122, opened 2016-05-05
Include cpack

maorv:
Using cpack  I can create deb package for ubuntu/debian


equalsraf:
Merged. Thanks @maorv.

Issue #126, opened 2016-05-24
Set additional linespace below each line


equalsraf:
In gVim there was set linespace= to set an extra space the bottom of each line.

ping @khalidchawtany

khalidchawtany:
I always go back to **iTerm** because of this :(

equalsraf:
I've merged this into the PR into master the `GuiLinespace` command should be available now.

PR #127, opened 2016-05-24
GUI: Handle event Gui Linespace

equalsraf:
gVim and MacVim support the linespace option that sets a margin space below
each line (in pixels).

The GUI now handles a notification ['Gui', 'Linespace', 4] that sets this space
to a positive integer. This is achieved by adding the extra space as part of the
cell metrics.

For #126, this first commit only adds the event handling, we are still missing a command in the shim. For testing try
```vim
    :call rpcnotify(0, 'Gui', 'Linespace', 4)
```


khalidchawtany:
@equalsraf  I think **MacVim** and others put the space above the line not below it.

screen_shot_2016-05-25_at_12_32_33_am

Issue #128, opened 2016-05-25
When searching, matched string is not highlighted as I am typing (incsearch has been turned on)


JasonXJ:
With option "incsearch" being turned on, the matched string should be highlighted immediately as I am typing (before I type "enter" to finish the searching). This works in terminal neovim/vim or gvim, but not in nvim-qt.

Here is a screen shot of terminal neovim, gvim and nvim-qt when I typed "/world" (without "enter") in normal mode.

![selection_003](https://cloud.githubusercontent.com/assets/6949499/15528719/d5b0801a-2289-11e6-92ca-c0a11edc1778.png)

JasonXJ:
It seems that nvim-qt also has problems with background highlight in neovim's terminal emulator. Here is how to reproduce the problem:

1. Run an instance of nvim-qt  /  Run neovim in a terminal
2. use command `:terminal<cr>` to open a terminal emulator
3. Type `man ls` in the emulator
4. Type `/ls<cr>` to search the keyword "ls"

In the case of "neovim in a terminal", the keyword "ls" is background highlighted. However, in the case of "nvim-qt", nothing happened. Here is the screen shot:

![selection_004](https://cloud.githubusercontent.com/assets/6949499/15530516/2f5d1f52-2298-11e6-8d83-29b32b433e50.png)

equalsraf:
Confirmed here, this seems to be caused by our lack of support for the **reverse** attribute in the highlight_set event. Starting a PR shortly.

PR #129, opened 2016-05-25
GUI: Support the reverse attribute in highligh_set

equalsraf:
For #128.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -237,6 +237,12 @@ void Shell::handleHighlightSet(const QVariantMap& attrs)
 		m_hg_special = special();
 	}
 
+	if (attrs.contains("reverse")) {
+		auto tmp = m_hg_background;
+		m_hg_background = m_hg_foreground;
+		m_hg_foreground = tmp;
+	}
+
 	m_font_bold = attrs.value("bold").toBool();
 	m_font_italic = attrs.value("italic").toBool();
 	m_font_undercurl = attrs.value("undercurl").toBool();
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -148,18 +148,23 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 				// Draw "undercurl" at the bottom of the cell
 				if (cell.underline || cell.undercurl) {
 					QPen pen = QPen();
-					// Use specail color if it is set, otherwise use foreground color
-					if (cell.specialColor.isValid()) {
-						pen.setColor(cell.specialColor);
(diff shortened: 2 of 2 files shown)


JasonXJ:
Just tried it. Working great!

But I found a new problem. The color of underline is different in nvim-qt. Here is an example of `man ls` in the terminal emulator.

xx

equalsraf:
I cannot reproduce here but it could be related to #121. @AdnoC should the special color be used for underline? https://github.com/equalsraf/neovim-qt/pull/121/files#diff-4f97cc211d81b6e872ac37bae6f5c5e4R152

It seems plausible the special color could be set, but the underline should always use the foreground color right?

equalsraf:
Maybe this works for the underline issue

```diff
diff --git a/src/gui/shellwidget/shellwidget.cpp b/src/gui/shellwidget/shellwidget.cpp                              
index 0d51698..aaaf0b2 100644                                                                                       
--- a/src/gui/shellwidget/shellwidget.cpp                                                                           
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -148,18 +148,23 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
                                // Draw "undercurl" at the bottom of the cell
                                if (cell.underline || cell.undercurl) {
                                        QPen pen = QPen();
-                                       // Use specail color if it is set, otherwise use foreground color
-                                       if (cell.specialColor.isValid()) {
-                                               pen.setColor(cell.specialColor);
-                                       } else if (m_spColor.isValid()) {
-                                               pen.setColor(m_spColor);
-                                       } else if (cell.foregroundColor.isValid()) {
-                                               pen.setColor(cell.foregroundColor);
-                                       } else {
-                                               pen.setColor(m_fgColor);
-                                       }
                                        if (cell.undercurl) {
+                                               if (cell.specialColor.isValid()) {
+                                                       pen.setColor(cell.specialColor);
+                                               } else if (m_spColor.isValid()) {
+                                                       pen.setColor(m_spColor);
+                                               } else if (cell.foregroundColor.isValid()) {
+                                                       pen.setColor(cell.foregroundColor);
+                                               } else {
+                                                       pen.setColor(m_fgColor);
+                                               }
                                                pen.setStyle(Qt::DashDotDotLine);
+                                       } else if (cell.underline) {
+                                               if (cell.foregroundColor.isValid()) {
+                                                       pen.setColor(cell.foregroundColor);
+                                               } else {
+                                                       pen.setColor(m_fgColor);
+                                               }
                                        }
 
                                        // TODO: draw a proper undercurl
```

JasonXJ:
Yes, that works. By the way, if you would like to reproduce the problem yourself, you could first use `-u none` to disable settings and then use `:colorscheme desert` in nvim-qt.

Issue #140, opened 2016-06-14
why this unicode char cannot show correctly


wushuzh:
I try to use a special char in neovim. But it does not appear correctly on my env (arch linux).

I opened neovim-qt, input [this special char](http://www.fileformat.info/info/unicode/char/1f390/fontsupport.htm) by typing its unicode value in insert mode, type Ctrl+V, then 'U1f390', then back to normal mode.

I just install infinality-bundle-fonts/ttf-symbola-ib in linux.

Here is the screenshot in neovim-qt:

![neovim](https://cloud.githubusercontent.com/assets/3007906/16034826/74ef17c4-3246-11e6-8d6f-55204e3d31c1.png)

I also launch gvim and input the same, it can show wind bell char correctly.

![gvim](https://cloud.githubusercontent.com/assets/3007906/16034851/9873a610-3246-11e6-83aa-d7b0a3935cef.png)

There are a few chars (emoji) have the same issue.

If there is setting I should do to make it work ?

equalsraf:
> If there is setting I should do to make it work ? 

No, this doesn't seem to be working right now.

suy:
I see that the paint event iterates over Cell, which is backed by a QChar. This character is two surrogate pairs, so it can't work that easily. Maybe each Cell should store two QChars because to draw the text you will need potentially two. The on the paint event join the two in the QString used for drawText.

equalsraf:
@suy but shouldn't Neovim treat this a single utf8 char?

equalsraf:
Might be easier to replace the use of QChar with QString instead.

equalsraf:
Opened https://github.com/equalsraf/neovim-qt/pull/141 it draws the char correctly.

One noticeable difference is that gVim (and Vim) seems to treat this as a wide char, but Neovim does not, in the terminal this is a bit harder to test, but check column number in the ruler.

suy:
I have no idea how data gets into the Cell objects. :)  I just started looking at neovim-qt code, and I found the rendering, but I don't know how it reaches there. Using a QString for each Cell might be a bit memory fragmented (and wasteful) without small string optimization (which may arrive in Qt6, who knows). I would go with the two QChar if possible. Can't the whole line be in a single QString?

equalsraf:
@suy, yes I ended up following your suggestion, in #141 the cell now holds a second QChar for the lower surrogate.

> Can't the whole line be in a single QString?

It could but it would likely be worse when redrawing parts of the line.

[...]

PR #141, opened 2016-06-14
Fix unicode surrogate chars in shellwidget

equalsraf:
The shell widget is backed up by QChar, a 16bit char. For surrogate
chars this is problematic because we are painting char by char, but
painting a surrogate requires painting it as a whole.

To address this we allow storing the low surrogate inside the cell and
add Cell::text() that returns accordingly.

### DIFF
--- a/src/gui/shellwidget/cell.h
+++ b/src/gui/shellwidget/cell.h
@@ -45,6 +45,19 @@ public:
 		doubleWidth = (konsole_wcwidth(c.unicode()) > 1);
 	}
 
+	inline void setLowSurrogate(const QChar& chr) {
+		lowSurrogate = chr;
+		doubleWidth = (string_width(text()));
+	}
+
+	inline QString text() const {
+		if (c.isHighSurrogate() && lowSurrogate.isLowSurrogate()) {
+			return QString(c) + lowSurrogate;
+		} else {
+			return c;
+		}
+	}
+
 	static inline Cell invalid() {
 		Cell c = Cell('X', Qt::white, Qt::red, QColor(), false, false, false, false);
 		c.valid = false;
@@ -52,6 +65,7 @@ public:
 	}
 
 	QChar c;
+	QChar lowSurrogate;
 	QColor foregroundColor, backgroundColor, specialColor;
 	bool bold, italic, underline, undercurl;
 	bool valid;
--- a/src/gui/shellwidget/shellcontents.cpp
+++ b/src/gui/shellwidget/shellcontents.cpp
@@ -230,9 +230,18 @@ int ShellContents::put(const QString& str, int row, int column,
 	}
 	
 	int pos = column;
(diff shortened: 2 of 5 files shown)


equalsraf:
The biggest downside to this PR is that we run into rendering issues with some chars (e.g. for #140) when using a Neovim version earlier 0.1.5 - This is not a consequence of this PR, it happens because neovim and neovim-qt don't agree about the width of the characters.

I will hold the PR until 0.1.5 is released.

jamessan:
Is this ready to be revisited now?  Note, I also just merged the `'emoji'` related patches from Vim (c.f., neovim/neovim#5372) which may have an impact.

suy:
Hi.

I want to apologize for the stupid advice I gave on #140. It is true that for the given character in the issue one single QChar won't work, and two QChars would do, because it's a code point that needs two code units in QString. However, assuming that one code point is going to be displayed in one single glyph is wrong, and I don't know why I did not realize at that time. Here is one example of [one emoji that I get texted often](http://emojipedia.org/woman-facepalming/): 

```
$ uniscribe 🤦🏻♀

  1F926 ├─ 🤦            ├─ FACE PALM
  1F3FB ├─ 🏻            ├─ EMOJI MODIFIER FITZPATRICK TYPE-1-2
   200D ├─ ][           ├─ ZERO WIDTH JOINER
   2640 ├─ ♀            ├─ FEMALE SIGN
```
As you can see, that should be rendered as one single glyph, but it would fit in no less than 6 QChars (2 for face palm + 2 for the modifier + 1 for the ZWJ + 1 for the sign).

So, as you can see, I don't think adding more QChars is going to be a solution. I don't know if there is a theoretical limit for how many code points can be necessary for a single glyph.

I think the only proper solution for all the rendering issues is using a whole QString for the line,and letting QPainter do it's job of text shaping. Otherwise, is not going to be correct for languages [with complex text layout](https://en.wikipedia.org/wiki/Complex_text_layout).

equalsraf:
> I think the only proper solution for all the rendering issues is using a whole QString for the line,and letting QPainter do it's job of text shaping. Otherwise, is not going to be correct for languages with complex text layout.

I agree. This was after all the same approach used in vim-qt.

The main difference here is that the nvim API may ask us to place (**put**) a string at an arbitrary cursor position in the line - i.e. we still need a way to map chars (or substrings) to screen cells, so that later we can perform a substitution.

I guess  the correct back end storage would be

- 1 QString per line
- An array of cell indexes that maps cells to QString indexes (if this even makes sense)

On a side note this particular example renders as separate chars in nvim and vim, two double-width chars (🤦🏻 turns into two, one that is not rendered by my terminal) and one single width (♀).

PS: optimisations likely require QStringRef QString::midRef()

jgehrig:
Wide character rendering works now. We no longer use `QChar` within `Cell` either.

Marking as closed.

PR #143, opened 2016-06-15
Add Mac OS X bundle

equalsraf:
- [x] It still fails to run nvim on the system where I'm testing
- [ ] The bundle has an icon, shouldn't the file show the proper icon in finder?


khalidchawtany:
Hi @equalsraf,
I tried yout **tb-macos-bundle** branch and tried to fix the icon issue. It turns out that the problem is the icon file format. OS X apps require their icons to be in `*.icns` format.

I have [this working branch](https://github.com/khalidchawtany/neovim-qt/commit/52f024fe31af5010e054bc5d30b436c8107b4b5e). However, I have no Idea what I may have broken, as I am not a C++ developer.

Thanks.

equalsraf:
Merged in another PR.

PR #151, opened 2016-07-01
GUI: set termguicolors on spawn

### LINKED ISSUE #124: Colours are off for nvim-qt, working in terminal nvim
2016-05-18-122613_1366x768_scrot
2016-05-18-122634_1366x768_scrot

equalsraf:
When spawning Neovim, set termguicolors at startup to enable
true color support.

- Should have the same effect as https://github.com/equalsraf/neovim-qt/pull/101 but for upcoming Neovim 0.1.5.
- Partially fixes #124

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -85,6 +85,8 @@ int main(int argc, char **argv)
 
 	int sep = app.arguments().indexOf("--");
 	QStringList neovimArgs;
+	neovimArgs << "--cmd";
+	neovimArgs << "set termguicolors";
 	if (sep != -1) {
 		QStringList args = app.arguments().mid(0, sep);
 		neovimArgs += app.arguments().mid(sep+1);


justinmk:
This should not be necessary, because attaching an rgb-capable GUI should enable GUI colors automatically. see `ui_rgb_attached()`

justinmk:
`ui_attach` takes an `enable_rgb` parameter. Is that being passed by neovim-qt?

equalsraf:
> ui_attach takes an enable_rgb parameter. Is that being passed by neovim-qt?

Yes the last argument to ui_attach is set to true. I've checked that ui_attach is called and ui->rgb is set to true in Neovim.

However ui_rgb_attached() seems to be called before the GUI connects (even with --headless) during init.vim initialization as part of do_highlight()/syntax.c.

#124 in particular pointed to issues with airline, which is the one difference I see when using this PR. Maybe airline caches the outcome of that initialization? 

The same used to happen with NVIM_TUI_ENABLE_TRUE_COLOR in https://github.com/rogual/neovim-dot-app/issues/215#issuecomment-190531066.

justinmk:
Ok, understood. This comes back to the same old problem which will be addressed by [UIEnter/UIAttached](https://github.com/equalsraf/neovim-qt/issues/94#issuecomment-191285507).

equalsraf:
This would make us "incompatible" with neovim < 0.1.5, so I need a way to skip this option in earlier versions. Not sure if this can be done through a command in `--cmd` - something like a conditional with v:version (but for the neovim version).

justinmk:
We can add an api function, `nvim_get_version`.

equalsraf:
Merged into master, now that 0.1.5 is out.

Issue #148, opened 2016-06-27
nvim-qt won't open files passed in as command line arguments


turtleDev:
```
nvim-qt foo.txt
```

nvim just opens up a new buffer, instead of opening the file.  

For reference, I built nvim using fix in #138. I also get these error message when ever I start nvim-qt https://github.com/equalsraf/neovim-qt/issues/146#issuecomment-228262160

suy:
Open with `nvim-qt -- foo.txt`.

turtleDev:
Thanks @suy. Is this documented somewhere ? `neovim` can open files fine even without the `--`( end of options flag.)

suy:
I think it's the usual convention of terminating the options for one app (nvim-qt) and making it pass everything after the `--` to the other (nvim).

turtleDev:
that seems a bit odd to me. @suy

suy:
Why? It's pretty common in Unix. And mentioned in #125 or #106.

equalsraf:
I suppose its annoying enough to have popped up in the issue tracker a couple times. I wouldn't mind a patch for this provided that

- we can't enumerate all neovim CLI options/flags. Seems unlikely the `--` separator is going away to distinguish between options handled by nvim-qt and those handled by nvim
- (maybe) file opening is a special enough case that we can pass filename arguments can be forwarded to nvim as filenames

i.e. the current CLI arguments and nvim call would change from

```
nvim-qt [options] [-- ...]
nvim [...]
```
to this
```
nvim-qt [options] [filename...] [-- ...]
nvim [...] [filename...]
```

I'll need to think a bit more about this before starting a PR on it, But some additional toughts or a good PR on the subject would be most welcome.

whitelynx:
I'd almost rather have the `nvim-qt` options be the ones that appear _after_ the `--`, since most people would expect `nvim-qt` to be able to be called the same way as `nvim` or `gvim`. I'm not sure how difficult that would be to accomplish, though.

whitelynx:
Also, even though we can't enumerate all the neovim CLI options/flags, could we enumerate all the ones that neovim-qt parses instead? Then, whatever's left over could be passed to `nvim`.

equalsraf:
> I'd almost rather have the nvim-qt options be the ones that appear after the --, since most people would expect nvim-qt to be able to be called the same way as nvim or gvim.

[...]

Issue #125, opened 2016-05-24
Command line arguments are not passed to neovim on Windows


robinkjoy:
Tried release v0.2.0 and latest build from AppVeyor. File names or other options are not passed to neovim. In the case of options 'unknown option' pops up, persumably from nvim-qt, not nvim.

equalsraf:
@robinkjoy see `nvim-qt --help`, to pass arguments to neovim put them after `--` e.g.

```
nvim-qt -- file1 file2 file3
```

robinkjoy:
Oh right. Sorry, I didn't notice the space after --. Thank you!

PR #155, opened 2016-07-14
GUI: Accept filenames as positional arguments

equalsraf:
Previously when spawning Neovim only arguments after "--" were forwarded
to Neovim. nvim-qt now takes filenames as positional parameters that are
appended to the Neovim arguments.

The following are equivalent and open three files

    nvim-qt -- file0 file1 file2
    nvim-qt file0 file1 file2

The following is also equivalent, but notice the order,
because positional arguments are appended at the end of
neovim arguments

    nvim-qt file1 file2 -- file0

For #148 https://github.com/equalsraf/neovim-qt/issues/125


whitelynx:
I'd love to see this PR make it in... it'd make neovim-qt much simpler to use. (as it is, I just made a `qvim` script that calls `neovim-qt -- "$@"`)

turtleDev:
perfect +1

PR #158, opened 2016-07-19
Improve shim loading and attach notification

### LINKED ISSUE #156: The shim is not loaded soon enough
I've tried to install [a plugin for a different GUI](https://github.com/rhysd/nyaovim-popup-tooltip) to see if I could implement the GUI-side feature in Neovim-Qt, and understand a bit more the whole architecture. I've tried to fix the loading condition (before the `finish`) to check for `g:GuiLoaded`. It didn't work, and it is because the shim is loaded too late.

I did look at the code, and even if the command line arguments trick seems ugly (is there something better on the works on the server side?), to me that is supposed to work. The runtimepath is modified before the vimrc, so I don't what what's wrong. The shim is loaded in position 150, while the plugins are loaded before. Here is the order of `scriptnames` after plain startup: https://gist.github.com/suy/970fea1211ba01de57876a29f7c8035b

So, is there some fiddling with pathogen or the vimrc files that I'm not doing properly? My `init.vim` is a symbolic link to `vimrc`, and `ginit.vim` just contains a `GuiFont` command.

Thanks!

suy:
- Instead of calling `runtime plugin/nvim_gui_shim.vim` at the moment
  the GUI attaches to Neovim, do it as a "pre-seeded" command that runs
  before init.vim is loaded, but just after the runtimepath sees added
  the path to the shim. That way the user can check for the presence of
  g:GuiLoaded.
- Since g:GuiLoaded is not sufficient for all cases, trigger a User
  NeovimGuiAttached auto command at the moment the GUI attaches. That is
  appropriate for calling GuiFont and friends from an auto command
  handler in init.vim.

Fixes #156.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -97,6 +97,8 @@ int main(int argc, char **argv)
 							.arg(d.path()));
 				}
 			}
+			neovimArgs.insert(2, QStringLiteral("--cmd"));
+			neovimArgs.insert(3, QStringLiteral("runtime plugin/nvim_gui_shim.vim"));
 			c = NeovimQt::NeovimConnector::spawn(neovimArgs);
 		}
 	}
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -134,8 +134,8 @@ void Shell::setAttached(bool attached)
 		if (isWindow()) {
 			updateGuiWindowState(windowState());
 		}
-		m_nvim->neovimObject()->vim_command("runtime plugin/nvim_gui_shim.vim");
 		m_nvim->neovimObject()->vim_command("runtime! ginit.vim");
+		m_nvim->neovimObject()->vim_command(QByteArrayLiteral("doautocmd User NeovimGuiAttached"));
 	}
 	emit neovimAttached(attached);
 	update();


suy:
This makes possible doing something like the following in init.vim (or vimrc in my case, since I just keep a symlink):

```viml
if has('nvim') && exists('g:GuiLoaded')
	function! s:NeovimGuiSetup()
		GuiLinespace 2
		GuiFont DejaVu Sans Mono:h8
		call GuiWindowMaximized(1)
	endfunction
	augroup neovimguiattached
		autocmd!
		autocmd User NeovimGuiAttached call s:NeovimGuiSetup()
	augroup END
endif
```

Whether the check for `g:GuiLoaded` is necessary, is debatable. I think I like it like this, since I mostly want to have that only when neovim is spawned by neovim-qt, and not when I attach to a console version. I hope this isn't too intrusive. I've read on several issues about a future UIEnter event, but I don't see the harm in a user-defined one. After all, the Gui* variables/functions are user defined as well. I would not mind a standardised approach across GUIs, though.

equalsraf:
> Whether the check for g:GuiLoaded is necessary, is debatable. I think I like it like this, since I mostly want to have that only when neovim is spawned by neovim-qt, and not when I attach to a console version. I hope this isn't too intrusive.

I agree.

The tests are failing. Apparenly the shim has not loaded early enough

```
NeovimQt::Test::guiShimCommands() "GuiFont DejaVu Sans Mono:h14" "Vim:E492: Not an editor command: GuiFont DejaVu Sans Mono:h14"
```

I believe reverting the `--cmd` https://github.com/equalsraf/neovim-qt/commit/189a312ed332be228c5c068f0da4ab81b88eba85 fixes it.

suy:
I tried rebasing the change in a more recent master, and see if it still works as I wanted, but I just gave up. Since the preferred solution is upstream (neovim/neovim#6917) I'll just close this one.

PR #159, opened 2016-07-20
Support force font assignment

lambdalisue:
To support assignment an invalid font. See #103 for detail.

I did some lint refactoring on doc and vim script. I'll remove these commits if you don't like.


lambdalisue:
Thanks a lot!

PR #161, opened 2016-07-20
Fix #134

### LINKED ISSUE #134: Typing "<" gets encoded as <S-lt>
Reported by @khalidchawtany. In Mac OS X US layout `<` is inserted by typing `Shift`+`,` but instead of getting \<lt\> neovim receives a `<S-lt>`. This happens because in input.c we remove the SHIFT modifier later in the function i.e.
1. Removal of SHIFT happens after `<` is processed
2. `<` is handled twice, one time as part of the specialKeys and another as a special escape case, and this case needs to be handled before either
3. Presumably this could also happen in layouts where `\` is the result of pressing shift and another key

Its a bit strange that on a Mac Qt reports this event key: `<` with modifier `Shift` and text: `<`

Tentative patch

``` diff
diff --git a/src/gui/input.cpp b/src/gui/input.cpp
index b517c4b..97cb019 100644
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -150,6 +150,10 @@ QString InputConv::convertMouse(Qt::MouseButton bt, QEvent::Type type, Qt::Keybo
  */
 QString InputConv::convertKey(const QString& text, int k, Qt::KeyboardModifiers mod)
 {
+       if (text == "<" || k == Qt::Key_Less) {
+               mod &= ~ShiftModifier;
+       }
+
        if (specialKeys.contains(k)) {
                return QString("<%1%2>").arg(modPrefix(mod)).arg(specialKeys.value(k));
        }
```

lambdalisue:
> The modifiers holds the keyboard modifiers, and the given text is the Unicode text that the key generated.

http://doc.qt.io/qt-5/qkeyevent.html#QKeyEvent

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -151,16 +151,12 @@ QString InputConv::convertMouse(Qt::MouseButton bt, QEvent::Type type, Qt::Keybo
 QString InputConv::convertKey(const QString& text, int k, Qt::KeyboardModifiers mod)
 {
 	if (specialKeys.contains(k)) {
-		return QString("<%1%2>").arg(modPrefix(mod)).arg(specialKeys.value(k));
+		return QString("<%1>").arg(specialKeys.value(k));
 	}
 
 	QChar c;
 	// Escape < and backslash
-	if (text == "<") {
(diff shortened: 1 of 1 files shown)


lambdalisue:
Oops.

PR #168, opened 2016-08-10
GUI: Fix icon for Mac OS X application bundle.

khalidchawtany:
Mac OS X Apps require their icons to be in *.icns format. This pull request adds the necessary file and makes change to CMakeList.txt to so that the file is copied on build.

NOTE: I am not a C++ developer and I googled to use git. So I don't know what I may have broken :)


equalsraf:
Cheers, thanks for having a look at this.

PR #169, opened 2016-08-11
GUI: Handle open file requests.

khalidchawtany:
In Mac OS X, this enables the application to handle
requests to open files. This can be through double
clicking a file in the finder, dropping a file on the
application icon (bundle), or dropping a file on the
application's icon in the dock.

This does not handle dropping directories on the app.
icon in the finder or the Dock. Since, I did not know
what should happen then.

I am a noob. I definitely have made many mistakes in this pr 😉

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -1,7 +1,6 @@
 #include "app.h"
 
 #include <QFileOpenEvent>
-#include <QMessageBox>
 
 namespace NeovimQt {
 
@@ -12,18 +11,13 @@ App::App(int &argc, char ** argv)
 
 bool App::event(QEvent *event)
 {
-  if( event->type()  == QEvent::FileOpen) {
-    //QFileOpenEvent is triggered for each file|URL seperately. Kepping
-    //the QList<QUrl> here does not cause any file|URL to be ignored.
-    QList<QUrl> urls;
-    QFileOpenEvent * fileOpenEvent = static_cast<QFileOpenEvent *>(event);
-    if(fileOpenEvent) {
-      QString file = fileOpenEvent->file();
-      urls.append(fileOpenEvent->url());
-      emit openFilesTriggered(urls);
-    }
-  }
-  return QApplication::event(event);
+	if( event->type()  == QEvent::FileOpen) {
+		QFileOpenEvent * fileOpenEvent = static_cast<QFileOpenEvent *>(event);
+		if(fileOpenEvent) {
+			emit openFilesTriggered({fileOpenEvent->url()});
+		}
+	}
+	return QApplication::event(event);
 }
 
 } // Namespace
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -1,6 +1,7 @@
 #include "app.h"
 
 #include <QFileOpenEvent>
+#include <QMessageBox>
 
 namespace NeovimQt {
 
@@ -12,13 +13,14 @@ App::App(int &argc, char ** argv)
 bool App::event(QEvent *event)
 {
   if( event->type()  == QEvent::FileOpen) {
+    //QFileOpenEvent is triggered for each file|URL seperately. Kepping
+    //the QList<QUrl> here does not cause any file|URL to be ignored.
+    QList<QUrl> urls;
     QFileOpenEvent * fileOpenEvent = static_cast<QFileOpenEvent *>(event);
     if(fileOpenEvent) {
       QString file = fileOpenEvent->file();
-      if(!file.isEmpty()) {
-        emit neovimOpenFileTriggered(file);
-        return true;
-      }
+      urls.append(fileOpenEvent->url());
+      emit openFilesTriggered(urls);
     }
   }
   return QApplication::event(event);
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -1,0 +1,27 @@
+#include "app.h"
+
+#include <QFileOpenEvent>
+
+namespace NeovimQt {
+
+App::App(int &argc, char ** argv)
+:QApplication(argc, argv)
+{
+}
+
+bool App::event(QEvent *event)
+{
+  if( event->type()  == QEvent::FileOpen) {
+    QFileOpenEvent * fileOpenEvent = static_cast<QFileOpenEvent *>(event);
+    if(fileOpenEvent) {
+      QString file = fileOpenEvent->file();
+      if(!file.isEmpty()) {
+        emit neovimOpenFileTriggered(file);
+        return true;
+      }
+    }
+  }
+  return QApplication::event(event);
+}
+
+} // Namespace
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim
@@ -64,4 +64,5 @@ function GuiDrop(...)
 	let l:fnames = deepcopy(a:000)
 	let l:args = map(l:fnames, 'fnameescape(v:val)')
 	exec 'drop '.join(l:args, ' ')
+	doautocmd BufEnter
 endfunction
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -138,11 +138,11 @@ void Shell::setAttached(bool attached)
 		m_nvim->neovimObject()->vim_command("runtime plugin/nvim_gui_shim.vim");
 		m_nvim->neovimObject()->vim_command("runtime! ginit.vim");
 
-    // Noevim was not able to open urls till now. Check if we have any to open.
-    if(!m_urls.isEmpty()){
-      openFiles(m_urls);
-      m_urls.clear();    //Neovim may change state. Clear to prevent reopening.
-    }
+		// Noevim was not able to open urls till now. Check if we have any to open.
+		if(!m_urls.isEmpty()){
+			openFiles(m_urls);
+			m_urls.clear();    //Neovim may change state. Clear to prevent reopening.
+		}
 
 	}
 	emit neovimAttached(attached);
@@ -979,11 +979,11 @@ void Shell::openFiles(QList<QUrl> urls)
 		}
 		m_nvim->neovimObject()->vim_call_function("GuiDrop", args);
 	} else {
-    // Neovim cannot open urls now. Store them to open later.
+		// Neovim cannot open urls now. Store them to open later.
 		foreach(QUrl u, urls) {
-      m_urls.append(u);
-    }
-  }
+			m_urls.append(u);
+		}
+	}
 }
 
 } // Namespace
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -139,9 +139,9 @@ void Shell::setAttached(bool attached)
 		m_nvim->neovimObject()->vim_command("runtime! ginit.vim");
 
 		// Noevim was not able to open urls till now. Check if we have any to open.
-		if(!m_urls.isEmpty()){
-			openFiles(m_urls);
-			m_urls.clear();    //Neovim may change state. Clear to prevent reopening.
+		if(!m_deferredOpen.isEmpty()){
+			openFiles(m_deferredOpen);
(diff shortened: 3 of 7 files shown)


justinmk:
Directories should be treated like files.

src/gui/shell.cpp
```diff
@@ -117,6 +117,11 @@ bool Shell::setGuiFont(const QString& fdesc, bool force)
 	return ok;
 }
 
+void Shell::neovimOpenFile(const QString &file)
+{
+  m_nvim->neovimObject()->vim_command(QByteArray::fromStdString("e " + file.toStdString()));
+}
```
equalsraf: Add a check for neovim connection before using neovimObject

'''
if (!m_nvim || !m_nvim->neovimObject()) {
		return;
}
'''

equalsraf:
It seems probable we can run into race conditions when this is used with 'open' from the console. What should happen if the notification arrives but the connection to neovim is not yet established?

equalsraf:
In https://github.com/equalsraf/neovim-qt/pull/152 i was using drop() instead of :e (and fnameescape for the path). What would be the expected behaviour in this case?

khalidchawtany:
@justinmk What should happen when there are more than one directories dropped on the app icon. For files it is OK, as the files can be seen in the `:ls`. However, for directories I don't know what should happen.

@equalsraf I though about that too. However, I did not want to break more stuff.  Does it make sense to kept the list of passed files till a connection is established.

khalidchawtany:
I did not see https://github.com/equalsraf/neovim-qt/pull/152. Should I use `:drop` instead of `:e`? I did not know about drop either :(

equalsraf:
> @justinmk What should happen when there are more than one directories dropped on the app icon. For files it is OK, as the files can be seen in the :ls. However, for directories I don't know what should happen.

They should appear as any file. After all `:e /` is perfectly valid.

> @equalsraf I though about that too. However, I did not want to break more stuff. Does it make sense to kept the list of passed files till a connection is established.

That would be neat. But its not a blocker for this PR.

> I did not see #152. Should I use :drop instead of :e? I did not know about drop either :( 

I don't know. Best bet here is to compare with the behavior of MacVim to determine if :e or :drop makes more sense. A couple cases are

- Using [open](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/open.1.html), might want to play with `-F` too
- opening from finder

equalsraf:
> This is what I did to open files from terminal using open:
> open -a nvim-qt.app --args ~/Desktop/info.txt

What about without `--args` just `open -a nvim-qt.app ~/Desktop/info.txt`

khalidchawtany:
You are right without `--args` the event is triggered. However, then nothing else handles the file open event. So the files is not opened in **nvim-qt.app**.

equalsraf:
> You are right without --args the event is triggered. However, then nothing else handles the file open event. So the file is not opened in nvim-qt.app.

Yay, what happens if you try opening a URL with open? 

QFileOpenEvent does have a url() method but I'm not sure if it works for non files.

khalidchawtany:
Dropping directories on the application icon (bundle) is handled using the URL method. For a directory `/Volumes/Home/Desktop/temp` the `url().toString()` has `file:///Volumes/Home/Desktop/temp` in it/

khalidchawtany:
@equalsraf MacVim handles opening files using `:drop`, unless the it's configured to do it another way.

equalsraf:
> This PR could very much benefit from #152 and reuse the Shell::openFiles.

Merged. Let me know if you run into issues. `openFiles()` transforms file: URLs into local paths, and leaves other URLs unchanged for netrw, but I'm not 100% this is working.

khalidchawtany:
@equalsraf this enables neovim-qt to delay opening files till neovim is attached.

src/gui/app.cpp
```diff
@@ -12,13 +13,14 @@ App::App(int &argc, char ** argv)
 bool App::event(QEvent *event)
 {
   if( event->type()  == QEvent::FileOpen) {
+    //QFileOpenEvent is triggered for each file|URL seperately. Kepping
+    //the QList<QUrl> here does not cause any file|URL to be ignored.
+    QList<QUrl> urls;
     QFileOpenEvent * fileOpenEvent = static_cast<QFileOpenEvent *>(event);
     if(fileOpenEvent) {
       QString file = fileOpenEvent->file();
```
equalsraf: `file` is no longer needed right?

src/gui/shell.cpp
```diff
@@ -146,6 +137,13 @@ void Shell::setAttached(bool attached)
 		}
 		m_nvim->neovimObject()->vim_command("runtime plugin/nvim_gui_shim.vim");
 		m_nvim->neovimObject()->vim_command("runtime! ginit.vim");
+
```
equalsraf: indentation is broken in this block

src/gui/shell.cpp
```diff
@@ -980,7 +978,12 @@ void Shell::openFiles(QList<QUrl> urls)
 			}
 		}
 		m_nvim->neovimObject()->vim_call_function("GuiDrop", args);
-	}
+	} else {
```
equalsraf: indentation broken on this block

khalidchawtany:
File is no longer needed. This also handles directories too. 

As for the indentation issues. In neovim they look the same indentation level. I don't know why they are broken. I will copy paste leading space from other lines (As I did before ).

src/gui/app.cpp
```diff
@@ -12,13 +13,14 @@ App::App(int &argc, char ** argv)
...
-      if(!file.isEmpty()) {
-        emit neovimOpenFileTriggered(file);
-        return true;
-      }
+      urls.append(fileOpenEvent->url());
+      emit openFilesTriggered(urls);
```
equalsraf: You should be able to replace the urls variable above and these two lines with
```C++
emit openFilesTriggered(QList<QUrl>() << fileOpenEvent->url());
```

or (c++11)

```C++
emit openFilesTriggered({fileOpenEvent->url()});
```

equalsraf:
> As for the indentation issues. In neovim they look the same indentation level. I don't know why they are > broken. I will copy paste leading space from other lines (As I did before ).

Some of the lines, but not all, are using spaces instead of tabs. Replacing spaces with tabs and aligning should fix it.

khalidchawtany:
I had `set expandtab`  and `set nolist`, so everything was "fine".

src/gui/shell.cpp
```diff
@@ -971,6 +978,11 @@ void Shell::openFiles(QList<QUrl> urls)
 			}
 		}
 		m_nvim->neovimObject()->vim_call_function("GuiDrop", args);
+	} else {
+		// Neovim cannot open urls now. Store them to open later.
```
equalsraf: m_urls.append(urls);

src/gui/shell.h
```diff
@@ -107,6 +107,8 @@ private slots:
 
 	NeovimConnector *m_nvim;
 
+	QList<QUrl> m_urls;
+
```
equalsraf: I would prefer this variable to have a more explicit name. But I don't have better alternatives e.g. `m_deferredOpen`

equalsraf:
@khalidchawtany after those two comments I'm good with this. I don't think we can easily write unit tests for this feature anyway, and I can't test locally since I don't have a Mac.

I'll wait for your thumbs up to merge this PR.

khalidchawtany:
@equalsraf this might work on other operating systems. How can I test this on a PC?

equalsraf:
Interesting, it doesn't seem to work for dirs, but I'm not certain that is the case, e.g.

1. Open nvim-qt. At this point `:ls` shows 1 No Name buffer
2. Drop a folder in the UI, for me this prints `"~/bin" is a directory` and the buffer is empty, `:ls` now holds a second buffer with the name of the path
3. Call `:bnext` and `:bprev` to switch buffers and back. Now the buffer shows the directory contents. `:ls` shows a single buffer.

equalsraf:
**Update:** my previous example requires `set hidden` other wise you need to create an empty buffer explicitly.

khalidchawtany:
My problem is similar (May be the same).
1- Drag a `dir` on to the app icon.
2- A message is shown `"dir_path" is a directory`.
3- Now `:ls` shows "only" one buffer named `dir_path`. 
4- Execute `:enew` (as `bnext` and `bprevious` require more than one buffer).
5- `:bprevious` switches back to prev. buffer and shows the correct directory listing.

khalidchawtany:
@equalsraf I changed the `GuiDrop` to the following just to see what is going on. It seems at the time a `dir` is sent to be opened the &rtp is not set and `scriptnames` returns nothing.

Opening dirs requires the some autocommands to show the directory listing, unlike opening files!
```vim
function GuiDrop(...)
	let l:fnames = deepcopy(a:000)
	let l:args = map(l:fnames, 'fnameescape(v:val)')
	let @r=execute('echo &runtimepath')
	let @s=execute('scriptnames')
	let @a= execute('set verbose=10 | drop '.join(l:args, ' ') . ' | set verbose=0')
endfunction
```

justinmk:
Yes, opening directories requires autocmd so netrw can handle it. If plugins aren't enabled the correct behavior is an empty buffer whose name is the directory path.

khalidchawtany:
@equalsraf do you have any ideas? What could be causing this problem?
I don't think this has anything to do with this PR. I beleive that the problem is due to how sent messages are handled by neovim (or how nvim-qt sends the message).

equalsraf:
No ideas, just wondering if there is something :GuiDrop could call to force the update

khalidchawtany:
@equalsraf We can force the directory listing through the following "hacky" solution:
```c
void Shell::openFiles(QList<QUrl> urls)
{
...
		m_nvim->neovimObject()->vim_call_function("GuiDrop", args);
		m_nvim->neovimObject()->vim_feedkeys(":e%\r", "n", false);
...
}
```
We are feeding `<c-\><c-n>:e%<cr>`, to escape other modes (eg. insert mode). This makes sense, since if you are opening a file, you will be out of insert mode. The above works for both files and dirs. Even if a file is modified in nvim-qt and it is dragged to the GUI again, the changes are not lost (Just the way vim works).

For some reason changing GuiDrop to the following did not work! (the `<cr>` was not interpreted as `<Enter>`)
```vim
function GuiDrop(...)
	let l:fnames = deepcopy(a:000)
	let l:args = map(l:fnames, 'fnameescape(v:val)')
	exec 'drop '.join(l:args, ' ')
	call feedkeys(":e%\<CR>", "n")
endfunction
```

khalidchawtany:
We can wrap the feedkeys section with a check to see if the last argument is a directory.
```C
		if(QFileInfo(args.last().toString()).isDir()) {
			m_nvim->neovimObject()->vim_feedkeys(":e%\r", "n", false);
		}
```

justinmk:
Try `:doautocmd BufEnter`.

khalidchawtany:
@justinmk thanks a lot. I have being throwing everything at this issue :)

`:doautocmd BufEnter` works flawlessly.

equalsraf:
> @justinmk thanks a lot. I have being throwing everything at this issue :)

:D @khalidchawtany are we missing anything here, or can I merge this?

khalidchawtany:
Yep, unless I have missed something :smile:

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -64,4 +64,5 @@ function GuiDrop(...)
 	let l:fnames = deepcopy(a:000)
 	let l:args = map(l:fnames, 'fnameescape(v:val)')
 	exec 'drop '.join(l:args, ' ')
+	doautocmd BufEnter
```
justinmk: FWIW, this is no longer required on master, it was a 30-year-old Vim bug that [was fixed](https://github.com/neovim/neovim/commit/42c922b32c0a22fbe078d03bac5de0cfa7bd0b9f).

It's harmless to leave it here, though it will cause a redundant BufEnter event now. Could wrap it in a `has('nvim-0.2')` check.

Issue #164, opened 2016-07-31
Problem with GuiFont and NerdFonts and Powerline


TheAifam5:
**OS**: ArchLinux
**Version**: git master
**Fonts**: https://github.com/ryanoasis/nerd-fonts and PowerlineFonts
**ginit.vim**: `Guifont MesloLGLDZ NF RegularForPowerline:h13`
**Error**: `Unknown font: MesloLGLDZ NF RegularForPowerline`

How to set Powerline fonts or NerdFonts?
On Gnome-Terminal fonts works.

equalsraf:
Lots of fonts, not entirely sure what is the right variation you wanted. I used this one 

```
./unpatched-sample-fonts/Meslo/L-DZ/Meslo LG L DZ Regular for Powerline.otf
```

The name for this one is `Meslo LG L DZ for Powerline`. The `Regular` in the name is just the font style but it is usually not part of the font name. A good way to find this in linux systems to call `fc-list` it should show all the font files in your system, followed by their name and some more attributes.

Jeffrey04:
but how to set the font? i typed

```
:set guifont
```

and get

```
E519: Option not supported: guifont
```

equalsraf:
> but how to set the font?

Instead of `:set guifont=font` use `:GuiFont font`, for example `:Guifont DejaVu Sans Mono:h16` see `:h GuiFont` for more options.

kierun:
I got a similar problem (shocker, right? ;). This is what I have in my `ginit.vim`:

`GuiFont Inconsolata for Powerline Nerd Font Complete:h8`

Now, the font is installed:

```
; ls .local/share/fonts/Inconsolata\ for\ Powerline\ Nerd\ Font\ Complete*
896K .local/share/fonts/Inconsolata\ for\ Powerline\ Nerd\ Font\ Complete.otf
1.2M .local/share/fonts/Inconsolata\ for\ Powerline\ Nerd\ Font\ Complete\ Mono.otf
1.2M .local/share/fonts/Inconsolata\ for\ Powerline\ Nerd\ Font\ Complete\ Mono\ Windows\Compatible.otf
896K .local/share/fonts/Inconsolata\ for\ Powerline\ Nerd\ Font\ Complete\ Windows\ Compatible.otf
```

So the fonts are there.  Yet, I get an `Unknown font` error. Any idea why?

rafaelndev:
@kierun try
`fc-list | grep Inconsolata`
[...]

Issue #154, opened 2016-07-11
Itallic fonts are clipped


equalsraf:
Reported by @khalidchawtany, italic text in some fonts is clipped

- Menlo
- Courier
- Source Code Pro
- Firea Code

![](https://files.gitter.im/equalsraf/neovim-qt/g35L/Screen-Shot-2016-07-10-at-7.23.40-PM.png)

Presumably this is expected behavior from Qt ([from SO](http://stackoverflow.com/questions/23737591/qpainter-drawtext-italic-writes-outside-of-the-rectangle)) a result of left/right bearing. This is a problem because we can't change the font metrics between italics and non italics (they must be the same width).

We can adjust the drawing function, but its still unclear if this would fix the issue or cause a similar issue on the right side.

equalsraf:
It would seem I was wrong here. Source Code Pro reports positive left and right bearings.

equalsraf:
I retract my previous statement, other fonts do report negative bearings, maybe there is something to be done here e.g.

```diff
index a02a1f3..9d24084 100644                                                                 
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -153,6 +153,18 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 
                                        // Draw chars at the baseline
                                        QPoint pos(r.left(), r.top()+m_ascent+m_lineSpace);
+
+                                       // Adjust according to font bearing
+                                       int rBearing = p.fontMetrics().rightBearing(cell.c);
+                                       int lBearing = p.fontMetrics().leftBearing(cell.c);
+                                       if (lBearing + rBearing < 0) {
+                                               // Not much we can do, the font does not
+                                               // fit the cell
+                                       } else if (lBearing < 0) {
+                                               pos.setX(pos.x()-lBearing);
+                                       } else if (rBearing < 0 ) {
+                                               pos.setX(pos.x()+rBearing);
+                                       }
                                        p.drawText(pos, QString(cell.c));
                                }
 
~
~
~
```

ping @khalidchawtany

khalidchawtany:
I will try it in a moment. Thanks for investigating.

khalidchawtany:
To be honest I don't see any difference in rendering.
Before:
-----------
![before](https://cloud.githubusercontent.com/assets/1681606/17811964/b93320a4-662d-11e6-81e8-c983e28fa6ff.png)


After:
---------------
[...]

PR #171, opened 2016-08-19
Fix cropped and double width chars.

khalidchawtany:
This PR solves two issues I had with neovim-qt.
----
1- Fixes cropped chars.
2- Fixes an issue where only first cell of double width chars were drawn.

Compare the Before and after photos.
Before:
----------------
cropped

After:
---------
fixed

Now I have two issues.
------------
1- When the cursor moves over the chars, the update call forces the chars to be drawn as they were before this PR.
2- I see a little `x` in the second cell of double width chars. How can I remove it.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -124,6 +124,10 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 				QRect r = absoluteShellRect(i, j, 1, chars);
 
 				if (j <= 0 || !contents().constValue(i, j-1).doubleWidth) {
+
+					// If parts of previous character were drawn in our rect, keep it.
+					p.setCompositionMode(QPainter::CompositionMode_DestinationOver);
+
 					// Only paint bg/fg if this is not the second cell
 					// of a wide char
 					if (cell.backgroundColor.isValid()) {
@@ -131,6 +135,9 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 					} else {
 						p.fillRect(r, m_bgColor);
 					}
+
+					// Put new char over the bg ( possible previous chars parts )
+					p.setCompositionMode(QPainter::CompositionMode_SourceOver);
 
 					if (cell.c == ' ') {
 						continue;


equalsraf:
This seems similar to #141 and in this case I'm not sure keeping the cell contents is correct.

khalidchawtany:
What could be the implications of keeping the cell content?

I tried to play with font left and right bearings, and they seem impractical. There is also the possiblity of drawing using drawStaticText for each line.

equalsraf:
> What could be the implications of keeping the cell content?

A bad font could spill out of the cell, e.g. like the ones in #164.

If Neovim fails to detect a char as being double width (its a bug, but can happen) then we would draw over the following char.

Double width characters should already be handled properly (except for those covered in #114) by drawing across two cells. Does your example also happen with #114?

> drawing using drawStaticText for each line.

This would require some deep changes in shellwidget.

khalidchawtany:
>Does your example also happen with #114?

The screenshot is for the latest master. I just compiled it.

>A bad font could spill out of the cell, e.g. like the ones in #164.

I don't think it is your issue if others have bad fonts :smile:

khalidchawtany:
@equalsraf  the current character has the priority over previous one. So the prev one cannot be overriding current one. However, the prev. char can still make the current unreadable ;(

equalsraf:
> When the cursor moves over the chars, the update call forces the chars to be drawn as they were before this PR

Because the cursor is placed on the cell, and the cell is not detected as double width, #114

> I see a little x in the second cell of double width chars. How can I remove it.

The little 'x' usually de fallback drawing for invalid cells i.e. something you shouldn't be drawing on the screen.

khalidchawtany:
Thanks for explaining. I really appreciate your efforts, I wish I could help :(

equalsraf:
> I don't think it is your issue if others have bad fonts :smile: 

I agree, but I prefer that users can immediately tell if the font is "bad", "good" or if its a bug on our part.

There are a couple separate issues here

- The double width chars being drawn as a single cell, for most cases I hope its fixed by #114 
- The italics, which maybe could be handled by changing the clipping like you do here, and possibly redrawing adjacent chars as well

khalidchawtany:
@equalsraf what if we keep only as much as the right bearing of the previous character and discard the rest? This makes it impossible for previous char to corrupt the current one.

Does it make sense? Or we should not be keeping any.

equalsraf:
> Does it make sense? Or we should not be keeping any and that is the END.

You mean for italics? I've added a comment in https://github.com/equalsraf/neovim-qt/issues/154#issuecomment-241021350 but it tries to shift the painting instead of allowing it to spill.

equalsraf:
In more detail.

In don't we should try to keep any, but I also think that is not the end of the issue.

In my mind monospace fonts should not spill onto the next cell, but maybe my assumption is wrong. If I am then I would need to figure out which cells need to be updated in addition to the cell being pained, knowing that

- cells can be painted in any order, because the cursor can move to any position. So if I am called to paint a cell I may need to paint the previous and following cell as well
- however the background and foreground color in the current/previous/next cells can be different
- I would really like to avoid the worst case when painting a cell causes the full line to be painted - this is painfully expensive

Side thought, instead of keeping the content of the previous cell you can redraw it clipped to the current cell, expensive though.

NoAnyLove:
Are we going to fix the cropped font, or just leave it cropped?

equalsraf:
I would like a fix, if anyone is giving them away I will take it :)

kierun:
@zgpio Since the bug is still open, it has not been fixed.

PR #177, opened 2016-08-31
Implement mousehide

equalsraf:
Ping @lambdalisue


lambdalisue:
Sorry for late reply. It seems wonderful. Thanks for implementation.

justinmk:
For these options where restoring them is low-impact, I'd be in favor of adding them to core Nvim.

Issue #184, opened 2016-09-14
Lots of Unsupported function attribute… calls


kierun:
When I run`nvim-qt`, I get lots of `Unsupported function attribute…` warning but `neovim-qt` seems to run just fine. This is on a CentOS 7 machine. Here is the fuill log: [log.txt](https://github.com/equalsraf/neovim-qt/files/472097/log.txt)

Do you need anything else for me?

equalsraf:
> When I runnvim-qt, I get lots of Unsupported function attribute… warning but neovim-qt seems to run just fine. This is on a CentOS 7 machine.

Those messages only appear in debug builds, and are a reminder that some functions or attributes are not handled by the bindings, but they harmless. If you build in release mode they should not appear.

The increase in messages happened because of recent changes in the Neovim API, but there was no compatibility issues so it still runs as intended. We should probably fix all these before Neovim 0.1.6 comes out.

> Do you need anything else for me?

Nope, thanks anyway.

kierun:
Hum, I thought I was building in release mode or at least I am doing what the `README.rst` file says.  Is there something else I should be doing?

equalsraf:
Remove the build/ folder and restart with `cmake -DCMAKE_BUILD_TYPE=Release ..`

kierun:
Did that (`cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME ..`) but I still get the warnings.

```
(master|✓); rm -rf ./build 
(master|✓); mkdir build
(master|✓); cd build 
(master|✓); cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME ..
-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Downloading Msgpack...
-- Found GTest: /usr/lib64/libgtest.so  
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.7") 
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
[...]

PR #185, opened 2016-09-14
Cleanup after Neovim API change

equalsraf:
Recent changes in the Neovim API cause nvim-qt to print a lot of warning messages due to the new attributes

- don't print attribute warnings in non debug builds
- add support for the new attributes in the binding generator and in function.cpp

To retain backwards compatibility this PR DOES NOT update the bindings.

ref https://github.com/equalsraf/neovim-qt/issues/184


equalsraf:
Will reopen, for now mostly covered by #188

PR #193, opened 2016-10-01
generate helptags

### LINKED ISSUE #190: helptags aren't generated for documentation
When installing the documentation, helptags should be generated along with them.  Without that, it's not easy for user's to actually view the docs.

jpleau:
This is to fix #190 

This commit generates helptags (using nvim). Running "make install" installs the tags file at the right location.

I'm open to suggestions if how I did it can be improved (I'm far from a CMake expert!)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .*
 build/
+src/gui/runtime/doc/tags
 third-party/msgpack*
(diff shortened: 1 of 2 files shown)


equalsraf:
@jpleau thanks for this.

Can you add an entry to .gitignore, for src/gui/runtime/doc/tags

jpleau:
Done!

jpleau:
Hm it seems that CI don't work anymore though, 'nvim' is not a command.

Should we add a test to only generate tags NEOVIM_EXEC exists ?

equalsraf:
> Hm it seems that CI don't work anymore though, 'nvim' is not a command.

- The windows build does not have nvim installed yet, I've been meaning to fix that
- Can you rebase on top of master, I think that should fix the other CI issues.

equalsraf:
Moved to #194.

PR #194, opened 2016-10-02
Appveyor: Install Neovim nightly nvim and enable tests

equalsraf:


equalsraf:
The tests will be continued in a new PR, merging for the tags.

Issue #198, opened 2016-10-12
flicker during startup


zhou13:
I found that during the startup of neovim-qt, it will flicker:

1.  It seems that neovim-qt shows the window before it loads "ginit.vim".  So I can even see the change of font,

2.  It seems that neovim-qt changes its windows size after the window shows up. I can also see the resize process.

Is it possible to resize the windows and change the font before the window of neovim-qt shows up?

nash900:
I see flicker issue with short time of part black windows at startup, see screenshot:
![black_window_at_nvim-qt_startup](https://user-images.githubusercontent.com/8497245/75346078-e1a3a580-58d8-11ea-89a0-671ac24be97b.png)

PR #212, opened 2016-12-08
"Flicker Free" startup

Dax89:
This pull request fixes some flickering issues during startup sequence when a custom font is selected (related to https://github.com/equalsraf/neovim-qt/issues/198).

The window is displayed after the first resizing sequence is completed, I have tested that and it doesn't slowdown the startup sequence.


zhaozg:
What progress?

@equalsraf How do you think about this

jgehrig:
Flicker will be fixed by #866

PR #226, opened 2017-01-26
Added link to README for shim plugin docs.

rdnetto:
This PR adds a section to the README on how to change the font, since it's not immediately apparent.

On a related note, the shim plugin isn't loaded when following the build instructions, because `NVIM_QT_RUNTIME_PATH` is hard-coded to `/usr/local/share/nvim-qt/runtime` (which only works if Neovim-QT has been installed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ for Windows both MSVC and Mingw are supported. Use the following
 
 The binaries will be under build/bin/. The GUI binary is called nvim-qt.
 
+## Configuration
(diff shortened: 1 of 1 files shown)


equalsraf:
> (which only works if Neovim-QT has been installed).

True. You can influence this with an environment variable $NVIM_QT_RUNTIME_PATH, or using regular Neovim settings.

https://github.com/equalsraf/neovim-qt/blob/master/src/gui/app.cpp#L179

Issue #222, opened 2017-01-07
Broken keymap


major-kolz:
I set keyboard layout on my desktop environment (KDE) to 'Programmer Dvorak'. I use my custom keymap for enabling Russian JCUKEN layout ([prog-dvorak.txt](https://github.com/equalsraf/neovim-qt/files/691416/prog-dvorak.txt) attached)

Rules for: 'Ü', '<' and '0'-'9' doesn't apply. In Neovim itself all it work.

Also, I get space inserted when I do "Ctrl+Shift" (shift+@ is "^" for my layout). 

---
OS: Kubuntu 16.10
Compiler: gcc version 6.2.0 20161005 (Ubuntu 6.2.0-5ubuntu12)
Qt: 5.6.1
Neovim: NVIM 0.2.0-dev (Build type: RelWithDebInfo)

major-kolz:
Figured it out!

Problem was in Shift-prefix. Programmer Dvorak move digits to upper layer — so you get `<S-7>` instead of `<7>` for example. 

I change _gui/input.cpp_ (line 184) to solve it for digits and 'Ü':
```c++
if (c.unicode() < 0x100 && !c.isLetterOrNumber() && c.isPrint()) {  // before
if (c.unicode() < 0x100 && c.isPrint()) {  // after
```
---
Rule for '<' doesn't applied also because of Shift. I do not know how to deal with it in _gui/input.cpp_ (_text_ parameter is empty and _k_ set to  '16777248'). So, add explicitly check for it in _gui/shell.cpp_ ('keyPressEvent' method, line 586):
```c++
if (m_nvim->encode(inp) == "<S-lt>")
        inp = "<lt>";
```

---

Space insertion when Ctrl+Shift. Similarly to #199. _gui/input.cpp_ (line 170):
```c++
} else if (k == Qt::Key_Shift && mod&ControlModifier) {
    return QString();
```


[patch.zip](https://github.com/equalsraf/neovim-qt/files/753046/patch.zip)

major-kolz:
Had issue with keypad's digits: it was handled as usual keys, so Programmer Dvorak translated it.
I add to  _modPrefix_ function in _gui/input.cpp_ next code:
```c++
if ( mod & Qt::KeypadModifier ) {
    modprefix += "k";
}
```

artkpv:
@major-kolz Is it this the related issue? https://github.com/equalsraf/neovim-qt/issues/608

jgehrig:
[...]

PR #233, opened 2017-02-11
Problem with Programmer Dvorak keymap (#222)

major-kolz:
Shift-prefix handled incorrectly with Programmer Dvorak. The keymap move digits to upper layer — so, for example, you get <S-7> instead of <7> and neovim ignore mapping because of it.
Also, add appropriate 'k' suffix to keypad keys.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -79,6 +79,10 @@ QString InputConv::modPrefix(Qt::KeyboardModifiers mod)
 	   ) {
 		modprefix += "A-";
 	}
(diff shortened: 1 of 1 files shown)


major-kolz:
It also has updated bindings — sorry for that, I totally miss that it would be in comment

equalsraf:
@major-kolz I've merged your keypad patch, and the `<lt>` was fixed in another PR, can you rebase this?

major-kolz:
Was confused with whole rebase thing — so create fresh PR #261 with without keypad and `<lt>` patches.
Sorry

Issue #237, opened 2017-02-14
$NVIM_QT_RUNTIME_PATH not set in Windows


xelra:
On Windows `$NVIM_QT_RUNTIME_PATH` doesn't seem to be set, which results in the neovim-gui-shim plugin not being loaded. In my case it should point to `C:\Program Files (x86)\Neovim-Qt\share\nvim-qt\runtime`.

`$VIMRUNTIME` is set and `C:\Program Files\Neovim\share\nvim\runtime` is properly loaded to the runtimepath.

xelra:
I've experimented a bit and added `NVIM_QT_RUNTIME_PATH` to the user environment variables.

I can print from inside Neovim-Qt now:
```
:echo $NVIM_QT_RUNTIME_PATH
C:\Program Files (x86)\Neovim-Qt\share\nvim-qt\runtime
```
The path is however still not added to the runtimepath.

xelra:
Seems like Neovim-Qt does actually try to add `C:\Program Files (x86)\Neovim-Qt\share\nvim-qt\runtime` to the runtimepath, but does so unsuccessfully. In the runtimepath there's `C:\Program`.

So this seems to be a quoting or escaping issue.

justinmk:
This seems to be the problem: https://github.com/equalsraf/neovim-qt/blob/489416f130c8b7d9297890ddf6f0e2fff940643f/src/gui/app.cpp#L188

Another case: https://github.com/equalsraf/neovim-qt/blob/489416f130c8b7d9297890ddf6f0e2fff940643f/src/gui/app.cpp#L182

Easiest and most robust approach might be:

```c
neovimArgs.insert(1, QString("let &rtp.='%1'")
    .arg(NVIM_QT_RUNTIME_PATH));
```

equalsraf:
Started a PR #238, @xelra can you check if this works on Windows?

xelra:
This is fixed by #238.

PR #238, opened 2017-02-15
Fix issue #237

equalsraf:
Fix expression quoting when changing the &rtp using nvim CLI
arguments.

As suggested by @justinmk.

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -179,13 +179,13 @@ NeovimConnector* App::createConnector(const QCommandLineParser& parser)
 		auto path = qgetenv("NVIM_QT_RUNTIME_PATH");
 		if (QFileInfo(path).isDir()) {
 			neovimArgs.insert(0, "--cmd");
-			neovimArgs.insert(1, QString("set rtp+=%1")
+			neovimArgs.insert(1, QString("let &rtp.=',%1'")
 					.arg(QString::fromLocal8Bit(path)));
 		}
 #ifdef NVIM_QT_RUNTIME_PATH
 		else if (QFileInfo(NVIM_QT_RUNTIME_PATH).isDir()) {
 			neovimArgs.insert(0, "--cmd");
-			neovimArgs.insert(1, QString("set rtp+=%1")
+			neovimArgs.insert(1, QString("let &rtp.=',%1'")
 					.arg(NVIM_QT_RUNTIME_PATH));
 		} else
 #endif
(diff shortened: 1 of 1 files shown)


justinmk:
There's another case on line 208.

xelra:
I've built it already with the third line also quoted, like @justinmk suggested. Unfortunately it doesn't fix #237.

xelra:
I've played all day with the code and on my system the `neovimArgs` in the return statement look like this `--cmd let &rtp.=',C:\Program Files (x86)\Neovim-Qt\share\nvim-qt\runtime' --cmd set termguicolors` when joined and separated with whitespace. This looks right to me. Somewhere after that the path gets truncated to `C:\Program`.

xelra:
Actually I've manually changed the `neovimArgs` and it didn't have any effect on the runtimepath. So I think the `C:\Program` doesn't even come from Neovim-Qt, at least not from this function.

My current hypothesis is that the `C:\Program` comes from somewhere else and that the `C:\Program Files (x86)\Neovim-Qt\share\nvim-qt\runtime` from `NVIM_QT_RUNTIME_PATH` doesn't get added to the runtimepath at all.

equalsraf:
> My current hypothesis is that the C:\Program comes from somewhere else and that the C:\Program Files (x86)\Neovim-Qt\share\nvim-qt\runtime from NVIM_QT_RUNTIME_PATH doesn't get added to the runtimepath at all.

Well it should try three different things in this order:

1. If $NVIM_QT_RUNTIME_PATH is set use that
2. if NVIM_QT_RUNTIME_PATH is defined by the preprocessor then use that. This one comes from the CMake install path https://github.com/equalsraf/neovim-qt/blob/67f796dad544abf61cce5ae7e3ce5b8131a226dc/src/gui/CMakeLists.txt#L26
3. Relative to the nvim-qt binary as ../share/nvim-qt/runtime

> My current hypothesis is that the C:\Program comes from somewhere else 

Can you test with nvim by itself and see what happens. From a console:

    nvim --cmd "echo &rtp | q"

justinmk:
```
C:\Program Files\Neovim>nvim --cmd "echo &rtp | q"
C:\Users\jkeyes\AppData\Local\nvim,C:\Users\jkeyes\AppData\Local\nvim\site,C:/Program Files (x86)/nvim/share/nvim,C:\Users\jkeyes\AppData\Local\nvim\site\after,C:\Users\jkeyes\AppData\Local\nvim\after
```

xelra:
I just hard reseted to this PR again and now it works. I could have sworn I had tried `"let &rtp.=',%1'"` everywhere.

justinmk:
Let's close this then.

equalsraf:
Great. Thanks for taking the time @xelra.

Issue #53, opened 2015-10-08
Quitting requires extra input before closing window


expipiplus1:
To reproduce:

Open neovim-qt
type `:q<CR>`
observe that the window doesn't close
click the mouse, or press any key
observe that only now does the window close.

I'm using HEAD of both neovim and neovim-qt

equalsraf:
@expipiplus1 are you having this issue Windows?

@raghur reported the same issue, and I can replicate, but only on Windows. It seems Neovim blocks in the event loop (`loop_close` [comment](https://github.com/neovim/neovim/pull/810#issuecomment-143092166)) until another event arrives.

expipiplus1:
Whoops! yes I am on Windows.

RedX2501:
This also happens when double clicking the icon on the top left corner to close. Just downloaded latest versions from appveyor.

wsdjeg:
I got the same issue in win7, just marked.

xelra:
This issue also happens when simply closing the window with the X in the top right corner. One has to press the X button twice.

Herz3h:
This issue happens to me too have to click the X button twice to close it. I'm on Windows 8.1 compiled from latest src

xelra:
So with the runtimepath bug fixed, this is the only remaining big show stopper on Windows as far as I can tell right now.

I'm already building bad muscle memory by clicking enter twice. 

I've tested Nyaovim and Oni on Windows and they don't seem to have this problem. So it's likely not a nvim issue.

equalsraf:
So there are at least a couple different cases:

1. user calls something like `:q` that terminates nvim
2. user presses the window close button (or equivalent) in this case nvim-qt executes the command "qa" for you

In either case nvim-qt should be notified about the event here https://github.com/equalsraf/neovim-qt/blob/28e305fe54725481daafcda84dce0cd99083f522/src/gui/mainwindow.cpp#L61

Poking a bit at it in https://github.com/equalsraf/neovim-qt/pull/241

equalsraf:
Ok, it was late laste night, but I think there is a workaround in https://github.com/equalsraf/neovim-qt/pull/241. Can someone with windows give it a shot.

xelra:
I tested closing the window with `:q!`, `:qa` and by pressing the close button on the window. All of them work perfectly.

I tested with https://ci.appveyor.com/api/buildjobs/w8bn10spxcorpour/artifacts/neovim-qt.zip.

[...]

PR #241, opened 2017-02-16
Debug #53

equalsraf:
Lets have a look at debugging the close window issue, I added a debug statement to the process exit event.

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -41,6 +41,8 @@ void MainWindow::init(NeovimConnector *c)
 			this, &MainWindow::neovimMaximized);
 	connect(m_shell, &Shell::neovimFullScreen,
 			this, &MainWindow::neovimFullScreen);
+	connect(m_shell, &Shell::neovimGuiCloseRequest,
+			this, &MainWindow::neovimGuiCloseRequest);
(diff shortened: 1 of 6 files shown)


justinmk:
During a (unrelated) debugging session I noticed that `os_inchar()` might get stuck waiting for input if the UI is detached and then a "press enter" prompt occurs or some other input wait. 

Might try removing removing all calls to ui_detach. Let nvim close the connection, and interpret that as "quit". Could listen to VimLeave to verify that the quit was intentional.

xelra:
About verifying whether the quit was intentional, maybe that would solve https://github.com/equalsraf/neovim-qt/issues/247 in one fell swoop.

equalsraf:
> About verifying whether the quit was intentional, maybe that would solve #247 in one fell swoop.

I was thinking the same. Pushed a couple commits, they seem to work for #247, but have not tested the windows issue yet.

equalsraf:
This seems to fix it for windows too - or more accurately it works around it, there is probably some deeper issue on my side.

Anyway I've cleaned this PR, and will merge tomorrow. If it works well its probably worth a tag too.

justinmk:
Tried https://ci.appveyor.com/api/buildjobs/w8bn10spxcorpour/artifacts/neovim-qt.zip

It works :)

Issue #130, opened 2016-05-26
Only a blank window is shown


yhylord:
I'm using Arch Linux with Linux 4.5.0-1. I've installed it by the package in AUR. The compiler is GCC 6.1.1 and the Qt version is 5.6.0-1.
I tried running `nvim-qt` under command prompt, but it complains nothing.
![screenshot_20160526_220449](https://cloud.githubusercontent.com/assets/5259071/15577171/f4ffbdb8-2320-11e6-8251-ce3d5943d4ee.png)

equalsraf:
@yhylord maybe its stuck on neovim initialization, what happens if you run `nvim-qt -- -u NONE` ?

yhylord:
It's the same. No output from the command line or anything displayed.

equalsraf:
Maybe its the same as https://github.com/equalsraf/neovim-qt/issues/120#issuecomment-214545358

yhylord:
Exactly! After installing Inconsolata everything turns fine. Thanks!

equalsraf:
Leaving this open because I have no way to reproduce ATM, need a way to reliably load a fallback monospace font in Qt.

htfy96:
Confirmed on my Arch Linux. Since I am using my own font configuration, installing Inconsolata doesn't work to me.

```
fc-match monospace
wqy-zenhei.ttc: "文泉驿等宽正黑" "Regular" // this is a Chinese font
```

kiryl:
It seems the problem triggers when font has ligatures. I use PragmataPro as monospace alias and it leads to blank screen. Changing it to PragmataProMono (which is the same font, but without ligatures) it works fine.

It also worth noting that setting font in ginit.vim doesn't help as it loaded too late when we already stuck at blank screen.

lilydjwg:
I've added this to my fontconfig file to avoid the problem:

```xml
  <match target="pattern">
    <test name="prgname" compare="eq">
      <string>nvim-qt</string>
    </test>
    <edit name="family" mode="prepend" binding="strong">
      <string>DejaVu Sans Mono</string>
    </edit>
  </match>
```

neovim-qt doesn't like Emoji One Color.

sliceofcode:
Okay, so it looks like non-monospace fonts cause it to happen.

My fontconfig configuration forced `monospace` to resolve to Iosevka:
```
[ch:~]$ fc-match monospace
iosevka-regular.ttf: "Iosevka" "Regular"
```

[...]

PR #242, opened 2017-02-16
Force default fonts even for bad Monospace fonts

equalsraf:
See #130 in some systems we fail to load a default font, force the
loading of the font for the default startup font.


sliceofcode:
Fixes my blank window issue. :+1:

PR #249, opened 2017-02-28
RFC: Background images & X11 Psuedotransparency

lharding:
### DISCLAIMER: This is not meant to be ready to merge just yet

Instead I'm trying to generate comment. I've been implementing this for my own use and it's reached the stage where it's working for my use case but isn't really in a state to go upstream because of the following questions:

1. How to turn this feature on and off? It's not properly neovim-internal or even (IMHO) semi-universal like GuiFont. Should there be a command-line switch/environment variable/etc? Add something like GuiFont after all, that toggles it from nvim?

2. Does anyone care about using a background image from a file instead? Displaying the image from that file in a non-screen-relative way?

3. It's in need of disablement on non-X11 platforms and I haven't looked into that yet.

4. My highly-visible-cursor hack is in here and probably needs a separate toggle.

5. Refactoring? I'm on the fence as to whether the bg pixmap loader should be part of MainWindow, Shell, live by itself, etc.

Thoughts?

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -7,49 +7,12 @@
 #include <QApplication>
 #include <QKeyEvent>
 #include <QMimeData>
-#include <X11/Intrinsic.h>
 #include "msgpackrequest.h"
 #include "input.h"
 #include "konsole_wcwidth.h"
 #include "util.h"
 
 namespace NeovimQt {
-        QImage* bgInit() {
-            Display *dsp = XOpenDisplay(0);
-            Window root;
-            unsigned long bg = 0x1200001;
-            int xo, yo;
-            unsigned int width, height, border_width, depth;
-            unsigned char* bgData;
-
-            XGetGeometry(dsp, 0x1200001, &root, &xo, &yo, &width, &height, &border_width, &depth); 
-            fprintf(stderr, "%d,%d %dx%d %dbpp", xo, yo, width, height, depth);
-
-            int pixelSize = depth/8;
-            
-            XImage *img = XGetImage(dsp, 0x1200001, 0, 0, width, height, 0xFFFFFF, XYPixmap);
-
-            bgData = (unsigned char*)malloc(width*height*pixelSize);
-
-            for(int y=0; y<height; y++) {
-                for(int x=0; x<width; x++) {
-                    unsigned long px = XGetPixel(img, x, y);
-                    unsigned char* data = (unsigned char*)&px;
-                    unsigned char* base = &bgData[y*width*pixelSize+x*pixelSize];
-                    base[0] = data[2];
-                    base[1] = data[1];
-                    base[2] = data[0];
-                }
-            }
-
-            return new QImage(bgData, width, height, QImage::Format_RGB888);
-        }
-
-        //QRect rec = QApplication::desktop()->screenGeometry();
-        //bkgnd = new QPixmap(bkgnd->scaled(rec.width(), rec.height(), Qt::IgnoreAspectRatio));
-        //QPainter p(&bkgnd);
-        //p.fillRect(QRect(0,0,width,height), QColor(0, 0, 0, 128));
-
 Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 :ShellWidget(parent), m_attached(false), m_nvim(nvim),
 	m_font_bold(false), m_font_italic(false), m_font_underline(false), m_font_undercurl(false),
@@ -62,8 +25,6 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 {
 	setAttribute(Qt::WA_KeyCompression, false);
 
-        QPixmap* bkgnd = new QPixmap(QPixmap::fromImage(*bgInit()));
-        setBackgroundPixmap(bkgnd);
 	setAcceptDrops(true);
 	setMouseTracking(true);
 	m_mouseclick_timer.setInterval(QApplication::doubleClickInterval());
@@ -637,7 +598,10 @@ void Shell::paintEvent(QPaintEvent *ev)
 		} else if (wide) {
 			cursorRect.setWidth(cursorRect.width()*2);
 		}
+
 		QPainter painter(this);
+                // white outline to show us where the cursor is even if colors would make this ambiguous:
+                painter.drawRect(cursorRect.adjusted(0,0,-1,-1));
 		painter.setCompositionMode(QPainter::RasterOp_SourceXorDestination);
 		painter.fillRect(cursorRect, m_cursor_color);
 	}
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -22,7 +22,7 @@ namespace NeovimQt {
 	// Stolen from https://gist.github.com/ehamberg/767824/b4b4b0d1732d565d96b585723ebcc976cf6ca6a9
 	Pixmap GetRootPixmap(Display* dsp, Window *root)
 	{
-		Pixmap currentRootPixmap;
+		Pixmap currentRootPixmap = 0;
 		Atom act_type;
 		int act_format;
 		unsigned long nitems, bytes_after;
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -13,66 +13,66 @@
 #include "util.h"
 
 namespace NeovimQt {
-        QPixmap* bkgnd = NULL;
+	QPixmap* bkgnd = NULL;
 
 #ifdef WITH_X11
-        #include <X11/Intrinsic.h>
-        #include <X11/Xatom.h>
+	#include <X11/Intrinsic.h>
+	#include <X11/Xatom.h>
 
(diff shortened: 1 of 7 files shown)


justinmk:
What does this have to do with X11? Doesn't Qt support setting a background image on the editor widget in a cross-platform way?

lharding:
It might, I'm not much of QT dev usually. I did observe that the way ShellWidget is written this will almost certainly be invisible because it explicitly paints with neovim's bg color when erasing, rather than erasing. Arguably that might be a separate PR.

What this has to do with X11 is that this PR is about implementing proper root-window-pixmap-based pseudotransparency and so the above change is really a side effect. A screenshot might elucidate:
2017-02-28-162957_1160x993_scrot

This way one gets a nice background image without underlying text or UI elements interfering. I'd imagine there's a way of getting something similar on other OSes, but it's inevitably going to be windowing-system dependent.

equalsraf:
Neat, for some reason transparency bg/pixmap is usually a bit tricky in Qt (or at least it was in Qt4).

I'm not sure what is the "correct" to do it using the Qt API

- there is a X11 specific module in Qt, QX11Info http://doc.qt.io/qt-5/qx11info.html#appRootWindow but I dont see a way to get the pixmap using that
- no idea if the QDesktopWidget can be used to grab the pixmap using the QWidget API

> - How to turn this feature on and off? It's not properly neovim-internal or even (IMHO) semi-universal like GuiFont. Should there be a command-line switch/environment variable/etc?

No rush there, a command line switch is good enough to get it going. Adding something like GuiFont is simple enough. A function like 'GuiWindowMaximized' to toggle the pixmap is probably what you want, but if you wan to leave runtime toggle to a later PR that is fine be me.

>  Does anyone care about using a background image from a file instead? Displaying the image from that file in a non-screen-relative way?

> It's in need of disablement on non-X11 platforms and I haven't looked into that yet.

Since Qt loads its drawing backend at runtime (even in Linux you might be using wayland, etc) you need to check at runtime - http://doc.qt.io/qt-5/qx11info.html#isPlatformX11.

It think it is possible to do it all using the Qt API alone. e.g. http://stackoverflow.com/questions/25133517/how-to-get-pixel-under-mouse-cursor 

Tried a couple of options, but sadly its not working yet:

```c++

#include <QApplication>
#include <QDesktopWidget>
//#include <QWindow>
#include <QDebug>
#include <QPixmap>
#include <QScreen>
#include <QWindow>

int main(int argc, char **argv) {
	QApplication app(argc, argv);

	QDesktopWidget *desktop = app.desktop();
	// This gives you the default screen, but you can pick the one where your window
	// is placed with ::screenNumber(widget)
	QWidget *screenWidget = desktop->screen();

	// This seems ok to get the root window id
	qDebug() << "Root window Id" << screenWidget->winId();

	// I assumed this would work, but it gets me a blank pixmap
	screenWidget->grab().save("desktopwidget.png");

	// QScreen::grabWindow grabs pixels from the screen i.e. its a screenshot that
	// includes windows on top of the target window, so not what we want
	QScreen* screen = screenWidget->windowHandle()->screen();
	screen->grabWindow(screenWidget->winId()).save("screen.png");

	return 0;
}
```

> Refactoring? I'm on the fence as to whether the bg pixmap loader should be part of MainWindow, Shell, live by itself, etc.

I'm fine with it being part of the shell, since its grabbing the root pixmap. If it used compositor opacity to do it then we would have to move it someplace else, because the mainwindow would probably get in the way.

lharding:
Okay so I changed everything:

- Added a `Gui` event `BgImage`. Right now it takes one parameter, which if set to `desktop` will cause neovim-qt to grab the X11 root pixmap (if available) and set it as a background image with some (presently nonconfigurable) greyscale dimming for text readability. This was easier than a CLI switch so I just did it; I haven't added anything to the GUI shim plugin so for now one activates this with something like this in their `ginit.vim`: `call rpcnotify(0, "Gui", "BgImage", "desktop")`

- Moved the bg loading code into `shell.cpp`.

- By assuming that the pixmap fetched is in RGB32 format (should be almost always true nowadays) we save a lot of time at startup and one screen-sized framebuffer worth of RAM.

- Added error traps so that it should fail gracefully (or at least non-segfaultingly) if something fails, including not being able to get an X11 server connection in the first place. I'm just checking the result of XOpenDisplay for this so as not to haul in QtX11Extras.

- Tried to make the X11 dependency automatically optional at build time. Not sure if I succeeded due to lack of a non-Linux machine to test on, although it looks like the test runners think I did.

I think this addresses all comments and is *more or less* ready to merge (unless you want me to remove my visible-cursor hack).

src/gui/shell.cpp
```diff
@@ -500,6 +562,15 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			setLineSpace(val);
 			m_nvim->neovimObject()->vim_set_var("GuiLinespace", val);
 			resizeNeovim(size());
+		} else if (guiEvName == "BgImage" && args.size() == 2) {
+			QString imgDesc = m_nvim->decode(args.at(1).toByteArray());
+                        if(!imgDesc.compare("desktop")) {
+                                setBackgroundPixmap(bgInit());
+                        }
+                        else {
+                                setBackgroundPixmap(NULL);
```
equalsraf: nitpick join `} else {` into a single line

equalsraf:
> This was easier than a CLI switch so I just did it; I haven't added anything to the GUI shim plugin so for now one activates this with something like this in their ginit.vim: call rpcnotify(0, "Gui", "BgImage", "desktop")

:D I just at some point someone will want to use this to load arbitrary files as `file:///home/user/images/...`.

> Tried to make the X11 dependency automatically optional at build time. Not sure if I succeeded due to lack of a non-Linux machine to test on, although it looks like the test runners think I did.

Have not checked in Windows yet but appveyor succeeded, so it does build.

src/gui/shell.cpp
```diff
@@ -13,6 +13,68 @@
...
+                        if(!root) return NULL;
+
+                        Pixmap bg = GetRootPixmap(dsp, &root);
+                        if(!bg) return NULL;
+
+                        XGetWindowAttributes(dsp, root, &attrs);
+                        XImage* img = XGetImage(dsp, bg, 0, 0, attrs.width, attrs.height, ~0, ZPixmap);
+                        if(!img) return NULL;
+
+                        const int pixelSize = img->bits_per_pixel/8;
```
equalsraf: pixelSize is never used

src/gui/shell.cpp
```diff
@@ -500,6 +562,15 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			setLineSpace(val);
 			m_nvim->neovimObject()->vim_set_var("GuiLinespace", val);
 			resizeNeovim(size());
+		} else if (guiEvName == "BgImage" && args.size() == 2) {
+			QString imgDesc = m_nvim->decode(args.at(1).toByteArray());
+                        if(!imgDesc.compare("desktop")) {
```
equalsraf: The if statement can just use `==`.

src/gui/shell.cpp
```diff
@@ -500,6 +562,15 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
...
+                                setBackgroundPixmap(bgInit());
+                        }
```
equalsraf: Indentation in this block seems wrong, tabs instead of spaces.

src/gui/shell.cpp
```diff
@@ -13,6 +13,68 @@
...
+
+                return currentRootPixmap;
+        }
+#endif
+
+        QPixmap* bgInit() {
+                if(bkgnd == NULL) {
+#ifdef WITH_X11
+                        Display *dsp = XOpenDisplay(0);
+                        if(!dsp) return NULL;
```
equalsraf: Can we get some debug statements for the error cases?

equalsraf:
In my system it fails to get the ximage in `XImage* img = XGetImage(dsp, bg, 0, 0, attrs.width, attrs.height, ~0, ZPixmap);` maybe because I'm running i3.

src/gui/shell.cpp
```diff
@@ -500,6 +562,15 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			setLineSpace(val);
 			m_nvim->neovimObject()->vim_set_var("GuiLinespace", val);
 			resizeNeovim(size());
+		} else if (guiEvName == "BgImage" && args.size() == 2) {
+			QString imgDesc = m_nvim->decode(args.at(1).toByteArray());
+                        if(!imgDesc.compare("desktop")) {
+                                setBackgroundPixmap(bgInit());
```
equalsraf: If we fail here an error message would help. See setguifont for an example e.g.

```C++
m_nvim->neovimObject()->vim_report_error("Invalid font height");
```

lharding:
This should address all the comments. Still remaining seems to be building successfully on Mac, which seems to be failing because it disagrees with itself as to whether it has X11 or not. I'm not sure how to address this, since I'm not really set up to test on MacOS.

PR #254, opened 2017-03-17
Project tree

ferranpm:
I don't know if you're interested in having this but I think it's a nice feature to have.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -16,7 +16,7 @@ add_subdirectory(shellwidget)
 
 include(GNUInstallDirs)
 set(RUNTIME_PATH )
-add_library(neovim-qt-gui shell.cpp input.cpp errorwidget.cpp mainwindow.cpp app.cpp
+add_library(neovim-qt-gui shell.cpp input.cpp treeview.cpp errorwidget.cpp mainwindow.cpp app.cpp
 	${CMAKE_SOURCE_DIR}/third-party/konsole_wcwidth.cpp
 	${NEOVIM_RCC_SOURCES})
 target_link_libraries(neovim-qt-gui qshellwidget neovim-qt)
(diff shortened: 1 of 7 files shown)


justinmk:
What's the benefit vs netrw ?

suy:
On Friday 17 March 2017 14:24:08 Justin M. Keyes wrote:
> What's the benefit vs netrw ?

I think it should be pretty obvious. :-)

GUIs offer a superset of the features of CLIs if the implementation is right 
and complete. Yes, I would not merge this as-is, because it might lack a 
proper plan on how to change focus between the file manager and the editor 
(and other possible "features" or "panes", or whatever might be called), or at 
least that seemed so after a very quick review. But having a GUI with perfect 
keyboard-only usage is perfectly doable with Qt. My company implemented that 
for a Qt-based project, and I just published a blog post on the topic:

http://www.vikingsoft.eu/keyboard-navigation-and-the-event-system/

I imagine that was your concern.

ferranpm:
I don't have any preference, but I like it, it's simple and works. I find nerdtree to be a bit cumbersome.

I'll take a look at @suy 's post and try and improve it.

nbicalcarata:
This is awesome! Thank you so much for this!

captura de pantalla de 2018-02-07 19-22-04

kidproquo:
This is very nice! Any chance of an update?

tkdchen:
Is it possible to add multiple projects into this tree? It would be super nice to have it.

equalsraf:
@gagbo generally speaking you should wait for neovimconnector::ready before calling neovimObject i.e. you cannot expect the object to be available before a connection with neovim has been established.

gagbo:
I got it, I changed the code so that the constructor actually connects an "init callback" to the ready signal from NeovimConnector. I messed up the formatting too, so I'm currently using johnmcfarlane/unformat on the current masterto try and deduce a `.clang-format` to use locally, and I'll push again.

Do you prefer that I open a new PR, or maybe should I just send a PR to @ferranpm ?

Issue #230, opened 2017-02-07
mappings with < don't work


BjornMelgaard:
I have found, that mappings with < don't work only in neovim-qt, from terminal all work fine.
[Related issue](https://github.com/rhysd/clever-f.vim/issues/30#issuecomment-277901873)

What I do
```
nvim-qt -- --noplugin -u /dev/null
:vnoremap < <gv
:vnoremap > >gv
```
\> work fine, < work like ordinary <

Archlinux
```
NVIM 0.1.7
Build type: Release
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -Wconversion -O2 -DNDEBUG -DDISABLE_LOG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -D_GNU_SOURCE -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.1.7/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
Compiled by builduser

Optional features included (+) or not (-): +acl   +iconv    +jemalloc +tui      
For differences from Vim, see :help vim-differences

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

```

```
 pacman  -Qi neovim-qt-git 
Name            : neovim-qt-git
Version         : v0.2.4.r6.gfd85978-1
```

justinmk:
can be closed

BjornMelgaard:
:bowing_man:

Issue #204, opened 2016-10-30
Broken mapping in visual mode


ferranpm:
```
vnoremap > >gv
vnoremap < <gv
```

I have these mappings to indent/unindent multiple times without having to reselect the region. The first one works fine. The second one unindents the region but then the visual selections breaks. It works fine in the console.

This is in Archlinux with a US keyboard.
I've installed neovim-qt from the package in AUR
Qt version 5.7.0
Using this version: v0.2.2

amagee:
Same issue here:

Ubuntu 16.10
Qt 5.6.1
Nvim 0.2.0-dev
Neovim-qt from source: revision 4015d8cc5b75b56ffa758be115ea3309c58058b6

PR #260, opened 2017-03-30
Fix issue #230 and #204

ferranpm:
Don't know if I'm missing something here, but since the text is already `"<"` it should send the input `"<lt>"`


equalsraf:
The only hypothetical would be if if `<S-lt>` or `<C-lt>` were possible but I dont think they are.

Thanks @ferranpm

PR #261, opened 2017-03-31
Fix numbers in Programmer Dvorak keymap (#233, #222)

major-kolz:
Shift-prefix handled incorrectly with Programmer Dvorak. The keymap move digits to upper layer — so, for example, you get `<S-7>` instead of `<7>` and neovim ignore mapping because of it.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -1,4 +1,3 @@
-
 #include "input.h"
 #include <QDebug>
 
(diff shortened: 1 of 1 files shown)


major-kolz:
Solved via [this commit](https://github.com/equalsraf/neovim-qt/commit/3fd464624db1810ced546ac7e98d4c4483542167)

PR #273, opened 2017-05-05
Touchpad scroll fixes

koalefant:
I am running neovim-qt on a laptop with a touchpad that sends fine wheel events way more often than a discrete mouse wheel would do. This fixes super-fast scroll and flips horizontal scroll direction to match other applications.


equalsraf:
Thanks @koalefant

PR #280, opened 2017-05-10
[RFC] Only use mouse scroll pixel delta on OSX

equalsraf:
3a5a3d1 added support for high resolution mouse scrolling, but this
is not available in all platforms. In platforms that do not mouse scrolling is currently not working.

Need to figure out in which platforms we use pixelDelta, or fallback to angleDelta.

ping @koalefant @diefans

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -667,6 +667,7 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
 
 void Shell::wheelEvent(QWheelEvent *ev)
 {
+#ifdef Q_OS_MAC
 	// For some reason <ScrollWheel*> scrolls multiple lines at once
 	// we have to account for it, to make sure that pixelDelta() is used correctly.
 	const int scroll_columns = 6;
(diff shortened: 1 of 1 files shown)


diefans:
FYI: http://bazaar.launchpad.net/~mir-team/qtmir/trunk/revision/544

equalsraf:
Merging, but could use some feedback on Mac.

koalefant:
Sorry the for the breakage. Original change was tested on X11 / ArchLinux and this patch essentially disables it in that case.

koalefant:
(by original I am speaking about pixelDelta change here)

equalsraf:
@koalefant no worries, I just rushed the merge to get scrolling back. What we really need is a way to figure out when we can use pixelDelta. But the Qt docs were not very helpful in this.

koalefant:
Qt documentation suggests testing for pixelDelta().isNull() in order to determine if HiDPI scrolling is present.

koalefant:
Here is the source: http://doc.qt.io/qt-5/qwheelevent.html#pixelDelta

Issue #274, opened 2017-05-05
Arrow keys broken while in normal mode (OSX)


coddingtonbear:
A quick bisect shows that this problem was introduced in c208c1a1b5b4283dea130dbde589939f11245e3a, but glancing at the commit, I can't imagine how.

Because the summary description might not be very descriptive, here are a couple videos showing how the problem exhibits itself:

Not working: https://www.youtube.com/watch?v=EC6gcJ-7aa4&feature=youtu.be
Working: https://www.youtube.com/watch?v=yQNXCi_f868&feature=youtu.be

equalsraf:
Honestly I cant exactly figure out the exact problem from the video :D

Regardless lets have a look at it, with the following patch to track input. I don't have a Mac to test but hopefully someone can help out

```diff
diff --git a/src/gui/shell.cpp b/src/gui/shell.cpp
index f169276..a3a98cc 100644
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -587,6 +587,7 @@ void Shell::keyPressEvent(QKeyEvent *ev)
                return;
        }
 
+       qDebug() << "vim_input" << inp;
        m_nvim->neovimObject()->vim_input(m_nvim->encode(inp));
        // FIXME: bytes might not be written, and need to be buffered
 }
```

Pressing the down arrow here prints out `vim_input "<Down>"`.

The commit you mention adds the keypad modifier 'K', wild assumption, maybe on OSX its sending `<K-Down>` which would definitely be invalid.

coddingtonbear:
Ahh -- it looks like you're right -- when press the arrow keys while in normal mode, the logs print:

```
vim_input "<kUp>"
vim_input "<kUp>"
vim_input "<kUp>"
vim_input "<kUp>"
vim_input "<kUp>"
vim_input "<kUp>"
```

What I was trying to demonstrate in those videos, by the way, was that one can neither access previous commands from Normal mode by pressing the up arrow, nor could one scroll vertically/horizontally through options presented by tools like Ctrl+P using the up/down arrows.

equalsraf:
Quick search found this issue that seems very similar http://gnuplot.10905.n7.nabble.com/Arrow-keys-not-working-on-qt-terminal-for-Mac-OS-X-td19090.html

> The KeypadModifier value will also be set when an arrow key is pressed as the arrow keys are considered part of the keypad.

[...]

PR #282, opened 2017-05-16
Add explicit check for keypad keys

equalsraf:
Neovim expects keypad keys to be represented with the sequence `<k0-9>`
or `<kHome>` `<kEnd>`, etc.

In OSX the arrow keys are also considered as part of the key pad which
caused them to be encoded as `<kUp>` which is not a valid key for nvim.

More generally the previous solution allowed invalid keys to be marked as keypad keys.
Add an explicit check for the supported keys. I've allowed these keys
to have modifiers e.g. `<S-k8>` but in practice this seems to be system
dependent - for example in some systems Shift-keypad5 will generate no
key event, or Shift-keypad8 will be the Up arrow. Meaning that these
will never be accurately detected as keypad keys.

Ping https://github.com/equalsraf/neovim-qt/issues/274 - @coddingtonbear can you test with this?

With num lock on, in my system all the keypad keys seem to work without modifiers and with Control. Shift seems to cause some unexpected outcomes - it works for some keys but not others (but it makes sense based on the markings on my keyboard).

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -79,9 +79,6 @@ QString InputConv::modPrefix(Qt::KeyboardModifiers mod)
 	   ) {
 		modprefix += "A-";
 	}
-    if ( mod & Qt::KeypadModifier ) {
-        modprefix += "k";
-    } 
 
 	return modprefix;
 }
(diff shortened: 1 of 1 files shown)


coddingtonbear:
Thanks for fixing this the right way, @equalsraf!  You're a way, way better project maintainer than me :-).

PR #283, opened 2017-05-17
gui: Add correct check for high resolution trackpads

equalsraf:
ping @koalefant ref https://github.com/equalsraf/neovim-qt/pull/280


equalsraf:
Testing it on a linux system now. pixel delta is not null here, however small wheel scrolls dont  translate into a scroll. pixelDelta().y() reports `-15` but this translates into a vertical scroll of `0`.

PR #285, opened 2017-05-20
Fixing touchpad scroll - attempt #2

koalefant:
After a bit of digging into Qt source code, this is now similar to how QScrollBar implements handling of the mouse wheel event.  To fix speed it was crucial to divide angleDelta by 120 (QWheelEvent::DefaultDeltasPerStep).

I also tried to compare pixelDelta() to this approach and it does not feel to be smoother. nvim scrolls 3 lines at a time anyway.

This is tested on X11 with touchpad and mouse.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -21,7 +21,8 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 	m_hg_foreground(Qt::black), m_hg_background(Qt::white), m_hg_special(QColor()),
 	m_cursor_color(Qt::white), m_cursor_pos(0,0), m_insertMode(false),
 	m_resizing(false),
-	m_mouse_wheel_delta_fraction(0, 0),
+	m_scroll_remainder(0.0f, 0.0f),
+	m_scroll_last_direction(0, 0),
 	m_neovimBusy(false)
 {
 	setAttribute(Qt::WA_KeyCompression, false);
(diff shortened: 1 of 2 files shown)


viktorstrate:
I can confirm this works.

Neovim mouse scroll is way too fast current neovim-qt, which makes it nearly impossible to use the mouse to scroll. This PR makes scrolling the same speed as neovim in the terminal.

equalsraf:
CI is failling to build `error: ‘DefaultDeltasPerStep’ is not a member of ‘QWheelEvent’` I assume DefaultDeltasPerStep was introduced in Qt 5.5 (semaphore uses 5.4, appveyor 5.3) a version check should work there (I cant find a definitive version in the docs though)

```c
#if (QT_VERSION >= QT_VERSION_CHECK(5, 5, 0))
```

What happens if the feature is not avaialble? in https://github.com/equalsraf/neovim-qt/pull/283 there was

```c
if (ev->pixelDelta().isNull()) {
```

koalefant:
I believe it safe to use value of 120 instead of DefaultDeltasPerStep with older versions, it is (was) hardcoded in Qt sources all over the place.

koalefant:
DefaultDeltasPerStep was first introduced in Qt release 5.5.0 and pixelDelta() is not used anymore.

koalefant:
I suspect that remaining test was failing due to issue on master. Should I rebase?

equalsraf:
The failling test in unrelated, its been failling in other PRs too.

equalsraf:
But please rebase into a single commit just for sanity.

equalsraf:
https://github.com/equalsraf/neovim-qt/pull/630

PR #288, opened 2017-06-03
Use Q_ENUM for MsgpackError and NeovimError

jamessan:
Q_ENUM gives better feedback when checking the value of an enum in
tests.

```
FAIL!  : NeovimQt::Test::connectToNeovimTCP() Compared values are not the same
   Loc: [/«PKGBUILDDIR»/test/tst_neovimconnector.cpp(65)]
```

vs.

```
FAIL!  : NeovimQt::Test::connectToNeovimTCP() Compared values are not the same
   Actual   (c->errorCause())             : NoError
   Expected (NeovimConnector::SocketError): SocketError
   Loc: [/home/jamessan/neovim-qt-0.2.7/test/tst_neovimconnector.cpp(65)]
```

Also, Q_ENUMS is [deprecated] in favor of Q_ENUM, introduced in Qt 5.5.

[deprecated]: https://doc.qt.io/archives/qt-5.5/qobject-obsolete.html#Q_ENUMS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ include(MacOSXPaths)
 
 # Qt
 set(CMAKE_AUTOMOC ON)
+find_package(Qt5Core REQUIRED)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5Network REQUIRED)
 find_package(Qt5Test REQUIRED)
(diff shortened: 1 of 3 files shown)


equalsraf:
5.5 is required for this, which is a problem since we still try to build with 5.3 but it should be possible to do with an ifdef.

jamessan:
Updated to use ifdef.  I left the QtCore `find_package` call in CMakeLists for completeness.

PR #307, opened 2017-07-11
Draw a proper undercurl

jbradaric:
Draw a proper undercurl instead of a dotted underline.


equalsraf:
@jbradaric awesome, thanks.

PR #323, opened 2017-08-19
Create entry-point to interact with neovim-qt

nbetm:
This PR is based on #156 and #158. The only difference is that it only
introduces the autocmd. So we can leave the shim loading optimization
for later.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -139,6 +139,7 @@ void Shell::setAttached(bool attached)
 		}
 		m_nvim->neovimObject()->vim_command("runtime plugin/nvim_gui_shim.vim");
 		m_nvim->neovimObject()->vim_command("runtime! ginit.vim");
+		m_nvim->neovimObject()->vim_command(QByteArrayLiteral("doautocmd User NeovimGuiAttached"));
 
 		// Noevim was not able to open urls till now. Check if we have any to open.
 		if(!m_deferredOpen.isEmpty()){


equalsraf:
Actually I was intending to do this upstream https://github.com/neovim/neovim/pull/6917 but i never got to finish it.

jeaye:
Any update on this?

equalsraf:
errr, nope, rebased but I dont have tests. If anyone wants to spin off the upstream PR

jeaye:
It sounds like you're not open to merging this PR and would instead do this upstream. If that's so, can we close this? If not, can we merge so people can take advantage of the hook.

equalsraf:
Correct, I would prefer to see this upstream. There is little benefit to rushing it here, since you can also do it as a hack inside ginit.vim. Also one of the original points of #156 can only be fixed upstream.

suy:
To me the changes in #158 did fix all what I expected, so it was good enough to have it merged here, and wait for an upstream solution with some easy workaround. Your pull request in neovim looked fine enough for me. Without being familiar with their internals, though. And there were tests, isn't it? I don't see what's missing there.

equalsraf:
@suy still missing that last comment https://github.com/neovim/neovim/pull/6917#discussion_r135390411 and the asserts are wrong and break the build.

jgehrig:
Closing this PR, since the behavior was implemented upstream with `UIEnter` and `UILeave`.

Issue #334, opened 2017-09-22
Document the need to run `make install` for Gui commands to work in the Readme 


e00E:
I built `neovim-qt` from source on Linux like the Readme explains. However when running it from the built directory the `:GuiFont` and `:help nvim-gui-shim` commands would not work.

I ran `make install` and then ran the installed version of `neovim-qt` and the commands worked.

This should be documented in the Readme so other people can spend less time pondering this issue.

equalsraf:
Closing after #337

PR #337, opened 2017-09-25
Update README with how to run GUI from source

equalsraf:
To run nvim-qt from source one needs to add the GUI shim to the vim
runtime manually, this can be done with the NVIM_QT_RUNTIME variable.

ref #334

@e00E does this work for you?


e00E:
Yes, thanks for fixing!

PR #346, opened 2017-10-30
GUI: tabline support

equalsraf:
Just fooling around.

- [x] do a correct version check, right now it just calls api1 with no checks
- [x] someone forgot to fix extension types https://github.com/equalsraf/neovim-qt/blob/master/bindings/generate_bindings.py#L81 🤓


equalsraf:
This is starting to look usable. Mouse clicking and scrolling works too.

Interestingly I have no idea how to remove a tabpage.

khalidchawtany:
Can we inherit the original **tabline** colors or customize the gui ones from `ginit.vim`?

I have a made the neovim-qt window frameless and this is how it appears with the gui_tabline:
screen shot 2017-11-01 at 12 43 09 pm

Without the gui_tabline
screen shot 2017-11-01 at 12 53 03 pm

equalsraf:
> Can we inherit the original tabline colors or customize the gui ones from ginit.vim?

I have nothing right now, and I'm not sure how to handle that.

Worst case scenario you can apply arbitrary Qt stylesheets. e.g with a file style.qss.

```css
QTabBar { background-color: red; border: none; }
```

I've pushed a patch to allow the use of `--stylesheet` from the CLI ([Qt docs](http://doc.qt.io/qt-5/stylesheet-reference.html)). So calling `nvim-qt --stylesheet style.qss` gives me

style

I can add some rpc to call from ginit.vim, but i'm not sure how that would work out.

khalidchawtany:
This works. However, On Mac OS we have application bundles. I cannot pass args to an app unless launched from a terminal, which defeats the purpose of the app bundle :(.  Some RPC calls from ginit.vim would be a great solution for both Macs and PCs.

equalsraf:
I think I can get an env var to have the same effect. I cant seem to find one in the Qt docs that does this.

justinmk:
Note that full 'tabline' customization will be handled in https://github.com/neovim/neovim/pull/6607

Would be better not to have GUI-specific options duplicating what's already in core, but I understand if it takes too long.

khalidchawtany:
It seems `qgetenv()` gets only those env. variables that are set from the terminal. And NOT those that are set using ```launchctl setenv``` which is available to GUI apps.

```
launchctl setenv NVIM_QT_STYLESHEET  SOMEVALUE
launchctl getenv NVIM_QT_STYLESHEET  //this should return SOMEVALUE
```

equalsraf:
I'm not sure about the launchctl specifics on mac. But it probably launches a different process everytime, so the variable does not persist. We do load environment variables from the shell on mac OS, so if you place the var .profile or similar it should work.

khalidchawtany:
Thanks you were right. I don't know why I didn't try that in the first place :)

equalsraf:
Ok I think feature wise this is complete then. Some more testing and this could be merged.

I'm sure someone will popup eventually and ask for a close button, or moveable tabs, but I have not idea how to put those together with the nvim api yet.

PR #150, opened 2016-07-01
Support for GUI popup menu

equalsraf:
Check https://github.com/neovim/neovim/pull/4432 for upstream WIP on allowing the GUI to handle the popup menu.

Indirectly this PR ended up holding other fixes too. Some of these are not backward compatible i.e. this only works with  neovim >= 0.1.5

PR #349, opened 2017-11-01
[RFC] GUI: Implemente popup menu

equalsraf:
Respining https://github.com/equalsraf/neovim-qt/pull/150. The first commits come from https://github.com/equalsraf/neovim-qt/pull/346 (fixes for the api version, attach, etc).


equalsraf:
This is not something I use a lot, so if anyone has suggestions or is aware of corner cases this is the time.

justinmk:
I'll try it out, looking forward to this!

justinmk:
Functionally I don't see any problems at all, overflow works correctly, CTRL-E, CTRL-N, CTRL-P.

On Windows 7 the selection could use a bit more contrast (maybe this is an OS default, but on this machine most context-menus highlight items as blue, not light-grey).

screen shot 2017-11-05 at 19 15 47

(Of course, eventually Nvim core should be enhanced to send the user's colorscheme preferences for popupmenu.)

prabirshrestha:
vim/neovim in termainl has flickering issue. Would be good to test if this has the same issue too. https://github.com/vim/vim/issues/1713

equalsraf:
>    Not sure what Nvim-Qt should do here. Don't know of a VimL function nor API function that can choose an item from a popupmenu. (@Shougo do you know of a way to do this?)

I think there is no way right now. 

Concerning the mouse I still have to figure out what to do there

1. I can allow mouse events to propagate to the parent - but this might feel awkward, clicking the menu will dismiss it as if the user clicks some other position in the editor
2. Ignore them altogether. (should be possible, a bit of work though)

BoltsJ:
I was toying around with using QPalette to style the popupmenu. It works on KDE, but the documentation for QPalette warns:
>Warning: Some styles do not use the palette for all drawing, for instance, if they make use of native theme engines. This is the case for both the Windows Vista and the macOS styles.

It doesn't say which color groups are overridden though, so it may still allow theming. Unfortunately, I have no way to test on windows or mac, and those particular style plugins only exist on their respective platforms. This is the patch to change the colors if anyone wants to test on Windows and on OSX to see if this approach to styling will work.

```patch
diff --git a/src/gui/popupmenu.cpp b/src/gui/popupmenu.cpp
index 18e692e..d99a936 100644
--- a/src/gui/popupmenu.cpp
+++ b/src/gui/popupmenu.cpp
@@ -1,6 +1,7 @@
 #include "popupmenu.h"
 #include "popupwidgetitem.h"
 
+#include <QPalette>
 #include <QScrollBar>
 
 namespace NeovimQt {
@@ -11,6 +12,13 @@ PopupMenu::PopupMenu(QWidget *parent)
 	setFocusPolicy(Qt::NoFocus);
 	setAttribute(Qt::WA_TransparentForMouseEvents);
 	setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+	QPalette a;
+	a.setColor(QPalette::Window, Qt::blue); // Effectively: Border
+	a.setColor(QPalette::Base, Qt::red); // Pmenu bg=
+	a.setColor(QPalette::Text, Qt::white); // Pmenu fg=
+	a.setColor(QPalette::Highlight, Qt::magenta); // PmenuSel bg=
+	a.setColor(QPalette::HighlightedText, Qt::black); // PmenuSel bg=
+	setPalette(a);
 }
 
 QSize PopupMenu::sizeHint() const {
```
image

equalsraf:
@BoltsJ you can also load a custom [qt stylesheet](https://doc.qt.io/qt-5/stylesheet-reference.html#selection-color-prop) with `--stylesheet`. An example similar to the one you use above would be

```css
QListWidget{
	border-color: blue;
	background: red;
	color: white;
	selection-background-color: magenta;
	selection-color: black;
}
```

But we should also be able to set the correct colors from neovim.

justinmk:
Is there a way to remove the icons? IMO that should be done by default.

equalsraf:
> Is there a way to remove the icons? IMO that should be done by default.

Yeah, i will likely remove those before merging. I wanted to somehow pass on the extra fields in the menu as icons, but could never find a solution I liked.

BoltsJ:
As far as clickability, I played around with that a bit as well and it's actually pretty simple. Creating these two slots allows the user to select the desired completion item by single clicking and insert it with double click.
```c++
void Shell::handlePopupMenuClicked(const QModelIndex &index)
{
	m_nvim->api2()->nvim_get_color_by_name("Pmenu");
	QByteArray inp, key;
	int delta = index.row() - m_pum.oldIndex();
	if (delta > 0) key += "<C-N>";
	else if (delta < 0) key += "<C-P>";
	for (delta = abs(delta); delta > 0; --delta) inp += key;
	m_nvim->api0()->vim_input(inp);
}

void Shell::handlePopupMenuDoubleClicked(const QModelIndex &index)
{
	m_nvim->api0()->vim_input("<C-Y>");
}
```
and connecting them like so in Shell
```c++
connect(&m_pum, &PopupMenu::clicked,
		this, &Shell::handlePopupMenuClicked);
connect(&m_pum, &PopupMenu::doubleClicked,
		this, &Shell::handlePopupMenuDoubleClicked);
```
As for grabbing the highlights from the colorscheme, API level 3 (I generated the bindings with the 0.2.2 appimage) provides `nvim_get_hl_by_name("HLGroup")` which reports the requested colors as a dictionary in vimL and a QVariantMap in the autogenned bindings. 

For the icons, would it be possible to have them blank, but load found icons from a location such as `~/.config/nvim/icons/` if they exist? It looks like the `kind`, `extra` and `info` fields are empty using filename completion without any plugins, but could those fields be used for determining which icon to use? I think that having configurable icons on the pum would be a feature a lot of people would want.

justinmk:
> configurable icons on the pum would be a feature a lot of people would want

Maybe, but it's out of scope for this first step, isn't it?

equalsraf:
> As far as clickability,

The single click behaviour works for selecting the menu entry, followed by a second click to finish completion- but should this be the expected behaviour?

> It looks like the kind, extra and info

These are not always empty. Usually extra holds the name of the file where completion comes from. The kind can also hold values, I believe oni uses the later to define the icon of the entry which is also what I was doing in this PR.

> As for grabbing the highlights from the colorscheme, API level 3

Ok including the bindings in this PR.

> Maybe, but it's out of scope for this first step, isn't it?

Either way I'm running out of time, so I'll probably merge as is.

BoltsJ:
>Maybe, but it's out of scope for this first step, isn't it?

Yeah, it's probably better to stub icon stuff out until it's completely ready.

>should this be the expected behaviour?

I checked the behavior for other guis. Nvim-gtk and eovim insert the item that is clicked. Matching those is probably better.

clason:
Would it be possible to show *all* non-empty fields (e.g., `word`, `abbr`, `menu`, `kind` for omnicomplete)? For example, vimtex makes use of them for additional completion information, so this is a significant regression with the popupmenu (which is not opt-in).

I can open a new issue if you prefer.

equalsraf:
> Would it be possible to show all non-empty fields (e.g., word, abbr, menu, kind for omnicomplete)?

Definitely. But I'm not sure when and how to display the additional fields.

> I can open a new issue if you prefer.

Done, https://github.com/equalsraf/neovim-qt/issues/441. We can continue this there. If you need to disable the popup you can use `GuiPopupmenu 0`.

clason:
> Definitely. But I'm not sure when and how to display the additional fields.

I would assume the same way the normal pum displays them?

clason:
And may I make the suggestion that as long as the new pum does not have feature parity with the old, it should be disabled by default (i.e., need to explicitly use `GuiPopupmenu 1`)?

Issue #347, opened 2017-10-30
GUI: externalize cmdline?


equalsraf:
Opening this because @khalidchawtany pinged me earlier but I really have no idea what to do with the externalized cmdline yet :)

teto:
having a cmdline popup in the middle of the window  would be cool. I dsilike the CLI cmdline at the bottom because I have to look down on it instead of looking at the text. I would vote for something like sublimetext palette.

jgehrig:
Closing this issue. The work can be tracked via Pull #586.

PR #351, opened 2017-11-04
WIP - GUI: Implement external cmdline

equalsraf:
Nothing to see here yet, just working out the api.

- [ ] I think this is api lvl 3, but I have not checked. the attachment code needs to be updated and we need to generate bindings.

ref https://github.com/equalsraf/neovim-qt/issues/347

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -205,6 +205,9 @@ void Shell::init()
 	if (m_options.enable_ext_popupmenu) {
 		options.insert("ext_popupmenu", true);
 	}
+	if (m_options.enable_ext_cmdline) {
+		options.insert("ext_cmdline", true);
+	}
 	options.insert("rgb", true);
 
 	MsgpackRequest *req;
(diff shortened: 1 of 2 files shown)


gagbo:
tl;dr : should we actually implement the cmdline as a sub-`Shell` widget ?

At first I thought that using a [QTextEdit](http://doc.qt.io/qt-5/qtextedit.html#details) widget would be the best solution for the issue, because QTextEdit can show/edit Rich Text. Rich text support is important to do something with the formatting information, contrary to [QLineEdit](http://doc.qt.io/qt-5/qlineedit.html#details).

Another advantage of QTextEdit over QLineEdit is the ability to show multiple lines/paragraphs. QLineEdit is not able to properly answer `cmdline_block_show` and `cmdline_block_append`

But then I thought that if we're editing text in this window we probably want a sub `Shell` there don't we ? And the `wildmenu` should then be implemented as a special instance of `PopupMenu` ?

jgehrig:
Superseded by Pull #492 and Pull #586.

Marking as closed.

Issue #352, opened 2017-11-05
Tabline and viewport size with maximized window


WolfgangMehner:
I noticed a problem with the new tabline. When I maximize the window, approx. two lines on the bottom are not visible anymore. (Meaning the Vim command-line is not visible.) If I do not maximize the window, even resizing it to take up the entire screen, everything looks OK.
I would assume that when the window is maximized, the size of the viewport is by mistake that of the whole window, while in the other case the size of the viewport is correctly calculated from the size of the widget which contains the viewport.

I have Ubuntu 16.04, with MATE desktop, and with
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.5)
and Qt5 (5.5.1+dfsg-16ubuntu7.5) from Ubuntu packages

Neovim is build from source https://github.com/neovim/neovim/commit/54b79f19d7c63e5a7673264665d95dac335ede2c
Neovim-Qt is build from source https://github.com/equalsraf/neovim-qt/commit/0551ef29108ebcfad2f9bcaa8658317bf8ca886c

Maximized:
![nvim-qt-fullscreen](https://user-images.githubusercontent.com/2720024/32418891-cbf3af46-c271-11e7-8c2e-74cc14662bdf.png)
Resized to fill up the screen:
![nvim-qt-non-fullscreen](https://user-images.githubusercontent.com/2720024/32418892-cc0bf434-c271-11e7-9445-d11d373787ad.png)

WolfgangMehner:
Same phenomenon with the standard Ubuntu Desktop

equalsraf:
Does it help if we disable the auto hide of the tabline? patch in https://github.com/equalsraf/neovim-qt/pull/353

equalsraf:
I cant reproduce in my window manager with a maximized window. But using fullscreen also works e.g. open with `nvim-qt --fullscreen` and then open a new tab.

WolfgangMehner:
The patch does not help, sorry. Running with `--fullscreen` shows the same undesired behavior on my system.

equalsraf:
Updated #352. Does this work for you now?

WolfgangMehner:
This seems to fix it for me, I will also try your solution later:
https://gist.github.com/WolfgangMehner/ada83e664af5dc5c431dd933fe571a8e

equalsraf:
[...]

PR #353, opened 2017-11-05
Fix layout issue with new tabline

equalsraf:
For #352. I assume the resize reports the wrong size to neovim. But I cant really reproduce myself


equalsraf:
Merged minimal fix for now. Keep the ball rolling in #352

Issue #376, opened 2018-01-19
FR: Please support qt5-static link.


zhaozg:
If yes, I'll make a PR.

equalsraf:
I dont mind supporting static linking in the build scripts but this is usually hard to do in windows without a custom Qt build and some Qt plugins are always dynamic libraries.

@zhaozg what did you mean with `cut down depends`?

zhaozg:
@equalsraf when static link QT, it has no depends, this is my mean.

zhaozg:
The support for qt5-static not finished, but you can look a head from https://ci.appveyor.com/project/zhaozg/neovim-qt/build/0.2.8.11.

PR #378, opened 2018-01-19
[RDY] support qt5-static

zhaozg:
Try to fix https://github.com/equalsraf/neovim-qt/issues/376


src/gui/CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ set(QTLIBS Qt5::Network Qt5::Widgets)
 include_directories(..)
 qt5_add_resources(NEOVIM_RCC_SOURCES data.qrc)
 if(WIN32)
+	if ($ENV{QT_STATIC})
```
equalsraf: Why are you using an environment variable here? Instead can you set it as a cmake option and use something like `-DUSE_STATIC_QT=1` instead?

Otherwise you might end up mixing up build option if you forget to set the variable.

zhaozg: ok, learning.

src/gui/main.cpp
```diff
@@ -6,6 +6,9 @@
 #include "neovimconnector.h"
 #include "app.h"
 
+#include <QtPlugin>
+Q_IMPORT_PLUGIN (QWindowsIntegrationPlugin);
+
```
equalsraf: Wrap the include and Q_IMPORT in a guard for example

```C++
#if defined(Q_OS_WINDOWS) && defined(USE_STATIC_QT)
```

This would also require adding a definition in CMakeLists.txt

zhaozg: Done

src/gui/CMakeLists.txt
zhaozg: Done

zhaozg:
hi @equalsraf `contrib/appveyor-msys2-qt5-static.bat 64` build with msys2/mingw64-x86-64 and `contrib/appveyor-msys2-qt5-static.bat 32` build with msys2/mingw64-i686,
default `contrib/appveyor-msys2-qt5-static.bat` will build 64 bits.
and you can give set env MBITS 32 or 64 to active matrix ci.

contrib/appveyor-msys2-qt5-static.bat
```diff
@@ -0,0 +1,52 @@
+set BITS=""
+set ARCH=""
+if "%1%"=="" (
+  if "%MBITS%" == "" (set BITS=64) else (set BITS=%MBITS%)
```
equalsraf: Can you keep the default to 32 bits? since all the other scripts do it too.

equalsraf:
Looking good. Can you add a line to contrib/appveyor.yaml to enable the build? This should work

```yaml
 - SCRIPT: contrib/appveyor-msys2-qt5-static.bat
```

src/gui/CMakeLists.txt
```diff
@@ -3,6 +3,11 @@ set(QTLIBS Qt5::Network Qt5::Widgets)
 include_directories(..)
 qt5_add_resources(NEOVIM_RCC_SOURCES data.qrc)
 if(WIN32)
+	if (${USE_STATIC_QT})
+		add_definitions(-DUSE_STATIC_QT)
+		link_directories($ENV{QTDIR}/share/qt5/plugins/platforms)
```
equalsraf: Is this link_directories necessary?

If it is, I think you can use ${CMAKE_PREFIX_PATH} instead of $ENV{QTDIR}

contrib/appveyor-msys2-qt5-static.bat
zhaozg: Done

src/gui/CMakeLists.txt
zhaozg: Sure need, libqwindows.a store in that path.
Done

equalsraf:
if you are bored of waiting for appveyor,  you can always move the `- SCRIPT` line to the top, that would make your version build first 😄

zhaozg:
32 bits, can't pass, because
```
error: target not found: mingw-w64-i668-gcc
error: target not found: mingw-w64-i668-cmake
error: target not found: mingw-w64-i668-jasper
error: target not found: mingw-w64-i668-python2
error: target not found: mingw-w64-i668-qt5-static
```

Msys2 pacman to sync download some package fail. that's strange.
Night 1 clock here, I have to go to bed, and let's fix it in the tomorrow.

equalsraf:
Ok, thanks.

zhaozg:
low level spell, sign

equalsraf:
Merged thanks @zhaozg.

justinmk:
Would it be worth adding this as an artifact? Or at least exercising it in travis.yml/appveyor.yml somehow.

zhaozg:
@justinmk If need,I will make another RP for nvim linked almost static libs.

justinmk:
@zhaozg If there's an advantage to that I'd be interested in using it for the Nvim windows package. One advantage I can think of: it makes it clear which of the DLLs are actually needed. Currently we ship a lot of Qt DLLs but some of them may not be needed.

zhaozg:
I have done with iconv and intl,now working on wintty,maybe in a fews day.

justinmk:
@zhaozg  wintty requires a separate `.exe` agent (`winpty-agent.exe`) anyways, so IMO it's not important to build it statically.

Issue #369, opened 2018-01-12
GuiLinespace does not correctly accept negative numbers


Leandros:
- `nvim --version`: v0.2.2
- Operating system/version: Windows 10 Version 1703 (OS Build 15063.540)
- Terminal name/version: nvim-qt.exe
- `$TERM`: xterm-256color

### Steps to reproduce

- Install the gui shims (github.com/equalsraf/neovim-gui-shim)
- Execute the following in nvim: `GuiLinespace -2`

### Actual behaviour

Linespace is changed to -2.

### Expected behaviour

Linespace stays at 0.

### Notes

Calling `call rpcnotify(0, 'Gui', 'Linespace', -2)` has the correct behavior, though.

equalsraf:
Looks like some type of strange cohercion is going on

With `call rpcnotify(0, 'Gui', 'Linespace', -2)`

```
raw val QVariant(qlonglong, -2) coherced 4294967294
```

Obviously wrong!

But If you use a string (which I assume is what GuiLinespace does) `
call rpcnotify(0, 'Gui', 'Linespace', '-2')` then

```
raw val QVariant(QByteArray, "-2") coherced 0
```

This can also be verified by checking g:GuiLinespacing getting large values.

Looking at the Qt docs its seems it fails to detect the overflow in the conversion http://doc.qt.io/qt-5/qvariant.html#canConvert

> Warning: If the value is convertible to a QMetaType::ULongLong but is too large to be represented in an unsigned int, the resulting arithmetic overflow will not be reflected in ok. A simple workaround is to use QString::toUInt().

equalsraf:
@Leandros #375 probably does the opposite of what you want but at least now it is consistent. I don't oppose negative linespace, I just wanted to fix the overflow first.

Leandros:
Alright, no problem. :+1:

PR #380, opened 2018-01-20
Accept negative GuiLinespace values

equalsraf:
Finishing #369 this should allow the use of negative linespace with the GuiLinespace command.

ping @Leandros


Leandros:
:+1:

PR #381, opened 2018-01-21
Support forwarded UI options

equalsraf:
nvim 0.2.3 will start forwarding UI configuration options as redraw
events. This adds support for

- guifont
- linespace

and ignores the following

- guifontset
- guifontwide

a quick note about 'guifont' is that the options requires escaping
spaces e.g.

    :set guifont=DejaVu\ Sans\ Mono:h24

where the shim command did not.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -459,6 +459,10 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
 		}
 
 		emit neovimTablineUpdate(curtab, tabs);
+	} else if (name == "option_set") {
+		if (2 <= opargs.size()) {
+			handleSetOption(opargs.at(0).toString(), opargs.at(1));
+		}
 	} else {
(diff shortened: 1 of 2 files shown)


justinmk:
:set guifont=DejaVu\ Sans\ Mono:h24

:let &guifont='DejaVu Sans Mono:h24' should work also.

src/gui/shell.cpp
```diff
@@ -459,6 +459,10 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
 		}
 
 		emit neovimTablineUpdate(curtab, tabs);
+	} else if (name == "option_set") {
+		if (opargs.size() == 2) {
```
justinmk: For forward compatibility it's better to check `opargs.size() >= 2`. Additional items will always be "optional" semantically.

equalsraf:
Thanks @justinmk, I was looking for that :)

PR #383, opened 2018-01-26
fix minor bug about font process

zhaozg:


equalsraf:
I assume this is an issue when the default font is already an italic/bold variant, correct?

zhaozg:
Sorry, I'm wrong

equalsraf:
@zhaozg no problem

Issue #384, opened 2018-01-26
Issuer about show half character on status bar.


zhaozg:
I install spacevim for nvim, and I got
![image 2](https://user-images.githubusercontent.com/542599/35446213-f63c93c2-02ee-11e8-870a-9b96f2da1f39.png)

And half character on status, please reference to https://spacevim.org/

equalsraf:
This sounds like a duplicate of https://github.com/equalsraf/neovim-qt/issues/110 or https://github.com/equalsraf/neovim-qt/issues/140. Check the discussion in the comments of https://github.com/equalsraf/neovim-qt/pull/141 we dont have a working solution.

zhaozg:
I merged trunk `tb-unicode-chars`, but show same with not patch.

equalsraf:
Ok this one seems to be different.

I also tested this in

- gvim cuts the char in half
- libvte cuts the char in half
- konsole renders it past its cell
- both vim and neovim treat this symbol as a single width char

I dont think this char is meant to be a double width char, at least both vim and nvim are consistent on this and so is libvte/konsole. I would suspect the symbols are missing from the fonts.

test snippet

```
//345
//⓿45
//❶45
//龍5
```

zhaozg:
@equalsraf I checked some font related exmaple int https://github.com/equalsraf/neovim-qt/tree/master/src/gui/shellwidget

```
> fontinfo.exe "DejaVu Sans Mono for Powerline"                     
Font family: DejaVu Sans Mono for Powerline                         
Exact match: 1                                                      
Real family: DejaVu Sans Mono for Powerline                         
Font size: 10                                                       
!  (Regular) Average char width: 8 Max char width: 17 Width(MM): 16 
!   (Italic) Average char width: 8 Max char width: 17 Width(MM): 16 
!     (Bold) Average char width: 9 Max char width: 18 Width(MM): 18 
!  (Bold+It) Average char width: 9 Max char width: 18 Width(MM): 18                                              
```
It's strange that font width is not fixed.  Maybe this is root reason.

zhaozg:
This https://github.com/equalsraf/neovim-qt/issues/266 same.

zhaozg:
@equalsraf Maybe help to you .

when utf-8 encode chinese character
```
> fontinfo.exe "DejaVu Sans Mono for Powerline"                                    
Font family: DejaVu Sans Mono for Powerline                                        
Exact match: 1                                                                     
[...]

Issue #110, opened 2016-03-13
Emoji support on OSX


dvcrn:
Not a big issue but definitely nice to have. The current version of neovim-qt does not seem to render emojis:

<img width="732" alt="screen shot 2016-03-13 at 1 31 22 pm" src="https://cloud.githubusercontent.com/assets/688326/13726929/20018b8a-e920-11e5-8652-45c8a5c22e20.png">

equalsraf:
How does this work? AFAIK this uses unicode chars, and should work provided that fonts support them. and as far as Neovim is concerned if the font is monospaced and these chars are considered to be wide (not sure if they are, surprisingly it seems [are not](https://news.ycombinator.com/item?id=10206380)).

A strange issue with your screenshot is that even the substitution char seems to rendered incorrectly, maybe a bug on our part.

Also tried to use these in my terminal (konsole) and got mixed results, sometimes they worked, but not when the cursor was over them, which suggests it does not consider them to be wide chars.

Sidenote: We use konsole's implementation of [wcwidth](https://github.com/equalsraf/neovim-qt/blob/master/third-party/konsole_wcwidth.cpp#L98) to detect the width of a character. Not sure what other terminals do, [iterm2 seems to do some funky magic](https://wiki.dequis.org/notes/emoji/)

dvcrn:
I am not sure to be honest. I think the emojis on mac are provided by the OS and not by the font. Both screenshots, terminal (iterm2) and qt, are using the same font (Hack). 

Besides in ui elements, the same happens when opening a document that has emojis or trying to paste a emoji into a new document. 

Also not sure if this helps but Neovim.app renders emojis correctly

equalsraf:
Can any one check if https://github.com/equalsraf/neovim-qt/pull/141 fixes this issue? You might need to use the neovim from master instead of 0.1.4.

dvcrn:
@equalsraf looks like it's fixing it! Emojis are now rendered correctly

dvcrn:
There are occasionally visual glitches with the width of emojis though. 
<img width="179" alt="2016-07-16 9 44 37" src="https://cloud.githubusercontent.com/assets/688326/16891747/f79f66ae-4b39-11e6-9031-e2b74514bc89.png">

equalsraf:
@dvcrn thanks for checking, the visual glitches are not completely unexpected. Neovim and Neovim-qt can have disagreements about character width, I need to figure out how to deal with those cases.

[...]

PR #386, opened 2018-01-27
Fix Qchar width

zhaozg:
We follow a basic assumption: the Latin character is one character width and the non-Latin character is two characters width.

 I am in the Far East, this approach is usually used to deal with.
Now https://github.com/equalsraf/neovim-qt/issues/384, https://github.com/equalsraf/neovim-qt/issues/110, https://github.com/equalsraf/neovim-qt/issues/140 shoud be fixed.

### DIFF
--- a/src/gui/shellwidget/CMakeLists.txt
+++ b/src/gui/shellwidget/CMakeLists.txt
@@ -8,6 +8,10 @@ set(CMAKE_AUTOMOC ON)
 find_package(Qt5Widgets REQUIRED)
 find_package(Qt5Test REQUIRED)
 
+if (WIN32 AND USE_STATIC_QT)
+	add_definitions(-DUSE_STATIC_QT)
+endif ()
+
 set(SOURCES shellcontents.cpp helpers.cpp shellwidget.cpp konsole_wcwidth.cpp)
 add_library(qshellwidget STATIC ${SOURCES})
 target_link_libraries(qshellwidget Qt5::Widgets)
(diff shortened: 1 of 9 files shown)


equalsraf:
> We follow a basic assumption: the Latin character is one character width and the non-Latin character is two characters width.

The unicode spec already defines which chars are half and full width. This is how the GUI and nvim know the correct width - nvim also needs to know this to move the cursor correctly.

- Neovim uses these tables https://github.com/neovim/neovim/tree/master/unicode
- neovim-qt gets the same information from https://github.com/equalsraf/neovim-qt/blob/master/third-party/konsole_wcwidth.cpp

Both of them must be the same. Right now both consider ❶ to be of width 1.

zhaozg:
Thanks, I will do more check.

zhaozg:
OK, according. http://www.unicode.org/reports/tr11/ 

```
The Unicode Character Database [UCD] assigns to each Unicode character as its default width property one of six values: Ambiguous, Fullwidth, Halfwidth, Narrow, Wide, or Neutral (= Not East Asian). For any given operation, these six default property values resolve into only two property values, narrow and wide, depending on context.
```
and 
image show relation.

And we mapping https://github.com/neovim/neovim/blob/master/unicode/EastAsianWidth.txt#L17
to width.

-  "A"   width is 2.
- "F" and "W"  width is 2
- "H", "N", "Na" width is 1

https://github.com/neovim/neovim/blob/master/unicode/EastAsianWidth.txt#L1659
E000..F8FF is A , width is 1 or 2. So we choice to is safe enough.

I will do some change for this PR.

zhaozg:
Neovim to handle this use https://github.com/neovim/neovim/blob/master/src/nvim/mbyte.c#L455-L462

zhaozg:
I have new RP. https://github.com/equalsraf/neovim-qt/pull/388

PR #388, opened 2018-01-28
QChar of CJK Chacater width fixed.

zhaozg:
Replace https://github.com/equalsraf/neovim-qt/pull/386


zhaozg:
#384, #110, #140 should be fixed apply the patch.

zhaozg:
@equalsraf How about this?

zhaozg:
ping @equalsraf ？

equalsraf:
Dont you have problems with cursor placement with this? e.g.
nvim-qt-388

Applying this patch here without changing nvim will always cause issues with cursor movement, because nvim still thinks the char only takes a single cell.

Can you try using `set ambiwidth=double` first?

zhaozg:
@equalsraf you are right, please keep this PR open, I'll report bug to neovim, thanks

justinmk:
ambiwidth is sent in the UI event `option_set` in Nvim 0.2.3

zhaozg:
nvim works fine，just nvim-qt need handle ambiwidth correctly

equalsraf:
> ambiwidth is sent in the UI event option_set in Nvim 0.2.3

Cool. Then we should be able to get the option in here https://github.com/equalsraf/neovim-qt/blob/master/src/gui/shell.cpp#L599 and store it.

We would need to update the shellwidget. Maybe add an ambiguos option to Cell::setChar and wcwidth. The tricky part is that we may need to recalculate the entire shell contents.

PR #389, opened 2018-01-30
GUI: Increase default timeout from 10 to 20s

equalsraf:
Originally reported in https://github.com/neovim/neovim/issues/7934#event-1449035408 by @adelarsq.


adelarsq:
``nvim-qt --timeout 20000`` didn't work. _neovim-qt_ just waits for nearly 10 seconds.

adelarsq:
@equalsraf  The problem is that the timeout argument is being ignored I think.

equalsraf:
@adelarsq yeah it is being ignored for the initial setup. I'll need to fix that. Does the build from this PR help with your problem?

adelarsq:
Still don't. But it loads more things (plugins) now:

image

On the logs shows that the editor was starting and was stopped:
```
Searching for "C:\Users\myuser\.vim\plugged\vim-maven-syntax\autoload/wintabs/element.vim"
Searching for "C:\Users\myuser.
```

PR #419, opened 2018-06-20
Keep arguments to nvim before any filenames; fixes #417.

### LINKED ISSUE #417: Extra buffer opening with neovim 0.3.0
I launch `neovim-qt` via this `zsh` function:

```
function gvim {                                                                 
    nvim-qt --geometry 970x512 --no-ext-tabline -- "$@" &                       
}   
```
I updated to `neovim 0.3.0` and when calling the above, I get a window to open just fine but there are two buffers: an empty one and another buffer the one with the file.

screenshot

roflcopter4:
As noted in issue #417, Neovim v3.* inexplicably seems to open an empty buffer when option parameters appear after filenames. This could be solved by adding a `--` parameter before the filenames, but the easier thing to do is to simply move nvim-qt's additional arguments before any user specified ones and let Neovim sort out this strange bug. Although it may not be 100% reproducible, at least a few people besides myself have noticed this, and this tiny patch seems to fix it without causing any problems I can think of, so it should be harmless at worst.

I also added one small comment (and initially committed with spaces instead of tabs: fixed).

EDIT: I have absolutely no idea why the build failed in the Windows test environment. I didn't change anything that has anything to do with that... I barely changed anything at all.

### DIFF
--- a/src/neovimconnector.cpp
+++ b/src/neovimconnector.cpp
@@ -195,12 +195,12 @@ NeovimConnector* NeovimConnector::spawn(const QStringList& params, const QString
 {
 	QProcess *p = new QProcess();
 	QStringList args;
+        // Neovim accepts a `--' argument after which only filenames are passed.
+        // If the user has supplied it, our arguments must appear before.
 	if (params.indexOf("--") == -1) {
-		args.append(params);
 		args << "--embed" << "--headless";
+		args.append(params);
 	} else {
-		// Neovim accepts a -- argument after which only
-		// filenames are passed
 		int idx = params.indexOf("--");
 		args.append(params.mid(0, idx));
 		args << "--embed" << "--headless";
--- a/src/neovimconnector.cpp
+++ b/src/neovimconnector.cpp
@@ -195,8 +195,8 @@ NeovimConnector* NeovimConnector::spawn(const QStringList& params, const QString
 {
 	QProcess *p = new QProcess();
 	QStringList args;
-        // Neovim accepts a `--' argument after which only filenames are passed.
-        // If the user has supplied it, our arguments must appear before.
+	// Neovim accepts a `--' argument after which only filenames are passed.
+	// If the user has supplied it, our arguments must appear before.
 	if (params.indexOf("--") == -1) {
 		args << "--embed" << "--headless";
 		args.append(params);


justinmk:
Did you try latest Nvim HEAD?  As mentioned in https://github.com/equalsraf/neovim-qt/issues/417#issuecomment-397908803 I am wondering if https://github.com/neovim/neovim/pull/8576 (which was merged yesterday) resolves this without the need for a workaround in nvim-qt.

This would be helpful to know because it may affect other GUIs.

src/neovimconnector.cpp
```diff
@@ -195,12 +195,12 @@ NeovimConnector* NeovimConnector::spawn(const QStringList& params, const QString
 {
 	QProcess *p = new QProcess();
 	QStringList args;
+	// Neovim accepts a `--' argument after which only filenames are passed.
+	// If the user has supplied it, our arguments must appear before.
```
justinmk: This comment is definitely true, and if something like this was being constructed before:

    nvim -- --embed --headless

Then nvim should have opened files literally named `--embed` `--headless`.

roflcopter4:
It was certainly worth a shot to recompile nvim (which in Gentoo takes all of `emerge neovim` to source, compile and install -- I love Gentoo). It didn't help. Nvim-qt with my little patch works, but with it a blank buffer is still opened. The only other Nvim gui that isn't utter garbage that I can think of is Eovim (which is too buggy to use, and uses EFL which has awful font support) doesn't suffer any problems either way.

The arguments nvim-qt was supplying previously looked like 
```
--cmd let &rtp.=',/usr/local/share/nvim-qt/runtime' --cmd set termguicolors FILE_GOES_HERE --embed --headless
```
while this patch changes it to
```
--embed --headless --cmd let &rtp.=',/usr/local/share/nvim-qt/runtime' --cmd set termguicolors FILE_GOES_HERE
```
In neither case does `--` appear. That's why I consider it a bug in nvim. Both should work without problems. The only explanation I can think of is a problem in neovim's option parsing. Regardless, putting everything before the files is a functional temporary fix.

justinmk:
Thanks. It definitely would be a Nvim bug, but I can't reproduce it. I tried this:

    NVIM_LISTEN_ADDRESS=foo nvim --cmd "let &rtp.=''" --cmd "set termguicolors" README.md --embed --headless

Then connect to the `foo` server with the python client:

```
>>> import neovim
>>> n = neovim.attach('socket', path='foo')
>>> n.request('nvim_command_output', 'ls!')
'  1 %a   "README.md"                    line 1'
```

justinmk:
Yes, Nvim HEAD.

justinmk:
@equalsraf If you have a chance, would be helpful to get a tagged release which includes this fix, so that Nvim can ship it to Windows users.

equalsraf:
Tag in place but it seems appveyor builds are broken.

PR #428, opened 2018-07-23
Appveyor change Qt to version 5.11.1

equalsraf:
It seems the builds were failing because version 5.11.0 was not available at the expected path.


equalsraf:
Closing, will merge in #429

PR #433, opened 2018-07-25
Add StartupWMClass=nvim-qt to .desktop file

### LINKED ISSUE #401: WM_CLASS Should be neovim-qt
Hi.
When using `xprop`, it detects neovimQt as `.nvim-qt-wrapped`, which is not clean. It should be `neovim-qt` (`WM_CLASS(STRING) = ".nvim-qt-wrapped", ".nvim-qt-wrapped"`).
For example, when I _xprop_ quteBrowser, it prints `qutebrowser` (`WM_CLASS(STRING) = ".qutebrowser-wrapped", "qutebrowser"`).
Thank you.

equalsraf:
Fixes #401


equalsraf:
This is not working for me with gtk-launch. But it seems we no longer need it.

Issue #298, opened 2017-06-22
Integrate the clipboard provider into Neovim-Qt


xelra:
We talked about this some time ago on IRC, but I felt that it might be best to put the feature request down here.

I'm using Neovim-Qt mostly to connect to remote Neovim sessions over SSH. One of my initial motivations to do so was because I thought that would solve the clipboard problem. Unfortunately the clipboard is still handled by the remote Neovim session.

It would be great if the `+` and `*` registers would point to the local system clipboard, the one running Neovim-Qt.

Thank you very much for considering.

equalsraf:
Waiting on https://github.com/neovim/neovim/pull/6030 https://github.com/neovim/neovim/issues/6029.

If possible implement the provider in the shim.

Shougo:
Implemented the feature in neovim!

skt041959:
Hi @equalsraf 

When will the tb-clipboard branch merge into master branch?

equalsraf:
FYI #479 has been merged.It is NOT enabled by default. You need to call `GuiClipboard` to enable it

```
call GuiClipboard()
```

This essentially sets `g:clipboard`.

I plan to tag a new release with this so everyone can test it. Keep in mind this requires the master branch from nvim (0.3.2)

equalsraf:
Dont know if anyone has been using or following this. So here is a quick update has a note for me to follow later on.

There are a couple of problems with this right now

- [ ] neovim caches provider availability (search for CHECK_PROVIDER in eval.c) which means that if `GuiClipboard()` is called too late (TM) then attempting to access the clipboard always with an error about no clipboard being available, even if `checkHealth` says otherwise.
- [ ] neovim-qt sets the channel information about the clipboard too late, meaning that calling `GuiClipboard` in ginit.vim ALWAYS fails to setup the clipboard

The second part is already in #498. It is still unclear to me if the first part can or should be addressed in the shim.

TLDR; :man_facepalming:

kwon-young:
I've also been trying to use this but I can't seem to make it work when connecting to a remote neovim using `nvim-qt --server`.
Here the series I command run:
```bash
# on server foo
ky@foo $ nvim --listen foo:7777 --headless
```
```bash
# on local machine bar
ky@bar $ nvim-qt --server foo:7777
# in nvim-qt
: call GuiClipboard()
```
When running `:checkhealth` after calling `GuiClipboard`, neovim report that it found a `custom` clipboard provider.
[...]

PR #436, opened 2018-07-26
[WIP] Implement clipboard provider

equalsraf:
A clipboard provider implemented by the GUI

It currently supports both clipboards and stores the selection type. Big question now is what happens with multiple GUIs attached, etc.

This is based on the approach from [neovim-gtk](https://github.com/daa84/neovim-gtk/blob/master/runtime/plugin/nvim_gui_shim.vim#L7). As it stands the call signatures are not compatible, ~~but I think they could be.~~ we use all the arguments in the set event and the api in neovim-gtk only passes arg0 and 2. The event names are different so there should be no conflict.

## TODO

- [x] get the correct channel id instead of **1**
- [ ] Test on windows with multi-line data on the clipboard (newline char)
- [x] ~~Add guards to the provider, merge with the shim, etc~~
- [x] Remove from the shim for now
- [ ] There are some bugs on the provider, sometimes it returns invalid data, i.e. our function is returning an unexpected data structure
- [ ] Add tests - in particular cover different types of selections

ref https://github.com/equalsraf/neovim-qt/issues/298

### DIFF
--- a/contrib/appveyor.yml
+++ b/contrib/appveyor.yml
@@ -22,7 +22,7 @@ configuration:
 matrix:
   fast_finish: false
 install:
-- ps: appveyor DownloadFile -FileName Neovim.zip "https://ci.appveyor.com/api/projects/neovim/neovim/artifacts/build/Neovim.zip?branch=master&job=Configuration%3A%20MINGW_64"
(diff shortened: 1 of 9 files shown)


equalsraf:
Replaced by https://github.com/equalsraf/neovim-qt/pull/479

Issue #438, opened 2018-08-03
Can't minimize nvim-qt with C-z


SalahuddinAsh:
I'm running neovim-qt on Windows. When I try to minimize the window by pressing `C-z`, the window only flashes but nothing happens. I tried the same with gvim and it works. How can I get it to work with neovim-qt?

equalsraf:
This corresponds to the `suspend` command. Apparently we are not handling the corresponding event. I've opened #439 to add this.

On another note, in windows a shortcut key is set up for bringing back the window `<C-A-Z>` but this seems to be a property of the shortcut http://vim.wikia.com/wiki/Minimize_and_restore_gvim_under_Windows_using_shortcut_keys

equalsraf:
`:stop` supported was added into master.

SalahuddinAsh:
@equalsraf 
Thanks for your concern.
I've no idea how can I update nvim on Windows to include the added code? Could you please tell me?
Regarding the shortcut key, I've tried it already but had no effect. Don't know why. I also tried it for gvim but didn't work.

SalahuddinAsh:
I found a work-around that does the job for me. 
I attached `nvim-qt` to the the task bar to be the second left icon. 
I can now launch it/minimze it/restore it by pressing (Windows-2) buttons.
However, I still hope to learn how to get the supported `:stop` in my nvim.

PR #439, opened 2018-08-03
GUI: Minimize window on suspend event

equalsraf:
The suspend/stop() command in the console puts nvim in the background.
In the GUI it typically minimizes the window (gvim).

See #438


equalsraf:
@SalahuddinAsh I have tested this in ubuntu with gnome, and it works for all of the following

- pressing `C-z`
- calling `:stop`
- calling `:suspend`

can you check this in your system?

Issue #441, opened 2018-08-09
Display additional information in popup menu


equalsraf:
Currently the popup menu only displays the text being completed, but additional information may be available (extra, kind, info) that should be displayed too.

ref https://github.com/equalsraf/neovim-qt/pull/349#issuecomment-411699156

clason:
Here are some examples:

### minimal vimrc

```viml
call plug#begin('~/.vim/plugged')
Plug 'lervag/vimtex'
call plug#end()
```

### test files

`test.bib` (needs to be in the same directory)
```tex
@article{A1,
  author = {Author, A.},
  journal = {Good Journal},
  year = {2001},
  title = {Alpha},
}
@article{A2,
  author = {Author, A.},
  journal = {Very Good Journal},
  year = {2008},
  title = {Bravo},
}
@article{B3,
  author = {Buthor, C.},
  journal = {Also Good Journal},
  year = {2004},
  title = {Charlie},
}
```

`test.tex` (the main file)
```tex
\documentclass{article}
\begin{document}

\bibliographystyle{plain}
\bibliography{test.bib}
\end{document}
```

Classical pum:
![screenshot_20180809_120316](https://user-images.githubusercontent.com/2361214/43892511-7154a060-9bcc-11e8-9a2e-8308b2320f9b.png)
![screenshot_20180809_120342](https://user-images.githubusercontent.com/2361214/43892512-71706b74-9bcc-11e8-9035-bb52475caa2c.png)
New pum:
![screenshot_20180809_120412](https://user-images.githubusercontent.com/2361214/43892513-7191e16e-9bcc-11e8-99c5-7399e9f79ecc.png)
![screenshot_20180809_120432](https://user-images.githubusercontent.com/2361214/43892514-71ace75c-9bcc-11e8-99dd-ded4d1f993e4.png)

PR #442, opened 2018-08-09
Show additional information in the popup menu

equalsraf:
ref https://github.com/equalsraf/neovim-qt/issues/441

### DIFF
--- a/src/gui/popupmenumodel.cpp
+++ b/src/gui/popupmenumodel.cpp
@@ -5,7 +5,7 @@
 namespace NeovimQt {
 
 PopupMenuModel::PopupMenuModel(QList<PopupMenuItem> l)
-:QAbstractTableModel(), m_data(l)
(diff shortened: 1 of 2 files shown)


equalsraf:
@clason this should handle feature parity.

clason:
Yeah, sorry, I saw too late that this was a bug, not a "not yet implemented". Please ignore my comment about opt-in vs. opt-out.

equalsraf:
I've fixed the order. But I cant change the padding right away.

clason:
Thank you! I'll see about the padding; I believe the padding is already introduced by the plugin before the information is passed to the pum (at least for some fields), so it might be that `QString` strips the spaces or somesuch.

equalsraf:
See https://github.com/equalsraf/neovim-qt/pull/443 for implementing a custom delegate for the pum. At that point you can implement whatever widget or layout you want.

Issue #449, opened 2018-08-26
Avoid the need for "--" to separate nvim-qt arguments from file names


justinmk:
Currently nvim-qt requires `--` to know which arguments to send to `nvim.exe`.

It would be better to whitelist things like `--geometry`, and send everything _else_ to `nvim.exe`.

One problem with the `--` scheme is that it makes it difficult to associate `nvim-qt.exe` with filetypes. Because the OS sends the file list to the program, it doesn't know that `--` must be prepended.

ref https://github.com/neovim/neovim/issues/7222#issuecomment-416017347

equalsraf:
Some open points in my mind on this:

- what happens with `--help` or `--version` (currently this option does not exist but it was requested before) - I assume we don't forward these
- or more generally what happens when the user calls a command that does not start nvim (no point on stressing too much on this, the caller can already call `nvim-qt -- --help` for a guaranteed fail)
- this will likely break the current usage with two occurrences of `--`

> ref neovim/neovim#7222 (comment)

I'll need to have a closer look on this one, but I suspect the double `--` is the same bug discussed in other issues. To be clear - it should not be necessary to use `--` when passing file arguments. The use of cmd in there also seems strange to me, but maybe that is normal.

justinmk:
> what happens with `--help` or `--version` (currently this option does not exist but it was requested before) - I assume we don't forward these

I think we should forward them (see below), and skip `nvim-qt.exe` entirely. But `--version` and `--help` are special cases, nvim-qt might want to print its help/version info to stdout before passing control to `nvim.exe`.

> more generally what happens when the user calls a command that does not start nvim ... [for example] `nvim-qt -- --help`

Any case where the Nvim UI is not started should also not start the Nvim-Qt GUI (`exec()` to `nvim`). For example, Ex-mode probably won't work well with a GUI (and doesn't seem useful for a GUI). 

    echo "foo" | nvim-qt.exe -Es

iamkarlson:
Hi guys, 
Seems it's not a really popular issue. I suppose that it just takes some time to realize that if you want to open a file in neovim you have to pass two double dashes. 

```
❯ nvim-qt.exe settings.json # doesn't work
❯ nvim-qt.exe -- settings.json # doesn't work
❯ nvim-qt.exe -- -- settings.json # works oO
```

If that issue is discussed somewhere else could you point me there? Cuz it's really annoying.

equalsraf:
Hi @iamkarlson 

[...]

Issue #275, opened 2017-05-06
Open input from stdin


SevereOverfl0w:
`echo hi  | nvim -` opens `hi` in nvim buffer.

`echo hi | nvim-qt -` errors. Also `echo hi | nvim-qt --nofork -` errors. Error with nofork is:

```
QSocketNotifier: Invalid socket 9 and type 'Write', disabling...
Request 2 timed out: 76
Neovim fatal error "Neovim is taking too long to respond"
```

equalsraf:
The problem here is that (for the default case) nvim's stdin is already in use, because nvim-qt starts nvim as `nvim --embed`, the --embed option uses stdin/stdout as an rpc channel between nvim and nvim-qt.

So simply passing the stdin is not an option, unless we somehow change the startup sequence (like https://github.com/equalsraf/neovim-qt/issues/50 suggests) e.g. it is possible to write a vimscript to start the gui

```vim
" start-nvim-qt.vim
let job = rpcstart('./nvim-qt', ['--embed', '--nofork'])
```

and then, this should work

```shell
echo hello | nvim --cmd "source start-nvim-qt.vim" --headless -
```

Bear in mind that this example is lacking in that it does not ensure the GUI shim is loaded and so GuiFont/ginit.vim and other features will not work properly as is.

facundobatista:
Have been considered the case of having other sockets to RPC on?

Like, starting (for example `nvim --embed --rpc-sockets=foo,bar`, and then communicate with nvim through those sockets, not stdin/stdout. 

Of course, nvim would need to grow support for this, but it shouldn't be too hard (at least in POSIX systems).

Thanks!

justinmk:
@facundobatista  Nvim already supports that, so I don't understand the question?

facundobatista:
@justinmk hola! So you say that nvim already supports RPC on different channels than stdin/stdout? If that's the case, we only need to improve neovim-qt to use other channels, leaving stdin free?

justinmk:
:help serverstart()

facundobatista:
@equalsraf is it possible to make nvim-qt to use other socket, as @justinmk is indicating, so we release stdin to be used as text input for the buffer?

equalsraf:
> @equalsraf is it possible to make nvim-qt to use other socket

Probably not anymore. Some things changed since this issue was opened.

[...]

PR #455, opened 2018-08-31
Launch nvim-qt from nvim

equalsraf:
ref https://github.com/equalsraf/neovim-qt/issues/449 https://github.com/equalsraf/neovim-qt/issues/275

- [ ] rtp is not set i.e. the shim is not loaded
- There is no way to restart nvim from the GUI
- [ ] There is no way to move the process to the background except through the calling shell
- [ ] Currently GUI stderr is sent to the nvim stderr, this might not be what you want
- [ ] Define global variables so that we can just use a function call or a command `nvim --headless -c GuiStart`
- [ ] Does `--embed` work properly on windows?
 
Examples

```sh
nvim --headless -c 'call GuiStart(["nvim-qt", "--nofork", "--embed"], {"rpc":1})'
```

If for some reason `--embed` is not working, use `['nvim-qt', '--nofork', '--server', $NVIM_LISTEN_ADDRESS]` instead

### DIFF
--- a/src/gui/runtime/doc/nvim_gui_shim.txt
+++ b/src/gui/runtime/doc/nvim_gui_shim.txt
@@ -132,6 +132,15 @@ GuiClose() notifies the GUI that it should close. The GUI may or may not
 respect this request. The shim setups an autocommand to call this function
 when the |'VimLeave'| event occurs.
 
+GuiStart({cmd}, {opts})					*GuiStart()*
+GuiStart is a wrapper around jobstart that launches a GUI process. When
+the GUI process exists, nvim will attempt to exit using :qa.
+The argument {cmd} specifies the command to be executed.
+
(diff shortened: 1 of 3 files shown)


src/gui/runtime/doc/nvim_gui_shim.txt
```diff
@@ -132,6 +132,15 @@ GuiClose() notifies the GUI that it should close. The GUI may or may not
 respect this request. The shim setups an autocommand to call this function
 when the |'VimLeave'| event occurs.
 
+GuiStart({cmd}, {opts})					*GuiStart()*
+GuiStart is a wrapper around jobstart that launches a GUI process. When
+the GUI process exists, nvim will attempt to exit using :qa.
```
justinmk: typo: exits

Issue #391, opened 2018-02-05
Requires 


DustinVenegas:
# System Information

  - *Operating System:* Windows
  - *Build from Source:* No
  - *Binary Version:* Version 0.2.2 from (Chocolatey)[https://chocolatey.org/packages/neovim]. Looking at the `nvim-qt.exe` properties provides a Date Modified of 2017-11-18 14:39, but per #255, there's no version number in Properties,

# Problem Description and Reproduction Steps

I am using multiple monitors on Windows 10. The primary monitor is set to scale 200% at 2880x1800. Two other secondary monitors scale 100% at 1920x1080. 

Starting and running `nvim-qt.exe` on the first monitor works as intended. 

![image](https://user-images.githubusercontent.com/282890/35834628-849636a2-0a8b-11e8-88a6-b43ee64c95bf.png)

Moving or running `nvim-qt.exe` on the secondary monitors introduces scaling artifacts. 

![image](https://user-images.githubusercontent.com/282890/35834642-911f5dcc-0a8b-11e8-949c-b67ecbabf15e.png)

On the default installation through Chocolatey, there are no DPI scaling overrides. 

![image](https://user-images.githubusercontent.com/282890/35834674-b40b9d1e-0a8b-11e8-8348-299bb4c1575b.png)

Turning the setting marked "Override high DPI scaling behavior. Scaling performed by:" to "System (Enhanced)" appears to fix the issue. 

![image](https://user-images.githubusercontent.com/282890/35834744-ec53a482-0a8b-11e8-9386-3fabf0bb38cf.png)

# Conclusion

Application Scaling looks like the default option. I hope that QT has an easy way to add application-level scaling easily that can be toggled. If not, we might have to consider setting this property on installation. 

Is there a recommended documentation page I should append this to for future spelunkers?

equalsraf:
Thanks for reporting this, please leave the issue open so people will see it on a search.

Going trough the Qt docs it seems we will need to do some changes to support hidpi http://doc.qt.io/qt-5/highdpi.html#migration-of-existing-applications some of the environment vars there might help too.

Zyst:
Can also reproduce this issue. Indeed found it through google search.

![nvim-qt_2018-04-22_19-59-41](https://user-images.githubusercontent.com/3832380/39102301-2621a0b6-4668-11e8-8160-92e092fa685e.png)

[...]

PR #459, opened 2018-09-10
Enable HiDpi scalling

equalsraf:


equalsraf:
Tested this with linux multimonitor. It did not notice any difference :D

equalsraf:
Well apparently this does not work on windows https://github.com/equalsraf/neovim-qt/issues/391#issuecomment-420066543

equalsraf:
@youngmit pushing a new build, kind of a dirty hack but hopefully this one will work for windows.

equalsraf:
In *theory* it should be ok. Based on the Qt docs and open issues in their bugtracker we might run into other issues depending on which APIs we are using, that may or may not return the expected sizes (e.g. the is multiplied by a factor).

For now I only enabled this for Windows. There are some more open [issues](https://bugreports.qt.io/browse/QTBUG-52196?jql=status%20in%20(Open%2C%20Reported)%20AND%20text%20~%20%22HiDPI%22) in windows and other systems. The most significant in our case would be related to widget positions and maybe icons, since we dont use opengl.

stu-b-doo:
Is this fix available yet? I'm facing issue #391 . Not really sure how to build it from source to test it out

PR #460, opened 2018-09-16
[RFC] Set v:windowid

### LINKED ISSUE #64: v:windowid is not defined

equalsraf:
Is it safe to use vim_command to do this? I hope so. Need o test this.

Fixes #64

### DIFF
--- a/src/gui/runtime/doc/nvim_gui_shim.txt
+++ b/src/gui/runtime/doc/nvim_gui_shim.txt
@@ -80,6 +80,8 @@ override the window behaviour.
 
 							*g:GuiWindowId*
 g:GuiWindowId holds the window id (X11) or the window handle (Windows).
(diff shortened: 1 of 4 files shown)


justinmk:
Why wouldn't it be safe? It will block if user is at a hit-enter prompt, if that's what you mean.

equalsraf:
Basically that, but it is a non blocker.

Issue #425, opened 2018-07-17
after type key, can't shutdown nvim with `X` button


zgpio:
On windows, after I type `<c-w>`,  I can't shutdown nvim with `X` button immediately.
 
![default](https://user-images.githubusercontent.com/19503791/42791312-9a66ad7c-89a2-11e8-8fd2-fd3afa15f6ca.png)

equalsraf:
Confirmed in Linux. To reproduce

1. Open nvim-qt
1. Type `<C-w>`
1. Attempt to close the window either with  the window close button or Altf+f4/etc

The window will not close. If you press `j` the window will then close.

It would seem we are blocked waiting for the rest of the `<C-w>` command, before the call to vim_command is handled.

zgpio:
Although this button is not used often, this should indeed be fixed.

equalsraf:
@zgpio started testing a fix for this in #462.

PR #462, opened 2018-09-17
[RFC] Input Ctrl-C if Neovim is blocked waiting for input

equalsraf:
Currently this is only being used for #425. If the user is trying to close the window, we send in a Ctrl-C if neovim is blocked waiting for input.

We might use this in other cases too, but I have not looked into this further.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -1034,6 +1034,7 @@ void Shell::closeEvent(QCloseEvent *ev)
 		// If attached to a spawned Neovim process, ignore the event
 		// and try to close Neovim as :qa
 		ev->ignore();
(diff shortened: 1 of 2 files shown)


justinmk:
https://github.com/neovim/neovim/pull/8754 is intended to solve the "startup" case of this issue.

equalsraf:
I see. In any case https://github.com/neovim/neovim/pull/8754 would only work for `--embed` so it is worth adding a solution for the general case.

oni fixes it like [this](https://github.com/onivim/oni/blob/a4c8515d62683e8cbc5bbafce34981f114aad697/browser/src/neovim/NeovimInstance.ts#L711). It also tries to display error messages.

Issue #363, opened 2017-12-02
"nvim-qt +mak" is not responsive


shlomif:
"nvim-qt +mak" is not responsive and keeps saying:

```
Neovim fatal error "Neovim is taking too long to respond"
Neovim API version compatible with 0 supported 3
Request 1 timed out: 34
```

"nvim +mak" and "nvim-qt +echo" are fine. I am using it on xfce on Mageia v7 x86-64. I am using git commit 539522cf19633163f72a8c567960b06e133048da of neovim-qt and a recent commit of neovim.

jgehrig:
This scenario works now. Marking as closed.

Issue #322, opened 2017-08-17
Neovim-qt is not responding awaiting user input


iamkarlson:
Neovim-qt is hanging when neovim server is awaiting user input (most common swap is already exists)

jgehrig:
This issue has been fixed upstream in neovim as per discussions in #463.

Marking this issue as closed.

Issue #78, opened 2015-12-16
When neovim gets stuck while initializing neovim-qt does not respond


RedX2501:
I am testing out my .vimrc with neovim.

First i transferred it to the new XDG_CONFIG_HOME location and renamed it.

Now when i tried to start neovim-qt it would just state that neovim was not responding.

After some fiddling in the command line i was able to start neovim and it printed:

```
Error detected while processing C:\Users\uidw8099\AppData\Roaming\nvim\init.vim:
line   90:
E484: Can't open file d:\Playground\neovim\Neovim\bin\syntax\syntax.vim
line  563:
E185: Cannot find color scheme 'zenburn'
Press ENTER or type command to continue
```

Well now it is clear why neovim-qt said it was stuck but there was no indication on neovim-qt's window.

Is there any way to forward this console information from neovim to the user?

equalsraf:
Not much that can be done for this right now, its being fixed upstream (https://github.com/equalsraf/neovim-qt/issues/50#issuecomment-144577335,  neovim/neovim#3646)

equalsraf:
ref https://github.com/neovim/neovim/issues/3901

shlomif:
I'm just subscribing to this bug based on https://github.com/equalsraf/neovim-qt/issues/333 being closed as a dup of this one.

equalsraf:
> I'm just subscribing to this bug based on #333 being closed as a dup of this one.

:D FYI there should be a subscribe button on the right sidebar.

dumblob:
I've just reported a "subissue" of this one on https://github.com/neovim/neovim/issues/8038#issuecomment-367348125 . But it seems it's a more general issue and fits exactly here.

Any news on this?

equalsraf:
Added a patch for this in https://github.com/equalsraf/neovim-qt/pull/463. The only way I found to handle this case was to insert a `<C-c>` in the input.

PR #463, opened 2018-09-17
Handle blocked input

equalsraf:
Continues https://github.com/equalsraf/neovim-qt/pull/462

Related
- https://github.com/equalsraf/neovim-qt/issues/322
- https://github.com/equalsraf/neovim-qt/issues/363
- https://github.com/equalsraf/neovim-qt/issues/78

### Swap files :o:

To reproduce just open the same file twice with nvim.

In terminal nvim simply pressing `<C-c>` is not sufficient to dismiss the last error message. However in nvim-qt the behaviour seems to differ. More importantly as it stands the user never sees the error message, and is left with a Read Only buffer.

In other words we are better off than we were before (we dont just hang). But it is pretty cryptic for the user to figure out what happened.

### Startup errors :heavy_check_mark: 

To reproduce call `nvim-qt --cmd AAAAAA`. Or add some error command to init.vim.

Both cases work. ~~The primary issue is that the user gets no feedback. Errors are kept in `:messages`.~~

### Blocking on external program :x:

One case I've been unable to solve is `nvim-qt -- --cmd '!sleep 100'`. I never get a response for `nvim_get_mode`. The same can also be observed when attaching to an instance running a command.

Sending a `<C-c>` without the call to `nvim_get_mode()` cancels ongoing calls, but this might not be what we want.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -791,7 +791,7 @@ void Shell::paintEvent(QPaintEvent *ev)
 
 void Shell::keyPressEvent(QKeyEvent *ev)
 {
-	if (!m_nvim || !m_attached) {
+	if (!m_nvim) {
 		QWidget::keyPressEvent(ev);
 		return;
 	}
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -239,6 +239,8 @@ void Shell::init()
 	connect(req, &MsgpackRequest::finished,
 			this, &Shell::setAttached);
 
+	bailoutIfinputBlocking();
+
 	// Subscribe to GUI events
 	m_nvim->api0()->vim_subscribe("Gui");
 }


jamessan:
This should natively work better with 0.3.2, since https://github.com/neovim/neovim/pull/8754 has been merged.

justinmk:
> One case I've been unable to solve is `nvim-qt -- --cmd '!sleep 100'`. I never get a response for nvim_get_mode

I think we do `os_breakcheck` during a shell command (otherwise CTRL-c wouldn't work), but currently `enable_events=0`. 
 `nvim_input('<c-c>')` should work ....

equalsraf:
To be clear. While these patches dont do it yet. There is no reason why we cannot give the user some feedback when neovim is blocked for input.

> We do os_breakcheck during a shell command (so nvim_input('<c-c>') works), ... I can't think of why nvim_get_mode would not return.

That was what I was expecting.

FWIW I think in the case of the shell command the input is not considered to be blocked. The mode request gets pushed [here](https://github.com/neovim/neovim/blob/47d74bf32f3b95b28a3ef2a7840989eb9bed7480/src/nvim/msgpack_rpc/channel.c#L351).

In which case nvim_get_mode() would not help here, even if it should not block.

bfredl:
I have now tested that neovim master + nvim-qt master supports displaying of initial swap file dialog and `--cmd "echoerr bork"` messages. Unfortunately it no longer works with this branch, this is because early UI neovim/neovim#8754 (in its final implementation) detects an UI embedder by checking if `nvim_ui_attach` comes either as the first RPC call or the second one after `[n]vim_get_api_info` (or possibly as a part of a nested call, as long as that one is first). Currently adding an `nvim_get_mode` before attach disables this behaviour. 

So either neovim-qt will need to avoid doing `get_mode` when using `--embed`, or `nvim_get_mode` will need to support the new `--embed` behavor explicitly. The later would have the benefit of being able to feature detect neovim/neovim#8754, by giving startup its own mode letter, though it will make the description for how --embed works even more complicated. (the initial `--embed-ui` proposal was just "wait for nvim_ui_attach", nothing more...)

bfredl:
> Then non-UI embedders need to make a dummy call to nvim_ui_attach (followed by nvim_ui_detach). 

They can also do `--headless --cmd "call stdioopen({'rpc': v:true})"` whose behavior hasn't changed. 

> Or we could be smarter about waiting, by looking at the nvim_set_client_info "type" field: non-UI clients can opt-out of the wait by calling nvim_set_client_info with "type" other than "ui". Thus, legacy UIs benefit from the new --embed behavior while non-UI clients can opt out of it.

I don't like "smart" when "smart" means "complicated". It would be simpler to just have two different command lines, where we could have `--embed` mean "wait for attach" unconditionally as you said (to make the informed guess that pre-0.3.2 embedders are mostly UIs), and then `--headless --cmd "call stdioopen({'rpc': v:true})"` for an UI-less embedder (though I think it should be a goal to make the UI protocol so flexible and functional that all embedders "want" to use ui_attach)

justinmk:
@bfredl Completely agreed with all of those comments.

> though I think it should be a goal to make the UI protocol so flexible and functional that all embedders "want" to use ui_attach)

Yes, I was in fact thinking of ActualVim and VSCodeVim who are "non UI" embedders but still make use of UI events (and would greatly appreciate richer events such as structured dialogs).

bfredl:
@justinmk perhaps even `--embed --headless` should have the old behavior of `--embed` (no UI assumed)? If `--embed` from now on is assumed to be used by UI, it is not very headless anymore, right? Or is this too far-fetched?

justinmk:
> perhaps even --embed --headless should have the old behavior of --embed (no UI assumed)?

Maybe, but back to your other point, that does make the documentation a bit more complicated.

bfredl:
True, but my idea was that it doesnt imply headless any more, so that line would be removed then.

equalsraf:
In general I don't think the ui should be special (other than some ui specific functions) when it comes to the api.

I've added that `<C-c>` hack in here, but Ideally I would like to avoid it for startup (I do think it makes perfect sense for shutdown).

> of being able to feature detect neovim/neovim#8754, by giving startup its own mode letter

So a mode notification would report the mode as being startup? e.g. `mode_change, "startup"`.

It might be more useful to add a notification with the same info as get_mode, which always gets sent on a change and new connections.

Concerning --headless and --embed. --embed still prevents the use of a terminal interface right? It is only headless with regards to the terminal.

bfredl:
> Concerning --headless and --embed. --embed still prevents the use of a terminal interface right? It is only headless with regards to the terminal.

Correct. `--embed` means stdio will work as a msgpack rpc channel. With my latest proposal `--headless` means there will be no UI at least during startup, and conversely the _absence_ of `--headless` means that there will _be_ an UI during startup. In `--embed` mode that is an external UI, and when not embeded that will be the TUI.

equalsraf:
@bfredl so the meaning of the headless option changed to mean (without TUI) + delayed initialization of the UI. As a consequence --embed no longer implies headless, it only implies the old behaviour of headless (i.e. without TUI).

I'm getting some mixed behaviour here, e.g. for --headless in my terminal

```
$ NVIM_LISTEN_ADDRESS=/tmp/nvimS bin/nvim --headless -u NONE /tmp/testit
E325: ATTENTION
Found a swap file by the name "~/.local/share/nvim/swap//%tmp%testit.swp"
          owned by: raf   dated: Thu Sep 20 18:05:26 2018
         file name: /tmp/testit
          modified: no
         user name: raf   host name: quasar
        process ID: 11911 (still running)
While opening file "/tmp/testit"

(1) Another program may be editing the same file.  If this is the case,
    be careful not to end up with two different instances of the same
    file when making changes.  Quit, or continue with caution.
(2) An edit session for this file crashed.
    If this is the case, use ":recover" or "vim -r /tmp/testit"
    to recover the changes (see ":help recovery").
    If you did this already, delete the swap file "/home/raf/.local/share/nvim/swap//%tmp%testit.swp"
    to avoid this message.
"/tmp/testit" [New File]
```

The swap error message is displayed as output of the nvim process.
When i attach nvim-qt with ` nvim-qt --server /tmp/nvimS` I get a file in read only mode. 

But if I call nvim-qt to start nvim (i.e. --headless --embed), I get to see the swapfile message and prompt.

bfredl:
See https://github.com/neovim/neovim/pull/9024, I will update the description later...

justinmk:
@equalsraf this PR could be closed. https://github.com/neovim/neovim/pull/9024 was merged in Nvim 0.3.2 and is being used successfully in many GUI clients.

equalsraf:
Ack, closing this.

PR #477, opened 2018-11-16
Desktop file improvements

cjsthompson:
* Fix opening files with neovim-qt with the file manager context menu.
* Add german and french translations.

### DIFF
--- a/src/gui/nvim-qt.desktop
+++ b/src/gui/nvim-qt.desktop
@@ -1,8 +1,13 @@
 [Desktop Entry]
 Name=Neovim-Qt
 Comment=Qt GUI for Neovim text editor
+Comment[de]=Grafikschnittstelle für den Neovim Texteditor
+Comment[fr]=Interface graphique pour l'éditeur de texte Neovim
+Keywords=Text;Editor;Plaintext;Write;
+Keywords[de]=Text;Editor;Klartext;Schreiben;
+Keywords[fr]=texte;éditeur;texte brut;écrire;
 MimeType=text/english;text/plain;text/x-makefile;text/x-c++hdr;text/x-c++src;text/x-chdr;text/x-csrc;text/x-java;text/x-moc;text/x-pascal;text/x-tcl;text/x-tex;application/x-shellscript;text/x-c;text/x-c++;
-Exec=nvim-qt -- %F
+Exec=nvim-qt %U 
 Icon=nvim-qt
 Type=Application
 Terminal=false


equalsraf:
Hi @cjsthompson, thanks for the PR.

I have two questions:

- In practice what is main difference between %U and %F.
- Why did you remove the separator `--` from the call? Is this part of the change significant?

cjsthompson:
I think "--" is bourne shell syntax. But here "Terminal=false" so it's at best useless or might cause problems. I think what happened until now is, it passed "--" as a file to open which nvim-qt didn't find of course and so you get an empty nvim-qt window instead of nvim-qt opening the file for which nvim-qt was invoked from it's context menu. Also if your default shell is not a bourne-like shell that will very likely be a problem (I use fish for example). So it's good to keep things shell agnostic.

"%F" is for a list of files and "%U" is for a list of URLs (including files). I think Qt apps have support for opening files with other protocols. So that would allow to open a file accessible through "sftp://" from the file manager context menu for example.

equalsraf:
`--` is not bash syntax, both nvim-qt and nvim use this argument with special meaning

- nvim-qt treats all arguments after `--` as arguments to be passed to nvim. So `nvim-qt -- -u NONE` actually calls `nvim -u NONE`
- nvim considers all arguments after `--` as files regardless of their name, e.g. `nvim -- --false` will open a file named `--false`.

In light of this the correct format should be `nvim-qt -- -- %F`

> "%F" is for a list of files and "%U" is for a list of URLs (including files). I think Qt apps have support for opening files with other protocols.

Keep in mind that these filenames are not handled by Qt at all. They are passed to nvim. nvim does support sftp:// and some other protocols (from `:h netrw`). I have no idea if the file manager uses these accordingly. In any case :+1: if this works as intended.

> I think what happened until now is, it passed "--" as a file to open which nvim-qt didn't find of course and so you get an empty nvim-qt window instead of nvim-qt opening the file for which nvim-qt was invoked from it's context menu

Best way to figure out what is going on here is to inspect process arguments being used i.e.

- Using `ps ax` you should see the arguments passed to nvim-qt, the file arguments are then passed to an nvim process
- From inside nvim (or nvim-qt) you can use `:echo argv()`

I'm currently using nautilus and I can open files using nvim-qt, both one or multiple files at a time. If you are having problems, can you provide some more details on your setup?

jgehrig:
Merged excluding the `%U` changes in Pull Request #862.

PR #479, opened 2018-12-03
[WIP] Implement clipboard provider

equalsraf:
A clipboard provider implemented by the GUI. This requires neovim from the master branch (v0.3.2)

It currently supports both clipboards and stores the selection type. Big question now is what happens with multiple GUIs attached, etc.

 - [x] detect UI support at runtime
- [ ] Test on windows with multi-line data on the clipboard (newline char)
- [x] There are some bugs on the provider, sometimes it returns invalid data, i.e. our function is returning an unexpected data structure
- [ ]  Add tests - in particular cover different types of selections

ref https://github.com/equalsraf/neovim-qt/issues/298

### DIFF
--- a/contrib/appveyor.yml
+++ b/contrib/appveyor.yml
@@ -22,7 +22,7 @@ configuration:
 matrix:
   fast_finish: false
 install:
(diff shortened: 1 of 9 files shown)


equalsraf:
Looking good after fixing some bugs. Let the testing begin.

equalsraf:
Merged into master. Keep in mind this is not enabled by default - you need to call `call GuiClipboard()` to enable it. After some testing we can have a look at improving the shim experience with some proper commands and docs.

PR #480, opened 2018-12-15
surround autocmd with group

wsdjeg:


justinmk:
non-autoload plugins (in the `/plugin/*` directory) may get sourced many times, and each time an additional `VimLeave` event handler will be created, without removing the old ones. 

The PR here is "idiomatic" VimL.

equalsraf:
I'll quote you on that :D

wsdjeg:
Thanks. :)

PR #484, opened 2018-12-21
[WIP] Merge ferranpm project tree

gagbo:
I took over #254 as a starting point for adding a widget to a Qt application. At this point (feac4e6), the only difference with #254 is that I merged v0.2.11 into the branch.

Resolving conflicts meant : 
- changing the `NeovimQt::Neovim` namespace to `NeovimQt::NeovimApi1`
- changing the constructor of TreeView to connect the initialization to the `NeovimConnector::ready` signal

There is still polishing to do :
- Linting. I did not find a style file for the code so I messed up the style while saving the file (ignore whitespace changes for the time being). I'm currently running a format finder to produce a `.clang-format` file from v0.2.11 and I'll just apply that back on my PR (and make another PR to actually add this file to the repo)
- The constructor should not connect to the signal, but instead check if the connector is ready so it can directly initialize if the Treeview is constructed later ?
- Should we provide a better way to modularize the TreeView code (by that I mean provide the option of not compiling this part of the code) ?

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -1,323 +1,345 @@
 #include "mainwindow.h"
 
 #include <QCloseEvent>
-#include <QLayout>
 #include <QToolBar>
+#include <QLayout>
 
 namespace NeovimQt {
 
 MainWindow::MainWindow(NeovimConnector *c, ShellOptions opts, QWidget *parent)
-    : QMainWindow(parent),
-      m_nvim(0),
-      m_errorWidget(0),
-      m_shell(0),
-      m_delayedShow(DelayedShow::Disabled),
-      m_tabline(0),
-      m_tabline_bar(0),
-      m_shell_options(opts),
-      m_neovim_requested_close(false) {
-  m_errorWidget = new ErrorWidget();
-  m_stack.addWidget(m_errorWidget);
-  connect(m_errorWidget, &ErrorWidget::reconnectNeovim, this,
-          &MainWindow::reconnectNeovim);
-  setCentralWidget(&m_stack);
+:QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0),
+	m_delayedShow(DelayedShow::Disabled), m_tabline(0), m_tabline_bar(0),
+	m_shell_options(opts), m_neovim_requested_close(false)
+{
+	m_errorWidget = new ErrorWidget();
+	m_stack.addWidget(m_errorWidget);
+	connect(m_errorWidget, &ErrorWidget::reconnectNeovim,
+			this, &MainWindow::reconnectNeovim);
+	setCentralWidget(&m_stack);
 
-  init(c);
+	init(c);
 }
 
-void MainWindow::init(NeovimConnector *c) {
-  if (m_shell) {
-    m_shell->deleteLater();
-    m_stack.removeWidget(m_shell);
-  }
-  if (m_nvim) {
-    m_nvim->deleteLater();
-  }
+void MainWindow::init(NeovimConnector *c)
+{
+	if (m_shell) {
+		m_shell->deleteLater();
+		m_stack.removeWidget(m_shell);
+	}
+	if (m_nvim) {
+		m_nvim->deleteLater();
+	}
 
-  m_tabline_bar = addToolBar("tabline");
-  m_tabline_bar->setObjectName("tabline");
-  m_tabline_bar->setAllowedAreas(Qt::TopToolBarArea);
-  m_tabline_bar->setMovable(false);
-  m_tabline_bar->setFloatable(false);
-  // Avoid margins around the tabbar
-  m_tabline_bar->layout()->setContentsMargins(0, 0, 0, 0);
+	m_tabline_bar = addToolBar("tabline");
+	m_tabline_bar->setObjectName("tabline");
+	m_tabline_bar->setAllowedAreas(Qt::TopToolBarArea);
+	m_tabline_bar->setMovable(false);
+	m_tabline_bar->setFloatable(false);
+	// Avoid margins around the tabbar
+	m_tabline_bar->layout()->setContentsMargins(0, 0, 0, 0);
 
-  m_tabline = new QTabBar(m_tabline_bar);
-  m_tabline->setDrawBase(false);
-  m_tabline->setExpanding(false);
-  m_tabline->setDocumentMode(true);
-  m_tabline->setFocusPolicy(Qt::NoFocus);
-  connect(m_tabline, &QTabBar::currentChanged, this, &MainWindow::changeTab);
+	m_tabline = new QTabBar(m_tabline_bar);
+	m_tabline->setDrawBase(false);
+	m_tabline->setExpanding(false);
+	m_tabline->setDocumentMode(true);
+	m_tabline->setFocusPolicy(Qt::NoFocus);
+	connect(m_tabline, &QTabBar::currentChanged,
+			this, &MainWindow::changeTab);
 
-  m_tabline_bar->addWidget(m_tabline);
-  m_tabline_bar->setVisible(m_shell_options.enable_ext_tabline);
+	m_tabline_bar->addWidget(m_tabline);
+	m_tabline_bar->setVisible(m_shell_options.enable_ext_tabline);
 
-  m_nvim = c;
+	m_nvim = c;
 
-  m_tree = new TreeView(c);
+	m_tree = new TreeView(c);
 
-  m_shell = new Shell(c, m_shell_options);
+	m_shell = new Shell(c, m_shell_options);
 
-  m_window = new QSplitter();
-  m_window->addWidget(m_tree);
-  m_window->addWidget(m_shell);
+	m_window = new QSplitter();
+	m_window->addWidget(m_tree);
+	m_window->addWidget(m_shell);
 
-  m_stack.insertWidget(1, m_window);
+	m_stack.insertWidget(1, m_window);
 
-  m_stack.setCurrentIndex(1);
-  connect(m_shell, SIGNAL(neovimAttached(bool)), this,
-          SLOT(neovimAttachmentChanged(bool)));
-  connect(m_shell, SIGNAL(neovimTitleChanged(const QString &)), this,
-          SLOT(neovimSetTitle(const QString &)));
-  connect(m_shell, &Shell::neovimResized, this,
-          &MainWindow::neovimWidgetResized);
-  connect(m_shell, &Shell::neovimMaximized, this, &MainWindow::neovimMaximized);
-  connect(m_shell, &Shell::neovimSuspend, this, &MainWindow::neovimSuspend);
-  connect(m_shell, &Shell::neovimFullScreen, this,
-          &MainWindow::neovimFullScreen);
-  connect(m_shell, &Shell::neovimGuiCloseRequest, this,
-          &MainWindow::neovimGuiCloseRequest);
-  connect(m_nvim, &NeovimConnector::processExited, this,
-          &MainWindow::neovimExited);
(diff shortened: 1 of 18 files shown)


gagbo:
I could use some help about the failing tests here.

- The Appveyor tests does not build, I'll check the logs later
- The Semaphore tests fail because of the non-normalization of `:GuiFont Font:b:l` command, but they run fine in my dev environment (which is Fedora Linux, which is probably why Travis works)

I'll fix these later

src/gui/treeview.cpp
```diff
@@ -0,0 +1,79 @@
...
+  connect(m_nvim->neovimObject(), &NeovimApi1::neovimNotification, this,
+          &TreeView::handleNeovimNotification);
+
+  m_nvim->neovimObject()->vim_subscribe("Dir");
+  m_nvim->neovimObject()->vim_subscribe("TreeView");
+}
+
+void TreeView::open(const QModelIndex &index) {
+  QFileInfo info = model->fileInfo(index);
+  if (info.isFile() and info.isReadable()) {
```
equalsraf: &&

equalsraf:
The appveyor build only fails for visual studio, not for mingw, maybe some syntax that particular compiler does not support. I'm kind of surprised the other succeeded using **and** instead of **&&**.

gagbo:
Tests are fixed, the remaining failure was because I didn't catch the font being installed from the `CMAKE_SOURCE_DIR` at the beginning of the test.

Now, for polishing : 

- I'll probably make the `rpcnotify` calls into `Plug` mappings so users can map them where they want instead of having to write the call or use the command (so that means writing the doc too)
- I'll try to find a way to disable the feature on launch (so that means writing the doc too)
- Currently the TreeView widget is changing current directory to `$HOME` on initialization (on `NeovimConnector::ready`), is that an okay behaviour ? I think so, but it's a change so I want to check
- I'm defining `CMAKE_SOURCE_DIR` in the test folder `CMakeLists.txt` to ensure absolute paths to resources instead of using the `WORKING DIRECTORY` of the `add_test` command (since the TreeView widget changes working dir as soon as the connection is ready), I could not think of a faster way to get the path to the test resources

gagbo:
Current state (with uncommitted vimscript changes for the shim plugin). In this gif I'm showing basically all the interaction there is between the tree widget and the neovim instance : 
- toggling the view using neovim commands
- `:cd` in Neovim reflects in the tree
- clicking on a file in the tree opens it in Neovim


peek 22-12-2018 12-47

I forgot to test what happens with splits, my bad. I'll probably look into that at some point.

Also, note that there is no interaction button/info in the tree widget (the current directory in the KWin title is coming from my setting of `titlestring`). Maybe that's a minor issue too (minor because it's not feature-breaking, but still an issue because that's a pretty big lack in usability)

( For those interested in this kind of stuff : 

- Font : Source Code Pro
- Neovim theme : [srcery-colors](https://github.com/srcery-colors/srcery-vim) with [Lightline](https://github.com/itchyny/lightline.vim)
- KDE theme : Kvantum with [KvGlass](https://store.kde.org/p/1201321) but I don't think it really works as intended in the tree widget here

)

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -149,3 +149,32 @@ function GuiClipboard()
           \ }
 	call provider#clipboard#Executable()
 endfunction
+
```
equalsraf: Usually in the shim user facing commands/functions are named as GuiSomething.

equalsraf:
> I'll probably make the rpcnotify calls into Plug mappings so users can map them where they want instead of having to write the call or use the command (so that means writing the doc too)

If you want you can leave them as undocumented functions while you test this further. Provided the default behaviour of the editor remains unchanged.

I dont mind having gated features. For example clipboard integration is currently in the shim but only as a set of undocumented functions, since we are still testing.

> Currently the TreeView widget is changing current directory to $HOME on initialization (on NeovimConnector::ready), is that an okay behaviour ? I think so, but it's a change so I want to check

For now it should be ok.

> I'm defining CMAKE_SOURCE_DIR in the test folder CMakeLists.txt to ensure absolute paths to resources instead of using the WORKING DIRECTORY of the add_test command (since the TreeView widget changes working dir as soon as the connection is ready)

:+1: 

> I forgot to test what happens with splits, my bad. I'll probably look into that at some point.

good point, **cd** may change with the current window (:h lcd). DirChanged does provide info on the scope of the update (tab/window/global). So the real questions is: what directory is the treeview tracking? the global, tab or window directory?

again ff the treeview is disabled by default I'm good with merging it as is until we can figure out the details :)

equalsraf:
It seems double click works to open a file, but not Enter.

src/gui/mainwindow.cpp
```diff
@@ -1,328 +1,323 @@
 #include "mainwindow.h"
 
 #include <QCloseEvent>
```
equalsraf: It seems a lot of this file was reformatted. Can we revert this change and keep just the important bits. It makes this a lot harder to read.

gagbo: Yeah I'll rever as much as I can

src/gui/runtime/plugin/nvim_gui_shim.vim
gagbo: I'll rename the commands then, I thought that it would be fine to namespace them elsewhere since it's different widgets.

Do you also want the <Plug> mappings to be `<Plug>GuiTreeviewName` ?

gagbo:
> good point, cd may change with the current window (:h lcd). DirChanged does provide info on the scope of the update (tab/window/global). So the real questions is: what directory is the treeview tracking? the global, tab or window directory?

Currently the shim subscribes to all DirChanged events, and just calls `getcwd()` without any argument. So according to the doc, it will always be the "effective current working directory", which basically means the TreeView will always update on window/tab change to reflect the current working directory (if the change implies a new working directory, which will trigger DirChanged)

> again ff the treeview is disabled by default I'm good with merging it as is until we can figure out the details :)

I'll disable it by default then, I'm still not sure if I want to hide the widget on construction in c++ code, or send the hide rpc message in the shim

> It seems double click works to open a file, but not Enter.

True, I still cannot say that I "own" this code enough to do this "correctly"; which here means that I'd first like to have :
- a `<Plug>` mapping to change the focus from the `Shell` to the `TreeView`, and
- a (preferably configurable) mapping from the `TreeView` widget back to the `Shell`

But I think that before having keyboard shortcuts to actually go in and out of the widget, it's not really important to have keyboard navigation/selection in the tree

equalsraf:
> I'll disable it by default then, I'm still not sure if I want to hide the widget on construction in c++ code, or send the hide rpc message in the shim

I say hide it by default in C++. Remember that the shim might not be running at all. I frequently run neovim-qt attached to remote neovim instances in which case the treeview could never work, and sometimes the shim is not even available.

gagbo:
I set the default of Treeview to hide, directly in the constructor so it should not even flicker.

Keep in mind that the initialization callback still contains a `setDirectory` function which sets the current directory to `$HOME` (thanks to the 2nd `notify` parameter which is true by default). I never had any issue with that because I never use more that 1 instance of neovim, so I don't know if it's going to be an issue.

src/gui/treeview.cpp
```diff
@@ -0,0 +1,80 @@
...
+namespace NeovimQt {
+TreeView::TreeView(NeovimConnector *nvim, QWidget *parent)
+: QTreeView(parent), m_nvim(nvim) {
+	model = new QFileSystemModel(this);
+
+	setModel(model);
+
+	header()->hide();
+	hide();
+	// FIXME: MAGIC NUMBER!!
```
equalsraf: Whats up with this magic number comment?

src/gui/treeview.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+namespace NeovimQt {
+TreeView::TreeView(NeovimConnector *nvim, QWidget *parent)
+: QTreeView(parent), m_nvim(nvim) {
+	model = new QFileSystemModel(this);
+
+	setModel(model);
+
+	header()->hide();
+	hide();
```
equalsraf: I would prefer to place this hide() in the mainwindow.

equalsraf:
Looking good, comments aside let me know when you are ready to merge.

src/gui/treeview.cpp
gagbo: Columns contain additional information about the files, I'll change that to be more meaningful

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -152,3 +152,32 @@ function GuiClipboard()
...
+augroup guiDirEvents
+    autocmd!
+    autocmd DirChanged * call rpcnotify(0, 'Dir', getcwd())
+    autocmd WinEnter * call rpcnotify(0, 'Dir', getcwd())
+augroup END
+
+
+" Notifies the TreeView widget of a Show or Hide event
+function! s:TreeViewShowHide(show)
+    call rpcnotify(0, 'TreeView', 'ShowHide', a:show)
```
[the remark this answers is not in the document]
equalsraf: We usually do (Gui, EventNmae, args) but I was not too surprised to see it like that since the TreeView is subscribing to that specific event.

equalsraf:
>  The channel used for rpcnotify is the channel 0, maybe using ui_chan is better

channel 0 is a broadcast channel i.e. the message will be sent to all neovim clients.

>  If the rpcnotify commands to interact with the widget could be rpcnotify(ui_chan, 'Gui', 'Treeview', 'ShowHide', value) instead of having 'GuiTreeview', 'ShowHide' it would probably be clearer

:+1:

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -152,3 +162,45 @@ function GuiClipboard()
...
+
+
+" Notifies the TreeView widget of a Show or Hide event
+function! s:TreeViewShowHide(show)
+	  " Use the last UI in the list
+	  let ui_chan = s:get_last_ui_chan()
+    if (ui_chan == -1)
+      return
+    endif
+    call rpcnotify(ui_chan, 'Gui', 'TreeView', 'ShowHide', a:show)
```
equalsraf: Go back to using 0 here for the ui_chan - its ok since its a notification

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -152,3 +162,45 @@ function GuiClipboard()
...
+anoremenu <script> Gui.Treeview.Hide :call <SID>TreeViewShowHide(0)
+
+" Notifies the TreeView widget of a Toggle event
+function! s:TreeViewToggle()
+	  " Use the last UI in the list
+	  let ui_chan = s:get_last_ui_chan()
+    if (ui_chan == -1)
+      return
+    endif
+    call rpcnotify(ui_chan, 'Gui', 'TreeView', 'Toggle')
```
equalsraf: Go back to using 0 here for the ui_chan - its ok since its a notification

equalsraf:
Sorry I was not clear earlier. It is perfectly fine to use the chanel 0 for rpcnotify. In fact it may be preferable in this case.

I dont think there is a strong reason to restrict it (unless you are VERY concerned about performance/bandwidth), Can you revert that change? After that we can merge.

gagbo:
I wasn't sure about using broadcasting or not because usually I only have one instance of `nvim-qt` running, so I never had to think about which instance get the signal or not.

I reverted the change, and if someone complains about it someday then I guess we can go back to looking at how this is done.

equalsraf:
I actually "use" it. I dont keep multiple GUIs but I connect/disconnect the GUI often.

PR #485, opened 2018-12-21
[RFC] Clang-format : add/apply the results of whatstyle

gagbo:
This is basically me trying to guess the coding style of the project. I noticed a few inconsistencies here and there (there are indented `case` within `switch` statements and non-indented ones, same with the `&` and `*` placement, I don't know if there's a rule ?)

I tried to find out the least-changing style from clang-format and apply it to the codebase. I don't know if it's fine or not.

[Whatstyle](https://github.com/mikr/whatstyle) command :
```
./whatstyle.py -f clang-format ../../neovim-qt/src/*.cpp
../../neovim-qt/src/*.h ../../neovim-qt/src/gui/*.h
../../neovim-qt/src/gui/*.cpp ../../neovim-qt/src/gui/shellwidget/**.h
../../neovim-qt/src/gui/shellwidget/**.cpp --mode resilient
```

I tweaked a little the result to try to fix a few inconsistencies, and
also put a "no op" .clang-format file in the bindings/ folder so it is
left untouched


equalsraf:
You can skip the files in src/auto for now since they are auto generated.

src/gui/app.cpp
```diff
@@ -122,44 +121,43 @@ void App::showUi(NeovimConnector *c, const QCommandLineParser& parser)
...
-	parser.addOption(QCommandLineOption("fullscreen",
-				QCoreApplication::translate("main", "Open the window in fullscreen on startup")));
-	parser.addOption(QCommandLineOption("embed",
-				QCoreApplication::translate("main", "Communicate with Neovim over stdin/out")));
-	parser.addOption(QCommandLineOption("server",
-				QCoreApplication::translate("main", "Connect to existing Neovim instance"),
-				QCoreApplication::translate("main", "addr")));
-	parser.addOption(QCommandLineOption("spawn",
-				QCoreApplication::translate("main", "Treat positional arguments as the nvim argv")));
+										QCoreApplication::translate("main", "nvim executable path"),
```
equalsraf: This is one case where the format makes things really hard to read

gagbo:
I think most of the weirdness is also applied because the code uses tabs most of the time, but also has 4 char tabs and 8 char tabs in some places.

The `ForContinuationAndIndentation` setting was responsible for the very bad lines you pointed out, now the Tabs are only used for actual indentation (so there are no tabs added to try to over-indent continuated lines), it helps with these parser calls.

src/gui/input.cpp
```diff
@@ -108,50 +109,56 @@ QString InputConv::modPrefix(Qt::KeyboardModifiers mod)
  *
  * If the event is not valid for Neovim, returns an empty string
  */
-QString InputConv::convertMouse(Qt::MouseButton bt, QEvent::Type type, Qt::KeyboardModifiers mod, QPoint pos, short clickCount)
+QString InputConv::convertMouse(Qt::MouseButton bt, QEvent::Type type, Qt::KeyboardModifiers mod,
+                                QPoint pos, short clickCount)
 {
 	QString buttonName;
-	switch(bt) {
```
equalsraf: I guess this would be `IndentCaseLabels: false`

third-party/konsole_wcwidth.cpp
```diff
@@ -70,29 +70,28 @@
 #include "konsole_wcwidth.h"
```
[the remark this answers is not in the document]
gagbo: yeah I wasn't sure about this folder either :)

equalsraf:
>  I noticed a few inconsistencies here and there (there are indented case within switch statements and non-indented ones, same with the & and * placement, I don't know if there's a rule ?)

No particular rule. Most of the codebase is formatted based on my settings, with some exceptions that crept into the code as more people commited new stuff. For your specific points

- I dont indent case statements
- usually I place &/* on the right, but this has changed over the years (a consequence of writing rust) I have no special preference on this - I assume most of the code puts then on the right hand side.

src/gui/shellwidget/shellcontents.cpp
```diff
@@ -22,17 +23,16 @@ bool ShellContents::fromFile(const QString& path)
...
+		resize(_rows + 1, qMax(_columns, string_width(line)));
 		put(line, row, 0);
 		row += 1;
 	}
 	return true;
 }
 
-ShellContents::ShellContents(int rows, int columns)
-:_data(0), _rows(rows), _columns(columns)
+ShellContents::ShellContents(int rows, int columns): _data(0), _rows(rows), _columns(columns)
```
equalsraf: In general I call initializers in a separate line - but I cant find the correct option for this.

src/gui/shellwidget/shellcontents.cpp
```diff
@@ -58,28 +58,27 @@ ShellContents::ShellContents(const ShellContents& other)
...
-	_data = new Cell[_rows*_columns];
+	_data = new Cell[_rows * _columns];
 }
 
 void ShellContents::clearAll(QColor bg)
 {
-	for (int i=0; i<_rows; i++) {
-		for (int j=0; j<_columns; j++) {
-			_data[i*_columns+j] = Cell::bg(bg);
+	for (int i = 0; i < _rows; i++) {
```
equalsraf: Cannot find an option for clang-format to disable space around operators.

src/gui/shellwidget/shellcontents.cpp
gagbo: It's [Break Constructor Initializers](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) on this page, I don't know which one you prefer. There's also the same with Inheritance List.

ConstructorInitializerAllOnOneLineOrOnePerLine is relevant too, and I'm checking if there's a way to always break even if the line is short enough, but I'm not sure yet I'll find it

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -4,60 +4,62 @@
 #include "shellwidget.h"
 #include "helpers.h"
 
-ShellWidget::ShellWidget(QWidget *parent)
-:QWidget(parent), m_contents(0,0), m_bgColor(Qt::white),
-	m_fgColor(Qt::black), m_spColor(QColor()), m_lineSpace(0)
+ShellWidget::ShellWidget(QWidget* parent)
+: QWidget(parent), m_contents(0, 0), m_bgColor(Qt::white), m_fgColor(Qt::black),
```
equalsraf: SpaceBeforeCtorInitializerColon: false

src/gui/shellwidget/shellcontents.cpp
gagbo: I;ll test SpaceBeforeAssignmentOperators to false

src/gui/shellwidget/shellwidget.cpp
gagbo: I had this one already, I think that no formatting option yet allows no space there, I'll try to look further

src/gui/app.h
```diff
@@ -10,13 +10,12 @@
 namespace NeovimQt {
 
 class NeovimConnector;
-class App: public QApplication
-{
+class App: public QApplication {
 	Q_OBJECT
 public:
-	App(int &argc, char ** argv);
```
equalsraf: PointerAlignment: PAS_Right

src/gui/shellwidget/shellcontents.cpp
gagbo: Another solution for this would be to set `ColumnLimit` to 0, so the formatter never reformats long lines, but it always respect manual breaks

src/gui/main.cpp
```diff
@@ -33,8 +33,8 @@ int ui_main(int argc, char **argv)
 	QCommandLineParser parser;
 	NeovimQt::App::processCliOptions(parser, app.arguments());
 
-	int timeout = parser.value("timeout").toInt();
-	auto c = app.createConnector(parser);
+	int timeout= parser.value("timeout").toInt();
```
equalsraf: SpaceBeforeAssignmentOperators: true

equalsraf: I dont mind if it ends up changing the style a bit. What I would really like to avoid right now is having to review the entire code base just because of the format :)

gagbo: The goal is definitely to have only `clang-format` induced changes here, and be able to review code styles throughout the code

gagbo:
I definitely won't come back to this anytime soon. The discussion is pretty interesting though !

PR #486, opened 2018-12-22
Tests : Change the -u NORC invocations to -u NONE

gagbo:
This ensures that $HOME/.config/nvim/ginit.vim does not get picked up
when running the tests.

This caused issues when I tried to run tests locally, as my ginit.vim
tries to call an autoloaded function (which was not found during ctests)

I thought that replacing `NORC` with `NONE` would break the tests reading and using commands in `src/runtime/plugin/nvim_gui_shim.vim`, but I think that the `--cmd "set rtp+=...` flag works well because the tests still pass locally.

Here's to hoping they'll pass on the CI :beers:

### DIFF
--- a/test/tst_callallmethods.cpp
+++ b/test/tst_callallmethods.cpp
@@ -30,7 +30,7 @@ private:
 
 void TestCallAllMethods::initTestCase()
 {
-	m_c = NeovimQt::NeovimConnector::spawn({"-u", "NORC"});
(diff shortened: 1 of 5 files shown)


gagbo:
This works because the shim gets runtimed during Shell::setAttached() [here](https://github.com/equalsraf/neovim-qt/blob/4657c52d0bbf2f03eeeb83364004e8592c9bae70/src/gui/shell.cpp#L152), and the corresponding tests actually append the `src/runtime` to rtp in the invocation command.

Issue #487, opened 2018-12-23
Preview window positioning and colors are all weird 


s-kashif:
Completion window is not using correct colors on Preview (Using YoucompleteMe)
Also the position is offset

Occurs in 0.2.10 and 0.2.11, Doesn't happen in 0.2.8

![image](https://user-images.githubusercontent.com/9801383/50380524-2c120c80-06ad-11e9-81fc-786395724a87.png)

s-kashif:
But this wasn't happening in v0.2.8  (Refer screenshot of v 0.2.8). 0.2.8 was using correct widget and was respecting the color scheme
Seems like a regression in later versions
![image](https://user-images.githubusercontent.com/9801383/50390351-b0ca5c80-0779-11e9-8cd7-688f8789e671.png)

gagbo:
You can get this behaviour (let Nvim decide the popup stuff) with `:GuiPopupmenu 0`. Adding this in your `ginit.vim` should give you this behaviour.

In the following screenshots, I have 
- Qt theme : Kvantum KvGlass + ArcDark ; 
- Neovim theme : Snow

So anything light is from Neovim, anything dark is a Qt widget

With GuiPopupmenu (the default)  :

![screenshot_guipopupmenu](https://user-images.githubusercontent.com/10496163/50397746-def86e00-0772-11e9-8d40-0b5b5378793d.png)
 

After `:GuiPopupmenu 0` : 

![screenshot_noguipopupmenu](https://user-images.githubusercontent.com/10496163/50397787-f0da1100-0772-11e9-928d-786f5f7e8e36.png)

Note that the position is not the same (above or below text here), but I still have something reasonable (this is done while working on a feature branch that only adds a directory tree view, nothing at all about the Popupmenu).

I'll admit that we can work a little bit on alignment of columns and stuff like that, but the positioning and the theming seem fine to me

s-kashif:
@equalsraf  and @gagbo Thanks a lot. I'll check

Issue #471, opened 2018-10-02
complete popup have wrong color


trlsmax:
Arch linux x64, neovim 0.3.1, neovim-qt compiled with the latest code. No matter what complete plugin I used, popup window have wrong color.
![screenshot_2018-10-02_21-11-55](https://user-images.githubusercontent.com/5501321/46350642-d7e82500-c687-11e8-9827-5dfdc332aadd.png)

equalsraf:
The popup menu is provided by the GUI and it should be using the default Qt style for your system.
To disable the GUI popup you can use `:GuiPopupmenu 0` (you can also use this in ginit.vim).

Another strange issue with that menu is the line spacing between entries. Are you having similar issues in other Qt applications?

trlsmax:
neovim said GuiPopupmenu is not an editor command and use it in ginit.vim still not work.

My other Qt applications are seems just fine.

trlsmax:
Problem solved with disable plugin equalsraf/neovim-gui-shim and use :GuiPopupmenu 0 in ginit.vim
Thank you.

PR #492, opened 2018-12-31
[WIP] externalize cmdline

gagbo:
This is also a RFC since I'm still pretty unsure about the way to go with this.

## Summary
- [x] cmdline_show
    - [x] "functionnal"
    - [x] complete
- [x] cmdline_special_char
    - [x] "functionnal"
    - [x] complete
- [x] cmdline_hide
    - [x] "functionnal"
    - [x] complete
- [x] cmdline_block_show
    - [x] "functionnal"
    - [x] complete
- [x] cmdline_block_append
    - [x] "functionnal"
    - [x] complete
- [x] cmdline_block_hide
    - [x] "functionnal"
    - [x] complete

## Result
We want `ext_cmdline` to be fully respected

## How
The externalized command line is currently implemented as a `CmdWidget`, a list of which is owned by the main `ShellWidget`. There is one `CmdWidget` per `level` of cmdline as described in `:h ui-cmdline`

### cmdline_show
- The content is shown, but only as plain text
- Levels are handled by assuming cmdlines are only one-liners and not block (which is obviously going to be a problem later). See how `anchor_y` is computed in `Shell::handleCmdlineShow()`
- Tests should be added to test the correct handling of `firstc`, `prompt` and how `pos` is affected
- I can't find how to make the cursor visible yet, which is pretty sad.

### cmdline_special_char
- This works as expected, but it probably needs to get a special Rich Text treatment for clarity

### cmdline_hide
- Currently hides all cmdlines found in the Cmdline list member of `Shell`.
- Since there is no `level` argument yet, this is probably the good behaviour.

### cmdline_block_show
Unimplemented.

I think using the Paragraphs/Blocks features of the underlying QTextDocument should be really helpful there. Basically this would mean using [QTextEdit::append(QString)](http://doc.qt.io/qt-5/qtextedit.html#append) method to handle each line of the cmdline_block

### cmdline_block_append
Unimplemented.

This one is clearly for [QTextEdit::append(QString)](http://doc.qt.io/qt-5/qtextedit.html#append), at least in this starting implementation of CmdWidget

### cmdline_block_hide
Unimplemented.

This will probably just be the same `QWidget::hide()` function as in `cmdline_hide`

### DIFF
--- a/src/gui/cmdline.cpp
+++ b/src/gui/cmdline.cpp
@@ -1,0 +1,41 @@
+#include "cmdline.h"
+
+namespace NeovimQt {
+
+CmdLine::CmdLine(QWidget *parent) : QTextEdit(parent) {
+    hide();
+    cursor = QTextCursor(document());
+    setTextCursor(cursor);
+    document()->setDefaultFont(parent->font());
+}
+
+void CmdLine::compute_document(const QString& firstc, const QString& prompt, const QVariantList& content) {
+    QString plaintext_content;
+    foreach(QVariant piece, content){
+        qDebug() << piece.toList();
+        plaintext_content += piece.toStringList().at(1);
+    }
+    qDebug() << prompt << firstc;
+    setPlainText(firstc + prompt + plaintext_content);
+    cursor = QTextCursor(document());
+}
+
+void CmdLine::add_special_char(const QString &c, bool shift_c) {
+    if (!shift_c) {
+        setOverwriteMode(true);
+    }
+    cursor.insertText(c);
+    setOverwriteMode(false);
+}
+
+void CmdLine::setPos(int64_t pos) {
+    cursor.movePosition(QTextCursor::StartOfLine);
+    cursor.movePosition(QTextCursor::Right, QTextCursor::MoveAnchor, pos);
+}
+
+void CmdLine::keyPressEvent(QKeyEvent *ev)
+{
+    QWidget::keyPressEvent(ev);
+}
+
+}  // namespace NeovimQt
--- a/src/gui/cmdline.cpp
+++ b/src/gui/cmdline.cpp
@@ -4,113 +4,115 @@
 namespace NeovimQt {
 
 CmdLine::CmdLine(QWidget *parent) : QTextEdit(parent) {
-    hide();
+        hide();
 
-    setFrameShape(QFrame::NoFrame);
-    setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
+        setFrameShape(QFrame::NoFrame);
+        setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
 
-    document()->setDefaultFont(parent->font());
+        document()->setDefaultFont(parent->font());
 
-    auto cmdwidget_parent = dynamic_cast<CmdWidget*>(parent);
-    setTextBackgroundColor(cmdwidget_parent->shell()->background());
-    setTextColor(cmdwidget_parent->shell()->foreground());
+        auto cmdwidget_parent = dynamic_cast<CmdWidget*>(parent);
+        setTextBackgroundColor(cmdwidget_parent->shell()->background());
+        setTextColor(cmdwidget_parent->shell()->foreground());
 }
 
 void CmdLine::compute_document(const QString& firstc, const QString& prompt, const QVariantList& content) {
-    QString plaintext_content;
-    QString html_content;
-    clear();
-    style_cursor_CharFormat(QVariantMap());
-    textCursor().insertText(firstc + prompt);
-    foreach(QVariant piece, content){
-        qDebug() << piece.toList();
-        style_cursor_CharFormat(piece.toList().at(0).toMap());
-        plaintext_content = piece.toStringList().at(1);
-        textCursor().insertText(plaintext_content, textCursor().charFormat());
-    }
-    setFocus();
+        QString plaintext_content;
+        QString html_content;
+        clear();
+        style_cursor_CharFormat(QVariantMap());
+        textCursor().insertText(firstc + prompt);
+        foreach(QVariant piece, content){
+                qDebug() << piece.toList();
+                style_cursor_CharFormat(piece.toList().at(0).toMap());
+                plaintext_content = piece.toStringList().at(1);
+                textCursor().insertText(plaintext_content, textCursor().charFormat());
+        }
+        setFocus();
 }
 
-QColor CmdLine::color(qint64 color, const QColor& fallback) const
-{
-    if (color == -1) {
-        return fallback;
-    }
-    return QRgb(color);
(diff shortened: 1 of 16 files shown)


gagbo:
Next thing I'll do is probably to add tests properly so everything fails TDD-style

src/gui/cmdwidget.cpp
```diff
@@ -0,0 +1,61 @@
+#include "cmdwidget.h"
+
+namespace NeovimQt {
+
+CmdWidget::CmdWidget(ShellWidget *parent) {
+    setParent(parent);
```
equalsraf: The setparent should not be needed if you call the parent constructor, just `:QTextEdit(parent)`, or am I missing something.

src/gui/cmdwidget.cpp
```diff
@@ -0,0 +1,61 @@
...
+    hide();
```
equalsraf: This can probably be done in the parent widget instead of here.

src/gui/cmdwidget.cpp
gagbo: My bad, I definitely can just call QTextEdit(parent)

src/gui/cmdwidget.cpp
gagbo: I thought about that at first, but I wanted to put hide() in the constructor to avoid "boilerplate" in the parent widget constructor. I think it's safe to assume that all new widgets should just be hidden, and then just wait for the cmdline_show and cmdline_block_show calls

src/gui/shell.h
```diff
@@ -181,6 +189,8 @@ private slots:
 	ShellOptions m_options;
 	PopupMenu m_pum;
 	bool m_mouseEnabled;
+    std::vector<CmdWidget*> m_cmdline_list;
```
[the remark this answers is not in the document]
gagbo: Didn't read the doc about that, but this seems like what I had in mind.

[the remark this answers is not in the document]
gagbo: I think a QGridLayout would be better there. I like the idea of having the context of the command when spawning "inner" command lines like this (in the screenshot the cursor is missing because I need to fix that, and the `"` is the `cmdline_special_char` which is not styled yet) : 

screenshot_20190101_230013

src/gui/cmdwidget.cpp
equalsraf: I usually prefer this kind of things to be explicit, otherwise if someone else tries to use it, they spend some time scratching their head on why it is invisible.

equalsraf:
I am seeing a scroll bar on the right, I assume its the QTextEdit. Probably something like `setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff)`

gagbo:
Next commit is going to be pretty big, I made a few changes in design which should help separate the components better, I'll add the VerticalScrollBarPolicy then

gagbo:
Appveyor 404'd on me :(

On other things : 
- There is a ton of code duplication in `html_style_attribute` in cmdblock and cmdline. I don't know where the best place for this would be. It also borrows a lot of code from `Shell::handleHighlightSet()` (a lot meaning "it's the same code except it works with local variables instead of class attributes)
- The change in Palette done in 00ad4e7 may be a solution to get a consistent theming between Neovim colorscheme and the Qtwidget ( cf #487 #471 )
- With this implementation, it is impossible to follow properly `special` highlight. QTextEdit displays rich text using html, and the styling does not allow to have a different color for underline/undercurl. I don't think it's too critical for cmdline, but the limitation is still here.

gagbo:
[This regression](https://bugreports.qt.io/browse/QTBUG-59093?jql=text%20~%20"QPalette%20cursor") is probably blocking the possibility of having a visible cursor in the cmdline for the time being

src/gui/cmdline.cpp
```diff
@@ -0,0 +1,118 @@
...
+        if (color == -1) {
+                return fallback;
+        }
+        return QRgb(color);
+}
+
+void CmdLine::style_cursor_CharFormat(const QVariantMap& attrs) {
+        QTextCharFormat cur_format;
+
+        auto cmdwidget_parent = dynamic_cast<CmdWidget*>(parent());
```
equalsraf: Since we are dependent on the widget anyway - it is ok to store the proper pointer type in a member of this class and use it here, avoiding the dynamic_cast.

equalsraf: or the shell directly

gagbo: No problem. I'll use the CmdWidget for this, it seems more logical to me.

For the time being `CmdWidget::shell()` exists, but maybe at some point we could want to use separate styling (basically separate `QPalette` I guess) for Shell and CmdWidget

gagbo:
Now the `CmbBlock` widget moves to stay close to the `CmdWidget` holding the command lines. The placement is still not ideal I guess (since it covers the first line of the buffer), but I am bad at designing UI, so I'll wait on more input. I mostly did these changes to force myself to add utilities to move the widgets within a `Shell`.

## Resizing main window issue

I still have one weird issue : when I resize the main window, **and** the `CmdBlock` widget is visible, then only `CmdWidget` and `CmdBlock` are redrawn during the resizing. And everything stays blank meanwhile (Don't mind how the cmd widgets resize/move during the gif, it has been fixed)

cmdline_block_resize

## Better UI ?

I think the best solution to be sure that these widgets won't be a nuisance to buffer visibility would be either : 
- to actually create another window just for the sake of the commands. This way the use handles their workspace as they like. But this seems a little overkill
- or to make these widgets draggable within the parent `Shell` widget. That's the solution I like the most, but a quick SO search says that it is a lot of work.

gagbo:
Adding Drag and Drop support (at least dropping text in `CmdWidget`) is also on the list of features to add.

src/gui/cmdwidget.cpp
gagbo: Ok, I moved the calls in `Shell` construction like for `Shell::m_pum` 
f568e67

equalsraf:
Concerning the resize issue, it would seem that, either

1. neovim is not handling the resize event sent by the UI
2. We are not processing the resize event

The strange part of that issue is that it only happens when the CmdBlock is visible, but it does not happen when it is not.

Even more interesting if you do the same thing without this patch you also see a similar effect. So maybe this is something that needs some digging upstream.

gagbo:
This is very stale too. I don't think I'll work on that soon, and it might be better to start this work from scratch for anyone interested. I didn't keep up with neovim and I don't know if any protocol feature is actually superseding cmdline now

gagbo:
Probably superseded by #586 as I haven't touch this branch in a while and don't expect to anytime soon :(

jgehrig:
Superseded by Pull #586. This was a very useful reference :)

Marking as closed.

Issue #445, opened 2018-08-14
Tracking issue for unsupported neovim features


equalsraf:
- [ ] default_colors_set
- [ ] mode_info_set
- [ ] arabicshape
- [ ] ambiwdith
- [ ] emoji
- [ ] ext_cmdline
- [ ] ext_cmdline
- [ ] ext_wildmenu
- [ ] ext_linegrid

equalsraf:
Yeah we just had to set termguicolors the old days (we still set it actually). We still get the notification via **option_set**, should we do something with it.

bfredl:
No, I meant `option_set('termguicolors', true/false)` is specifically for a remote TUI that wants to to support both cterm and rgb colors the same way as the builtin TUI does, a GUI has nothing to do with it.

> Yeah we just had to set termguicolors the old days (we still set it actually). 

This is a hack for some color schemes that check it, I presume?

equalsraf:
> This is a hack for some color schemes that check it, I presume?

No. I just like to be able to run nvim 0.1.4 some times :D and I never removed the code that added that argument.

bfredl:
Did nvim 0.1.4 have a bug that made it ignore `ui_attach(..., rgb=true)` if `termguicolors` was not set? `ui_attach` should be older than `termguicolors`...

equalsraf:
I dont remember exactly but back them the GUI had to set the option for colorschemes to work correctly.

bfredl:
That was exactly what I said before with "a hack for some color schemes that check it". Buggy color schemes have to be fixed (and some have over the time I presume), this is independent of the nvim version.

jgehrig:
This issue has become stale. Marking as closed.

Several features have been implemented:
- ext_linegrid
- default_colors_set
- mode_info_set

One feature is still in progress:
- ext_cmdline

One feature has been superseded:
- ext_wildmenu

PR #494, opened 2019-01-05
default_colors_set : basic implementation

gagbo:
I don't know if it's better to add a few steps like this at a time, or just make a very big issue/PR about `ext_linegrid`. In any case, there's a very basic implementation of default_colors_set using the current `Shell` API.

I can't test it yet against 0.3.2 or master, so I won't try more than this for the moment.


gagbo:
Related issue #445

jgehrig:
Support for `default_colors_set` was added with `ext_linegrid`

Marking as closed.

PR #495, opened 2019-01-06
Fix issue that popup menu was not displayed

erw7:
Fix a problem that the popup menu did not display because the argument of
`popupmenu_show` was changed by https://github.com/neovim/neovim/commit/16c3337122955c1e18c5ff69dcb14b61c43c4ac0.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -502,7 +502,8 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
 			emit neovimSuspend();
 		}
 	} else if (name == "popupmenu_show") {
-		if (opargs.size() != 4
+		if (!(opargs.size() == 4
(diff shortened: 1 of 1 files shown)


src/gui/shell.cpp
```diff
@@ -502,7 +502,8 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
 			emit neovimSuspend();
 		}
 	} else if (name == "popupmenu_show") {
-		if (opargs.size() != 4
+		if (!(opargs.size() == 4
+          || (opargs.size() == 5 && opargs.at(4).canConvert<qint64>()))
```
equalsraf: Just change it to `size() < 4` and add a comment about the 5th argument, that way we wont run into more issues like it.

equalsraf:
Merged, many thanks @erw7.

PR #498, opened 2019-01-07
Continue implementation of GUI clipboard

equalsraf:


equalsraf:
Merging because its sunday and I want this on my monday. This still does not handle all cases but its an improvement.

PR #500, opened 2019-01-08
Fix #499

### LINKED ISSUE #499: is this a bug here? shouldn't we reset the end_row?
https://github.com/equalsraf/neovim-qt/blob/130baeea8f945cb4b99b81025e5974378aaa1453/src/gui/shellwidget/shellwidget.cpp#L117

equalsraf:
In the shellwidget the paint event rounds down the row/column to comply
to the actual shell dimensions. Turns out the statment that handles the
rows is actually readjusting the columns - this has two potential
consequences

1. we could be painting wrong data below the last row
2. we could be painting areas (columns) that do not need painting

I was unable to reproduce cases of 1. It should be noted that because
the widget always resizes the shell when a resize occurs, the difference
between the last row/column would always be fractional.

----

One instance of 2. would be a vertical resize that would cause other areas of the widget to be repainted without need. I cant think a concrete case though.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -114,7 +114,7 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 			end_col = m_contents.columns();
 		}
 		if (end_row > m_contents.rows()) {
-			end_col = m_contents.columns();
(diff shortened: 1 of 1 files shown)


equalsraf:
ref #499

Issue #502, opened 2019-01-09
Printing v:errmsg on attach is interfering with plugins that run at VimEnter


mhinz:
Hi,

assume this minimal vimrc:

```vim
autocmd VimEnter * silent! xxx
```

Usually, a manual `:silent! xxx` will simply output nothing, for obvious reasons. But during startup/VimEnter, using the vimrc from above, you'll get the message:

```
E492: Not an editor command: silent! xxx
```

It also gets logged to the message history: `:messages`

Bug? FWIW, this doesn't happen with nvim alone.

(For testing I used nvim-qt.app master and nvim master.)

mhinz:
Culprit: https://github.com/equalsraf/neovim-qt/commit/badc90ee32b5a973c3aa07ee3e4046da6bf09325

What's the intention behind that commit?

justinmk:
That commit was a workaround for https://github.com/equalsraf/neovim-qt/pull/463 , but it's no longer necessary with Nvim 0.3.2 [--embed behavior](https://github.com/neovim/neovim/wiki/Following-HEAD#20180922).

bfredl:
Half-related note: for any arbitrary string UTF-8 should be used, not `err.toLatin1()` as in that commit. Error messages can be translated... (the other cases in the codebase is ASCII generated strings, but one could consider changing them for the risk of being naively copied to a case where it matters).

equalsraf:
Reverted in master. Thanks @mhinz.

PR #503, opened 2019-01-09
Revert "GUI: On attach show v:errmsg"

equalsraf:
This reverts commit badc90ee32b5a973c3aa07ee3e4046da6bf09325. Because
the printing of v:errmsg was interfering with plugins (see #502).


justinmk:
@equalsraf When you have a chance it would be great to have this in a release, then we will include it with Nvim 0.3.4 maintenance release.

Issue #513, opened 2019-02-07
Cursor shape change does not work when statusline option is set


Shougo:
I have tested below minimal vimrc.

```vim
set noshowcmd
set noshowmode
set statusline=foo
```

If you test the vimrc, cursor shape does not change in insert mode.

equalsraf:
Confirmed here, it seems like we are failing to repaint the cursor.

For example changing the focus of the window does cause the cursor to be repainted.

#514 should fix this

equalsraf:
Merged, thanks @Shougo

PR #514, opened 2019-02-08
Repaint cursor after mode change

equalsraf:
When the mode was changed the cursor was not being repainted.

ref #513


Shougo:
I have confirmed the problem is fixed.

Issue #515, opened 2019-02-13
command line is hided by windows 10 task bar


ShuguangSun:
- `nvim --version`: 0.3.4 and 0.4.0
- Operating system/version: Windows 10

### Steps to reproduce 

run nvim-qt.exe and then maximize the window
or nvim-qt --maximized
or put `call GuiWindowMaximized(1)` in ginit.vim

### Actual behaviour
The nvim-qt is really maximized but it fit the whole screen (some kind like of fullscreen). And the windows taskbar (with start menu and those shortcut icons) hide the status/command line. I need to `set cmdheight=2` to make the area above windows task bar.

The fontsize and dpi has some impact.

### Expected behaviour
I expect the bottom of nvim-qt windows fit windows taskbar.

sprainbrains:
I guess, it's the same my [problem](https://github.com/equalsraf/neovim-qt/issues/511)

sprainbrains:
I know how to fix it. I will pull request later.

justinmk:
https://github.com/equalsraf/neovim-qt/pull/516

Issue #511, opened 2019-02-04
Incorrect footer size in full screen mode on Windows 10.


sprainbrains:
When i expand the window to full screen so then  part of the footer has disapeared(

![image](https://user-images.githubusercontent.com/10284309/52203653-dd968700-2882-11e9-9cca-da68337c7ec7.png)

jjohnson338:
I found that hiding the tab-bar (in ginit.vim) got the UI to line up correctly:
`GuiTabline 0`

sprainbrains:
> I found that hiding the tab-bar (in ginit.vim) got the UI to line up correctly:
> `GuiTabline 0`

Yep, I knew about it. But i want to solve this trouble with a gui tab)

justinmk:
closed by https://github.com/equalsraf/neovim-qt/pull/516

PR #516, opened 2019-02-13
Fixed for issues #511 and #515

sprainbrains:
I fixed  a height for command line interface in a fullscreen mode for Windows 10. (look  issues: [1](https://github.com/equalsraf/neovim-qt/issues/511) and [2](https://github.com/equalsraf/neovim-qt/issues/515)).

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -4,342 +4,348 @@
 #include <QToolBar>
 #include <QLayout>
 
-namespace NeovimQt {
+namespace NeovimQt
+{
 
 MainWindow::MainWindow(NeovimConnector *c, ShellOptions opts, QWidget *parent)
-:QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0),
-	m_delayedShow(DelayedShow::Disabled), m_tabline(0), m_tabline_bar(0),
-	m_shell_options(opts), m_neovim_requested_close(false)
+    : QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0),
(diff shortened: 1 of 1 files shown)


src/gui/mainwindow.cpp
```diff
@@ -4,342 +4,348 @@
...
-		m_shell->resizeNeovim(m_shell->size());
-	}
+    // Neovim finished resizing, resize it back to the actual
+    // widget size - this avoids situations when neovim wants a size that
+    // exceeds the available widget size i.e. the GUI tells neovim its
+    // size, not the other way around.
+    if (isMaximized() || isFullScreen()) {
+        QSize size = geometry().size();
+        if (m_tabline_bar->isVisible()) {
+            size.scale(size.width(),
```
equalsraf: The scale here is unnecessary right? You can just `size.setHeight(size.height() - m_tabline_bar->geometry().size().height())`

equalsraf:
Hi @sprainbrains. thanks for the fix.

Can you keep the style formatting changes out, the make the diffs rather hard to read.

sprainbrains:
> Hi @sprainbrains. thanks for the fix.
> 
> Can you keep the style formatting changes out, the make the diffs rather hard to read.

Sorry for that. I forgot to disable autoformatter in may editor.

src/gui/mainwindow.cpp
sprainbrains: > The scale here is unnecessary right? You can just `size.setHeight(size.height() - m_tabline_bar->geometry().size().height())`

You right. Enough to add this code to mainwindow.cpp : 

`if (m_tabline_bar->isVisible()) {
            size.setHeight(size.height() - m_tabline_bar->geometry().size().height());
 }`

Issue #478, opened 2018-12-03
Change GuiTabline interface style


larrytang1:
Is there any way to change the GuiTabline style ? In current Neovim implementation it looks like Windows 98/2000. Can we change it looks like a modern Operation System such as Windows 10 with flat UI.

equalsraf:
@davidgaleano these days I'm not much of a windows user but all the usual methods that Qt5 uses for interface styling should be available. Usually Qt should apply the native style, but maybe qtconfig can change the default.

Can you post a screenshot of what it looks like?

In addition you can also use a qss stylesheet through the command line with `nvim-qt --stylesheet file.qss`

davidgaleano:
@equalsraf I guess you meant to answer to @daviestang?

daviestang:
![nvim](https://user-images.githubusercontent.com/10758075/49447107-a79c3e80-f810-11e8-93b5-7c7e05adf673.png)

daviestang:
![gvim](https://user-images.githubusercontent.com/10758075/49447653-0dd59100-f812-11e8-88d5-93cd521cf482.png)

In gvim it's more like system UI styling. Is there any way to change nvim-qt gui tabline like this ?

equalsraf:
@davidgaleano sorry about that - I blame it on auto-complete :)

@daviestang I think this is not normal (but I cant test this right now). Where did you install nvim-qt from? Which windows version are you using?

kdurant:
Current GuiTabline are very ugly and difficult to use ！

daviestang:
@equalsraf I installed the nvim-qt from Neovim pre-release package [NVIM 0.3.2-dev](https://github.com/neovim/neovim/releases), and using Windows 10 pro version.

daviestang:
@equalsraf Hello, I checked in neovim-qt project latest release v0.2.11 today. This issue has been fixed. Many thanks!

equalsraf:
errr ... I did nothing to fix this. I dont know the specifics in windows, but the Qt style can usually be changed - maybe something else was going on.

teto:
That's the same style as notepad++. I don't think gvim UI is more system-like than notepad. Just tweak the Qt style.

jgehrig:
It is now possible to override the styling with `:GuiAdaptiveStyle`. 

You can see all available styles on your system with `:GuiAdaptiveStyleList`.

[...]

Issue #496, opened 2019-01-06
Proposal : use QPalette to theme Shell's child widgets


gagbo:
Possible relatives : #489 #478 

## Summary
I think that using [QPalette](http://doc.qt.io/qt-5/qpalette.html#ColorRole-enum), and updating its values for each `colorscheme` event would help a lot with Widget consistency in the whole application.

## Illustration
For a basic example, I added a QPalette update in `update_fg` and `update_bg` handlers in 00ad4e7 in order to inherit directly the correct background/foreground for `cmdline` from `Shell::QPalette`.

### When would QPalette update
I don't think we need to change the palette for more events than these 2 (See all the documented color roles in the link above). Which means that eventually QPalette changes would only happens in `update_fg`, `update_bg` and `default_colors_set` I guess.

### What in QPalette would update
I am currently (still in 00ad4e7 ) using only `QPalette::Base` and `QPalette::Text` because it worked well to obtain the correct background in the cmdline widgets I'm implementing. But `QPalette::Tooltip{Base,Text}` can also be modified, and probably `Window` and `WindowText` too if we want to use other widgets than `Shell` like `TreeView`

### How QPalette would update
Also, I think it would be better to group all these methods/calls somewhere, this way we will be able to let the user decide if they want to have Neovim's colorscheme change the QPalette, or if they want to use the QPalette imported from their Qt settings (meaning leaving the QPalette untouched).

Having a single point of entry for Palette modification would also allow to trigger a massive repaint() to make the changes appear (for example to change an already created `TreeView` after having called `:colorscheme`)

## Misc

In any case, it would be nice to be able to render a (Neo)vim colorscheme in a QPalette, so we can see how applying the colorscheme would render for other widgets (I think the 'Application Color Theme' menu in my Plasma system settings app is basically a QPalette viewer)

![screenshot_20190106_120159](https://user-images.githubusercontent.com/10496163/50735196-115e8e00-11ab-11e9-812f-fb5975e2d65b.png)

I think that having this kind of tool to start things would be a nice way to get other opinions on the matter. The code for this section can be found [here](https://cgit.kde.org/plasma-desktop.git/tree/kcms/colors), but it uses `.ui` files for the layout, so translating this in pure cpp might take some time.

jgehrig:
The idea is good. Tracking with Pull #719.

Marking as closed.

PR #517, opened 2019-02-18
The implementation of adaptive colorscheme and font for GUI.

sprainbrains:
I added an automatic  color scheme adaptation for the neovim-qt GUI based on nvim colorscheme.

This implementation uses QPalette to create and set colors in the GUI by parsing a "default_colors_set" parameter.

So I added 2 commands to activate the GUI adaptation (GuiColorsAdaptive) and font adaptation (GuiFontAdaptive) to a nvim_gui_shim.vim file.

How it looks:
(Mac OS Ui behavior )
2019-02-18 22 42 25 mov

(Windows 10, gruvbox colorscheme, GuiColorsAdaptive enabled,GuiFontAdaptive enabled )

I guess, this feature may be related to the following issues: #494 #478 #489

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -124,27 +124,28 @@ void MainWindow::updateStyle()
 
     m_palette.setColor(QPalette::Window, bg_color);
     m_palette.setColor(QPalette::WindowText, fg_color);
-    m_palette.setColor(QPalette::Disabled, QPalette::WindowText, QColor(127, 127, 127));//Disable default color
+    m_palette.setColor(QPalette::Disabled, QPalette::WindowText, Qt::white);
 
     m_palette.setColor(QPalette::Base, bg_color);
     m_palette.setColor(QPalette::AlternateBase, fg_color);
     m_palette.setColor(QPalette::Text, fg_color);
-    m_palette.setColor(QPalette::Disabled, QPalette::Text, QColor(127, 127, 127)); //Disable default color
+    m_palette.setColor(QPalette::Disabled, QPalette::Text, Qt::white); 
 
 	m_palette.setColor(QPalette::Highlight, fg_color);
-	m_palette.setColor(QPalette::Disabled, QPalette::Highlight, QColor(80, 80, 80)); //Disable default color
+	m_palette.setColor(QPalette::Disabled, QPalette::Highlight, Qt::white);
 	m_palette.setColor(QPalette::HighlightedText, bg_color);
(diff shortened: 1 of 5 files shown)


gagbo:
Probably related : #496 

Looks cool !

src/gui/mainwindow.cpp
```diff
@@ -9,9 +9,13 @@ namespace NeovimQt {
...
 :QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0),
 	m_delayedShow(DelayedShow::Disabled), m_tabline(0), m_tabline_bar(0),
-	m_shell_options(opts), m_neovim_requested_close(false)
+	m_shell_options(opts), m_neovim_requested_close(false), 
+	m_neovim_gui_style_requested_update(false), m_neovim_gui_font_requested_update (false),
+	m_default_font(font()), m_default_palette(palette())
 {
 	m_errorWidget = new ErrorWidget();
+    m_style = QStyleFactory::create("Fusion");
+    setStyle(m_style);
```
gagbo: Maybe some combination of [QStyleFactory::keys()](https://doc.qt.io/qt-5/qstylefactory.html#keys) and preprocessor if statements can be used to try to get good default style on different OSes

sprainbrains: > Maybe some combination of [QStyleFactory::keys()](https://doc.qt.io/qt-5/qstylefactory.html#keys) and preprocessor if statements can be used to try to get good default style on different OSes

I tried, but always got a stange behavior in Windows OS.

src/gui/shell.cpp
```diff
@@ -523,7 +523,25 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
...
 		// TODO
-	} else if (name == "default_colors_set") {
+        } else if (name == "default_colors_set") {
+
+        	//"default_colors_set" have 5 parameters
+        	// 0 - foreground (default = -1)
+        	// 1 - background (default = -1)
+        	//...
+
+        	//don't handle default
```
gagbo: We can probably safely assume the last 2 parameters will never be used; but is it safe  to ignore the whole default_colors_set call if *either* fg or bg is `-1` ?

sprainbrains: > We can probably safely assume the last 2 parameters will never be used; but is it safe to ignore the whole default_colors_set call if _either_ fg or bg is `-1` ?

Yep. We can ignore the whole "default_colors_set" because our colorscheme using always 2 colors (for fg and bg). We cant set a new theme without 2 colors.

jgehrig:
This is a good idea, I like it. Moving the code over to #719.

Marking as closed.

PR #520, opened 2019-02-27
 Add support for guicursor

b-r-o-c-k:
This adds support for changing the cursor's color, blink rate, and shape by setting the `guicursor` option.


equalsraf:
Hi @b-r-o-c-k, thanks for this.

Do you know in which version the neovim api for the mode change was modified to include the index? We might have to retain part of the old code for handleModeChange.

b-r-o-c-k:
It looks like it was added in version 0.2 with this commit: neovim/neovim@9cdbbd49825561d642705990a2704b2241cf0584

YorickPeterse:
This would be nice to have, as especially the blinking of the cursor can get a little annoying. Is there anything holding this PR up (apart from the merge conflict)?

b-r-o-c-k:
The conflicts are resolved now and I believe it is ready to be merged.

equalsraf:
Trying this but my cursor is invisible, anyone else seeing it?

YorickPeterse:
I'm building this PR at the moment. In the mean time, I did see this warning that might be worth looking at:

```
/home/yorickpeterse/Projects/c++/neovim-qt/src/gui/shell.cpp: In lambda function:
/home/yorickpeterse/Projects/c++/neovim-qt/src/gui/shell.cpp:45:11: warning: enumeration value ‘Disabled’ not handled in switch [-Wswitch]
    switch (m_cursor_blink_state_) {
```

YorickPeterse:
It appears the cursor is also invisible for me by default. Using `set guicursor=a:block` does not appear to make a difference.

b-r-o-c-k:
There was a problem when I resolved the merge conflict so I rebased it now.
To change the cursor color you need to specify the highlight group name (e.g. `set guicursor=a:block-Cursor` to use the `Cursor` highlight group).

YorickPeterse:
I tested this and it's working now! :smile:

equalsraf:
Looking good on this end.

One question, should there be a default background/foreground color - in the terminal a guicursor is valid even if it does not specify a highlight group. Or perhaps we need to grab those from the Cursor group.

justinmk:
>  Or perhaps we need to grab those from the Cursor group.

Yes.

equalsraf:
Ok, in some cases the highlight id is 0, which is invalid according to `:h hlID()` and falls under the error case for that call (making the bg/fg invalid)

Tried to default to Cursor instead, but this does not seem right - I get a yellow cursor instead.(the colors seem to be reversed):

```
index 5de6076..55deb0f 100644
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -684,7 +684,12 @@ void Shell::handleModeChange(qint64 mode_index)
 
        const auto hl_id = cursor_style.value("hl_id");
        if (!hl_id.isNull() && hl_id.canConvert<qint64>()) {
-               const auto req_cursor_hl = m_nvim->api3()->nvim_get_hl_by_id(hl_id.toLongLong(), true);
+               MsgpackRequest* req_cursor_hl;
+               if (hl_id.toLongLong() == 0) {
+                       req_cursor_hl = m_nvim->api3()->nvim_get_hl_by_name("Cursor", true);
+               } else {
+                       req_cursor_hl = m_nvim->api3()->nvim_get_hl_by_id(hl_id.toLongLong(), true);
+               }
                connect(req_cursor_hl, &MsgpackRequest::finished, m_nvim,
                        [this](quint32 msgid, quint64 f, const QVariant& r) {
                                const auto foreground = r.toMap().value("foreground");
```

justinmk:
> (the colors seem to be reversed):

Is that because `:hi Cursor` default is:

    Cursor         xxx guifg=bg guibg=fg

teto:
shouldnt this be closed now that there is https://github.com/equalsraf/neovim-qt/pull/623

PR #521, opened 2019-03-01
Use hollow block cursor when focus lost

equalsraf:


equalsraf:
Naturally, I forgot about insert mode :D it looks a little funky. And it seems like it is drawing the outside of the cursor space, while this is not a big deal we should probably fix it.

PR #526, opened 2019-03-23
Add support for floating-point font sizes

### LINKED ISSUE #221: "Invalid font height" error on non-integer font heights
Running the following command in neovim-qt (on `master` from a few minutes ago) gives an "Invalid font height" error:
```vim
GuiFont Input Mono:h10.5
```

----

According to the `guifont` help in `options.txt`:
```
	- takes these options in the font name:
		hXX - height is XX (points, can be floating-point)
```

However, according to the help for the `GuiFont` command in `neovim_gui_shim.txt`:
```
		The following attributes are available

			hXX - height is XX in points
```
with no mention of whether or not it may be floating-point.

----

The reason I need floating-point sizes is because this is setting the font size with _points_, not _pixels_. In order to match what my terminal is set to (14 pixels high), I need something between 10 and 11 points, which I currently can't do.

YorickPeterse:
This adds support for setting GUI font sizes using floating-points. For
example, to use "Source Code Pro" at a size of 7.5 you'd run:

    GuiFont Source Code Pro:h7.5

This fixes #221.

## Examples

`Source Code Pro:h8:l`:

8pt

`Source Code Pro:h7.5:l`:

`Source Code Pro:h7:l` (ignore the window title, I forgot to change it):

### DIFF
--- a/src/gui/runtime/doc/nvim_gui_shim.txt
+++ b/src/gui/runtime/doc/nvim_gui_shim.txt
@@ -28,9 +28,13 @@ GuiFont		When called with no arguments this command display the
 
 		    :GuiFont Monaco:h13
 <
+		You can also use floating-point sizes:
+>
+		    :GuiFont Monaco:h13.5
+<
 		The following attributes are available
 
-			hXX - height is XX in points
+			hXX - height is XX in points (can be floating-point)
 			b   - bold weight
 			l   - light weight
 			i   - italic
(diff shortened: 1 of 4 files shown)


YorickPeterse:
@equalsraf What do you think of these changes? I tried to keep the amount of changed code to a bare minimum. I tested this locally and it works well.

equalsraf:
Looks good, merging. Thanks @YorickPeterse.

PR #543, opened 2019-05-27
Add support for 'ext_linegrid'

jgehrig:
This change adds support for the`ext_linegrid` UI protocol.

This change is motivated by a desire to add support form modal confirm dialogs in Neovim-Qt (close, save, open, etc). In the past, there wasn't a good way to do this. NeoVim does not emit events for `:confirm(...)`. However, `ext_messages` was added recently to address this (and to make the command-bar extensible).

Unfortunately for Neovim-Qt, `ext_messages` is only supported when `ext_linegrid` is enabled. Currently, Neovim-Qt only supports the legacy cell-based API.

The `ext_linegrid` API is off by default, and can be enabled via command-line:
        `neovim-qt --ext-linegrid`

The API renders most things correctly but is not yet 100%.

**Outstanding Issues:**
1. ext-linegrid cannot be enabled 'on-the-fly'
2. 'default_colors_set' do not update previously-painted cells.

**Issue 1**
I am not sure if this is a limitation of Neovim's API or a short-coming in my implementation. The map for "hl_attr" is not re-sent for styles that have already been painted with the cell-based API. The default style is then used in places where it should not be.

I addressed this by setting 'ext_linegrid' at ui_attach time with a command-line argument.

**Issue 2**
To see the issues with default color update, try changing your color scheme.

The new API leaves default color changes for the client to handle. The client must now decide what should be re-painted given the new default colors. This is a departure from the old API behavior, which guarantees stale default colors are overwritten.

I think this can be addressed via `ShellWidget` in a future commit. A function which re-paints any cells using the default colors must be added. Such a function could be called from within `handleDefaultColorsSet(...)`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,6 +8,7 @@ set(NEOVIM_QT_SOURCES
   auto/neovimapi2.cpp
   auto/neovimapi3.cpp
   auto/neovimapi4.cpp
+  auto/neovimapi5.cpp
   auto/neovimapi6.cpp
   function.cpp
   msgpackiodevice.cpp
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,7 +2,20 @@
 configure_file(version.h.in ${CMAKE_CURRENT_BINARY_DIR}/version.h)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
-set(NEOVIM_QT_SOURCES util.cpp neovimconnector.cpp neovimconnectorhelper.cpp function.cpp msgpackrequest.cpp msgpackiodevice.cpp auto/neovimapi0.cpp auto/neovimapi1.cpp auto/neovimapi2.cpp auto/neovimapi3.cpp auto/neovimapi4.cpp auto/neovimapi6.cpp)
+set(NEOVIM_QT_SOURCES
+  auto/neovimapi0.cpp
+  auto/neovimapi1.cpp
+  auto/neovimapi2.cpp
+  auto/neovimapi3.cpp
+  auto/neovimapi4.cpp
+  auto/neovimapi6.cpp
+  function.cpp
+  msgpackiodevice.cpp
+  msgpackrequest.cpp
+  neovimconnector.cpp
+  neovimconnectorhelper.cpp
+  util.cpp)
+
 if(WIN32)
   list(APPEND NEOVIM_QT_SOURCES stdinreader.cpp)
 endif()
--- a/src/auto/neovimapi5.cpp
+++ b/src/auto/neovimapi5.cpp
@@ -1,0 +1,5135 @@
+// Auto generated 2019-06-08 16:53:42.502897 from nvim API level:5
+#include "auto/neovimapi5.h"
+#include "neovimconnector.h"
+#include "msgpackrequest.h"
+#include "msgpackiodevice.h"
+#include "util.h"
+
+namespace NeovimQt {
+/* Unpack Neovim EXT types Window, Buffer Tabpage which are all
+ * uint64_t see Neovim:msgpack_rpc_to_
+ */
+QVariant unpackBufferApi5(MsgpackIODevice *dev, const char* in, quint32 size)
+{
+	msgpack_unpacked result;
+	msgpack_unpacked_init(&result);
+	msgpack_unpack_return ret = msgpack_unpack_next(&result, in, size, NULL);
+
+	QVariant variant;
+
+	if (ret == MSGPACK_UNPACK_SUCCESS) {
+		switch (result.data.type) {
+			case MSGPACK_OBJECT_NEGATIVE_INTEGER:
+				variant = (qint64)result.data.via.i64;
+				break;
+			case MSGPACK_OBJECT_POSITIVE_INTEGER:
+				variant = (quint64)result.data.via.u64;
+				break;
+			default:
+				// TODO it would be nice if we could call back MsgpackIoDevice method or primitive types here
+				qWarning() << "Unsupported type found for EXT type" << result.data.type << result.data;
+		}
+	}
+
+	msgpack_unpacked_destroy(&result);
+	return variant;
+}
+#define unpackWindowApi5 unpackBufferApi5
+#define unpackTabpageApi5 unpackBufferApi5
+
+NeovimApi5::NeovimApi5(NeovimConnector *c)
+:m_c(c)
+{
+	// EXT types
+		m_c->m_dev->registerExtType(0, unpackBufferApi5);
+		m_c->m_dev->registerExtType(1, unpackWindowApi5);
+		m_c->m_dev->registerExtType(2, unpackTabpageApi5);
+		connect(m_c->m_dev, &MsgpackIODevice::notification,
+			this, &NeovimApi5::neovimNotification);
+}
+
+// Slots
+MsgpackRequest* NeovimApi5::nvim_buf_line_count(int64_t buffer)
+{
+	MsgpackRequest *r = m_c->m_dev->startRequestUnchecked("nvim_buf_line_count", 1);
+	r->setFunction(NeovimApi5::NEOVIM_FN_NVIM_BUF_LINE_COUNT);
+	connect(r, &MsgpackRequest::finished, this, &NeovimApi5::handleResponse);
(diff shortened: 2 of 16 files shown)


jgehrig:
**Related issues:**
- Change would conflict with #494 
- Adds unimplemented feature from #445

bfredl:
> I am not sure if this is a limitation of Neovim's API or a short-coming in my implementation. The map for "hl_attr" is not re-sent for styles that have already been painted with the cell-based API. The default style is then used in places where it should not be.

I'm not sure what the issue is here. Items in the map is very rarely overwritten, only if the internal map becomes full. And then Nvim will also resend the affected cells.

> The new API leaves default color changes for the client to handle. The client must now decide what should be re-painted given the new default colors. This is a departure from the old API behavior, which guarantees stale default colors are overwritten.

The simple solution would be to just repaint everything, that is what the TUI does. Changing the default colors is rare.

jgehrig:
> I'm not sure what the issue is here. Items in the map is very rarely overwritten, only if the internal map becomes full. And then Nvim will also resend the affected cells.

I'm not sure if we're talking about the same thing? Sorry for the confusion...

Most of the settings for neovim-qt are modified in `nvim_gui_shim.vim`. I am unable to set `ext_linegrid` on the fly there. This is more limitation than problem/issue :)

The following example won't work:
```patch
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim	Tue May 14 18:50:23 2019 -0400
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim	Sun May 26 01:51:00 2019 -0400
@@ -216,3 +216,8 @@
+
+function! s:GuiLineGrid(enable) abort
+	call rpcnotify(0, 'Gui', 'Option', 'LineGrid', a:enable)
+endfunction
+command! -nargs=1 GuiLineGrid call s:GuiLineGrid(<args>)

diff -r 57aa9ccef7bd src/gui/shell.cpp
--- a/src/gui/shell.cpp	Tue May 14 18:50:23 2019 -0400
+++ b/src/gui/shell.cpp	Sun May 26 01:51:00 2019 -0400
@@ -778,6 +798,8 @@
 		m_nvim->api2()->nvim_ui_set_option("ext_popupmenu", value.toBool());
 	} else if (name == "Cmdline") {
 	} else if (name == "Wildmenu") {
+	} else if (name == "LineGrid") {
+		m_nvim->api2()->nvim_ui_set_option("ext_linegrid", value.toBool());
 	} else {
 		qDebug() << "Unknown GUI Option" << name << value;
 	}
```
My fix for this is to set `ext_linegrid` before attaching the UI. An optional argument keeps `ext_linegrid` off by default.

> The simple solution would be to just repaint everything, that is what the TUI does. Changing the default colors is rare.
I will take a look at TUI for example code, thanks!

Anything that causes neovim to resend `grid_line` events will fix the display issue. I can manually trigger this with a window resize event. A hacky fix would be to re-size the window to the current window size.

Ideally, the refresh would be triggered via `ShellWidget` and require no additional `grid_line` input from neovim. It looks like this is possible, but may require modification to `ShellWidget`.

Also, the way `HighlightAttribute` stores default colors may need modification too. The functionality seems duplicate to `foreground()/background()/special()` anyways...

bfredl:
What would be the point of setting `ext_linegrid` "on the fly"? It is a protocol revision, not a feature. You can check if it is available in the metadata, if it is available there is no reason not to use it.

jgehrig:
>What would be the point of setting ext_linegrid "on the fly"? It is a protocol revision, not a feature. You can check if it is available in the metadata, if it is available there is no reason not to use it.

Setting the value "on the fly" isn't a necessity. This was merely an explanation as to why `ext_linegrid` is not modified via `nvim_gui_shim.vim` like most other settings.

The reason not to use it by default is stability on the Neovim-Qt side of things. Safer to make the feature opt-in at first, validate it, and then enable by default afterwards. However, this can easily be changed :)

Thanks!

equalsraf:
:+1:  on the approach for this.

First some high level points:

- I have no issues with the "on the fly" issue, several options can only be set on attachment
- any idea on which version ext_linegrid was first supported in nvim - I just ran this in 0.3.1 and got an unworkable UI. We may need a version check there (? 0.3.2 or 0.3.3).
- There is some overlap between shellwidget and HighlightAttribute (looks very similar to cell.h). I wonder if it makes sense to simply store the highlight attribute in the cell.

> Ideally, the refresh would be triggered via ShellWidget and require no additional grid_line input from neovim. It looks like this is possible, but may require modification to ShellWidget.

If it is infrequent I have no problem with this i.e. it is likely what happens when switching the colorscheme. This already happens when the background color is updated, a call to **update()** should suffice.

jgehrig:
Thanks! :smile:

>any idea on which version ext_linegrid was first supported in nvim - I just ran this in 0.3.1 and got an unworkable UI. We may need a version check there (? 0.3.2 or 0.3.3).

It looks like the feature existed as `ext_newgrid` for a short period of time, and is now supported as `ext_linegrid` for NeoVim 0.3.7+. The correct way to handle this is probably `Shell::handleSetOption(...)`, where we can enable the feature only when it is supported by the active NeoVim instance.

https://github.com/neovim/neovim/commit/43823acae279a09e4fb51407da86340a10714d38#diff-daa9e2681b8e40ea216038de23283f09

> If it is infrequent I have no problem with this i.e. it is likely what happens when switching the colorscheme. This already happens when the background color is updated, a call to update() should suffice.

Yes, the bug should only occur when the default colors are changed en mass.

It may be fixable if I handle default colors differently. I am not setting the default color via `ShellWidget`, instead I am storing the defaults fg/bg/sp as static values in `HighlightAttribte` and telling `ShellWidget` to paint the specific color. Performing default color handling in `ShellWidget` should help.

I will work on this, and report back...

>There is some overlap between shellwidget and HighlightAttribute (looks very similar to cell.h). I wonder if it makes sense to simply store the highlight attribute in the cell.

Regarding the overlap, I think we need both levels of abstraction. `HighlightAttribute` is intended to store the global highlight map specified by `ext_linegrid` protocol, and `Cell` is for storing the active values at any given cell.

In terms of storage for `Cell`, I agree it would be an improvement to store the `HighlightAttribute` object there. For example:
``` patch

struct Cell {
 ...
	QChar c;
+
+	HighlightAttribute hl_attr;
-	QColor foregroundColor, backgroundColor, specialColor;
-	bool bold, italic, underline, undercurl;
-	bool valid;
-	bool doubleWidth;
};
```
I would be happy to make the change above (or leave it alone), but would prefer to split it into a different Issue/PR.

bfredl:
> It looks like the feature existed as ext_newgrid for a short period of time, and is now supported as ext_linegrid for NeoVim 0.3.7+.

`ext_newgrid` was only in master, not in any released version. `ext_linegrid` is available since 0.3.2.

bfredl:
Though the canonical way to check for availability doesn't involve nvim version, instead the `ui_options` field of metadata should be used.

equalsraf:
Not sure if handleSetOption works for this.
Worst case scenario, you can use `nvim_apiN()` to get the api object. api level 5 seems to be 0.3.2. But apparently we never generated bindings for that version (:S only for api level 6 in e7eccc8a35054426d52ac77e7b76bff5ac215389).

So if anyway has a 0.3.2 lying around 

    make bindings NVIM=/path/to/nvim0.3.2

should create the files in src/autoneovimapi5. But using api6 is fine too.

I'm ok with leaving the cell changes to a followup PR.

jgehrig:
@bfredl Thanks for the version info!

jgehrig:
@equalsraf 
>Not sure if handleSetOption works for this.

You are correct.
The call to `handleSetOption()` occurs after the UI is attached, that won't work.

It looks like the "correct" way to handle this is to parse `nvim --api-info`. I will explore this route briefly.

If not, I can definitely make `nvim_apiN()` work.

> I'm ok with leaving the cell changes to a followup PR.

Sounds good. I will create an Issue/PR and link it here.

jgehrig:
Hey @equalsraf

The last commit makes persistently enabling this feature easier/cleaner.

~/.config/nvim-qt/nvim-qt.conf
```
[General]
ext_linegrid=true
```

I am not aware of any outstanding issues with the feature. It would be nice to get this checked in to master, so more people can use/test it.

Does you think this change needs any additional work before merge? Any other thoughts on ext_linegrid?

equalsraf:
Thanks @jgehrig this is great.

Issue #540, opened 2019-05-12
Window size and position


KVAnton-WEB:
Good job!
But there is not enough function to save the position and size of the window, to restore them when you restart program.

I'm use: KDE neon 5.15 (Ubuntu 18.04)

jgehrig:
Feature added with Pull #551.

Marking as closed.

PR #551, opened 2019-07-03
Save window size and position

jgehrig:
Adds support for the feature requested in issue #540.

When neovim-qt is restarted, the last known window size and position are restored.

The window state and geometry are stored via QSettings on each close event. Whenever a window is opened, QSettings is polled for the last known window geometry and state (if one exists).

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -107,6 +107,17 @@ void App::showUi(NeovimConnector *c, const QCommandLineParser& parser)
 	QObject::connect(instance(), SIGNAL(openFilesTriggered(const QList<QUrl>)),
 		win->shell(), SLOT(openFiles(const QList<QUrl>)));
 
+	// Window geometry should be restored only when the user does not specify
+	// one of the following command line arguments. Argument "maximized" can
+	// be safely ignored, as the loaded geometry only takes effect after the
+	// user un-maximizes the window; this behavior is desirable.
+	if (!parser.isSet("fullscreen") &&
+		!parser.isSet("geometry") &&
+		!parser.isSet("qwindowgeometry"))
+	{
+		win->restoreWindowGeometry();
+	}
+
 	if (parser.isSet("fullscreen")) {
 		win->delayedShow(NeovimQt::MainWindow::DelayedShow::FullScreen);
 	} else if (parser.isSet("maximized")) {
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -232,7 +232,11 @@ void MainWindow::reconnectNeovim()
 
 void MainWindow::closeEvent(QCloseEvent *ev)
 {
-	saveWindowGeometry();
+	// Do not save window geometry in '--fullscreen' mode. If saved, all
+	// subsequent Neovim-Qt sessions would default to fullscreen mode.
+	if (!isFullScreen()) {
+		saveWindowGeometry();
+	}
 
 	if (m_neovim_requested_close) {
 		// If this was requested by nvim, shutdown
@@ -276,9 +280,6 @@ void MainWindow::showIfDelayed()
 {
 	if (!isVisible()) {
 		if (m_delayedShow == DelayedShow::Normal) {
-			// Restore the last known window size/position.
-			loadWindowGeometry();
-
 			show();
 		} else if (m_delayedShow == DelayedShow::Maximized) {
 			showMaximized();
@@ -413,7 +414,7 @@ void MainWindow::saveWindowGeometry()
 	settings.setValue("window_state", saveState());
 }
 
-void MainWindow::loadWindowGeometry()
+void MainWindow::restoreWindowGeometry()
 {
 	QSettings settings{ "nvim-qt", "window-geometry" };
 	restoreGeometry(settings.value("window_geometry").toByteArray());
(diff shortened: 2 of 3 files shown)


equalsraf:
I am tempted to just merge this :D but i have no way to test it since i have a model window manager.

equalsraf:
> I'm curious about the implications of saving state.- i assume saves maximized, minimized, i wonder what else it stores.

Answering my own question: yes it saves fullscreen mode.

Config files gets stored in ~/.config/nvim/nvim-qt.conf:
1. this is probably not what we want. But I dont see a way to use XDG_DATA_DIR in QSettings.
2. Can we use `QSettings settings("nvim-qt", "nvim-qt");` otherwise people will start wondering what that file is.
3. Any ideas on how this interferes with **--geometry**?

jgehrig:
I can test anything you would like and report back :)

My primary platform is Linux, with VMs for Windows/Mac. I will do a test pass on all 3 platforms, and summarize what I tested/observed.

**1]** I'm not very familiar with the rules/conventions in this area... However, is `XDG_DATA_DIRS` correct? Or `XDG_CONFIG_HOME`? I think `XDG_DATA_DIRS` points to a shared location?

**2]** Yeah, I went back and forth on this. Using `~/.config/nvim-qt/nvim-qt.conf` seems reasonable to me.

**3]** Nope... But I can test it now! Let me know if there is anything specific (other than basic functionality) you feel needs testing.

equalsraf:
If you change the folder name to nvim-qt feel free to ignore 1) I dont see an easy way out - the proper location for non essential non-config files would be something like ~/.local/nvim-qt/nvim-qt.conf (or whatever is defined in $XDG_DATA_DIRS) but  QSettings does not support this.

- For 3) I would expect --geometry to be ignored if the file is in place
- I wonder if minimized windows would stay minimized

jgehrig:
I just looked... Qt handles the `$XDG_DATA_DIRS` stuff for us! :D
[QSettings Documentation](https://doc.qt.io/qt-5/qsettings.html#platform-specific-notes)

That all sounds good. I will do some cross-platform validation, and report back.

AllenDang:
Pls merge this! I just want this feature sooooo much!

jgehrig:
@equalsraf Sorry for the delay... The past few weeks have been busy for me.

Does `nvim-qt --geometry WxH` work on Windows/Mac? Testing with an (unmodified) `master` branch, I can only get this option do work on Linux...

Window size memory works well on all platforms. 

On Linux, everything behaves reasonably well. The `--geometry` argument will override the stored size as long as the last window state is not "maximized". If the window was previously maximized, it uses the geometry size when the user un-maximized their window. This behavior seems logical to me.

For example (Linux Only):
1. Open nvim-qt
2. Maximize nvim-qt
3. Close nvim-qt
4. Start command `nvim-qt --geometry 800x600`
5. Window starts in maximized state.
6. Un-maximize window.
7. Window is now 800x600

Let me know what you think :)

equalsraf:
> **Does nvim-qt --geometry WxH work on Windows/Mac? Testing with an (unmodified) master branch, I can only get this option do work on Linux...

In other systems it is **-qwindowgeometry** ([QGuiApplication](https://doc.qt.io/qt-5/qguiapplication.html))

jgehrig:
Ahhh, I got everything to work with `--qwindowgeometry`, thanks!

Windows/Mac exhibit the same maximize behavior as Linux. What do you think about the maximize behavior?

equalsraf:
I would like to avoid that, it is pretty awkward if a user explicitly calls --geometry and gets something else.

Perhaps the  way is to disable loading the saved geometry if geometry/qgeometry are passed in. There are also options --maximized and --fullscreen.

This looks like a good place to do it https://github.com/equalsraf/neovim-qt/blob/master/src/gui/app.cpp#L110 that way you can disable it for all similar cases

It might be easier to enable disable this with a flag that gets verified in `loadWindowGeometry()` e.g. assuming the default is NOT to load geometry 

```c++
        // shooting C++ off the hip - lets assume this is all wrong
	if (parser.isSet("fullscreen")) {
		win->delayedShow(NeovimQt::MainWindow::DelayedShow::FullScreen);
	} else if (parser.isSet("maximized")) {
		win->delayedShow(NeovimQt::MainWindow::DelayedShow::Maximized);
        } else if (parser.isSet("geometry") || else if (parser.isSet("qgeometry") ) {
		win->delayedShow();
	} else {
               // if the user did not specify any window geometry options, restore
               // previous state
                win->setRestoreGeometry(true); // default is false
		win->delayedShow();
        }
```

Does that sound ok?

jgehrig:
> I would like to avoid that, it is pretty awkward if a user explicitly calls --geometry and gets something else.

Sure, that seems reasonable to me. 👍 

Your example seems like a fine solution, I can write something along those lines. Thanks!

jgehrig:
Closed accidentally...

I've excluded restoreWindowGeometry() for the arguments `fullscreen`, `geometry`, and `qwindowgeometry`.

I disabled saveWindowGeometry() for `fullscreen`, preventing every subsequent open from also being fullscreen.

For `maximize`, we want to keep load/save behavior. The argument takes precedence over restoreWindowGeometry(), and the last window size/position will persist if the user un-maximizes the window.

equalsraf:
Happy times, thanks @jgehrig

PR #556, opened 2019-07-31
shellwidget: improve text clipping

bkoropoff:
- Strictly clip glyphs to the left/top/bottom of cells.
  This prevents pixel junk when a segment is later scrolled.
- Render right-to-left, and give extra clipping leeway on the right.
  This avoids clipping the right side of glyphs that are a bit too wide
  compared to the rest of the font.

Before:

before

After:

after


equalsraf:
The right to left painting was one smooth approach there.

I'm kind of wondering how the clipping will play out for people that force the use of non-monospace fonts and other irregular font issues. But I'm pressing the merge button since this should be the correct approach (TM).

Thanks @bkoropoff

PR #557, opened 2019-08-02
Fun times with unicode

bkoropoff:
This fixes various forms of misrendering involving Unicode:

- Surrogate pairs aren't handled
- nvim-qt disagrees with nvim about what characters are double-width
- wide characters sometimes cause misalignment of the rest of the line
- wide characters can leave cursor afterimages when scrolling

Some of the fixes are stopgaps until ext_linegrid support lands.  They aren't perfect, but are strictly better than before.  Consult the commit messages for explanations.

Before:
before

After:
after


bkoropoff:
OK, I guess there are two copies of `konsole_wcwidth.cpp`.

equalsraf:
Very cool.

Concerning konsole_wcwidth.cpp, it came from konsole, but has since been replaced upstream with https://invent.kde.org/kde/konsole/blob/master/src/CharacterWidth.cpp

I'm wondering if the scroll behaviour was different with earlier versions of Qt, i seem to remember this being different.

bkoropoff:
I took the new tables directly from the autogenerated code in nvim to be sure.  `ext_linegrid` avoids a lot of the problems fixed here by representing cells structurally, so nvim-qt can just defer to nvim's segmentation rules.

I determined the scroll behavior empirically with trial-and-error, as the documentation doesn't seem to say anything about it explicitly.  The new code seems to work on Linux with Qt 5.12 and Windows with Qt 5.13.  I also found at least once reference in the code to translating dirty regions during scrolling:

https://github.com/qt/qtbase/blob/5.12/src/widgets/kernel/qwidget.cpp#L10963

bkoropoff:
Is there anything blocking this from being merged?

bkoropoff:
@equalsraf This seems to work fine with `ext_linegrid` after rebasing

equalsraf:
Merged. @bkoropoff, truly, thanks for this.

equalsraf:
Strange unit tests failed on master, but why did they not fail before.

Issue #541, opened 2019-05-18
Multiple cursors displayed when used with neovim master


svermeulen:
Neovim version:
```
NVIM v0.4.0-770-ga3e0f05de
Build type: Debug
LuaJIT 2.0.5
Compilation: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/Users/svermeulen/Projects/Ave/neovim/build/config -I/Users/svermeulen/Projects/Ave/neovim/src -I/Users/svermeulen/Projects/Ave/neovim/.deps/usr/include -I/Library/Frameworks/Mono.framework/Headers -I/usr/include -I/Users/svermeulen/Projects/Ave/neovim/build/src/nvim/auto -I/Users/svermeulen/Projects/Ave/neovim/build/include
```

This issue only occurs with neovim-qt.  If I run terminal neovim and then create a terminal buffer it does not occur.

Repro steps:
- Execute `nvim-qt -- -u NORC`
- Enter terminal with `:terminal`
- Start typing and notice two cursors.  It seems as if one cursor is always one movement behind the other.  If I add whitespace by pressing space, the first cursor remains at the last letter.   If I add another letter, then it is always one character behind

![image](https://user-images.githubusercontent.com/810762/57974456-32ab6400-796d-11e9-88bf-9f3dd6e04d28.png)

![image](https://user-images.githubusercontent.com/810762/57974458-3c34cc00-796d-11e9-97f2-9570cafa9697.png)

bfredl:
Does https://github.com/neovim/neovim/pull/9882 (rebased on master) help?

svermeulen:
Nope.  If I build from that branch there's no bug.  But if I rebase that branch on to neovim master then I still get multiple cursors

Kraust:
[...]

PR #558, opened 2019-08-02
shell: fix duplicate cursors in terminal mode

bkoropoff:
Fixes issue #541 

In terminal mode, nvim draws the cursor itself by writing space characters with the "reverse" attribute.  The only apparent signal to the UI is that "busy" is turned on until terminal mode is left.  The UI API documentation recommends not drawing the cursor while nvim is busy, which is how double cursors are avoided by other UIs.  Do the same thing.

I kept setting the mouse to a busy icon, but it stays that way for the entirety of terminal mode despite nvim being responsive.  Maybe it should be disabled until the busy state is not semantically overloaded.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -613,6 +613,9 @@ void Shell::handleSetTitle(const QVariantList& opargs)
 
 void Shell::handleBusy(bool busy)
 {
+	if (busy != m_neovimBusy) {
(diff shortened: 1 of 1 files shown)


laurent-george:
Thanks, it works fine now !

equalsraf:
Thanks @bkoropoff

Issue #562, opened 2019-08-07
MSVC Compiler Errors msgpackc-3.0.1


jgehrig:
MSVC has several integer sign mismatch errors while compiling msgpackc-3.0.1:

MSVC Compiler Errors:
```
...\neovim-qt\third-party\msgpack-c-cpp-3.0.1\src\objectc.c(158,41): error C2220:  warning treated as error - no 'object' file generated [...\neovim-qt\build\third-party\msgpack-c-cpp-3.0.1\msgpackc-static.vcxproj]
...\neovim-qt\third-party\msgpack-c-cpp-3.0.1\src\objectc.c(158,41): warning C4267:  'return': conversion from 'size_t' to 'int', possible loss of data [...\neovim-qt\build\third-party\msgpack-c-cpp-3.0.1\msgpackc-static.vcxproj]
...\neovim-qt\third-party\msgpack-c-cpp-3.0.1\src\objectc.c(468,41): warning C4267:  'return': conversion from 'size_t' to 'int', possible loss of data [...\neovim-qt\build\third-party\msgpack-c-cpp-3.0.1\msgpackc-static.vcxproj]
```

It looks like these issues have been resolved upstream of 3.0.1 in commit https://github.com/msgpack/msgpack-c/commit/09e06a8cb7886543cd79ae20e97433194ae1e1e7.

@equalsraf Can we update `MSGPACK_VERSION` to the latest version, `3.2.0`?

jgehrig:
Version 3.2.0 resolves these issues, and doesn't appear to cause any problems.

I'd be happy to create a Pull Request for this :)

equalsraf:
Ack, I see no issue in bumping msgpack up, thank @jgehrig

equalsraf:
Merged, closing.

PR #563, opened 2019-08-08
Update MessagePack-C to version 3.2.0

jgehrig:
A new version of msgpackc is available.

This update was prompted by a MSVC compiler error described in Issue #562.


equalsraf:
Thanks @jgehrig.

Issue #568, opened 2019-08-13
Default colorscheme is broken


jgehrig:
The default colors are not handled well by ext_linegrid.

Both instances are started with `nvim-qt -- -u NONE`.

Symptoms:
![image](https://user-images.githubusercontent.com/11207308/62967136-619f6d80-bdd6-11e9-9947-274f3ab24868.png)

Both instances of nvim-qt are identical (built from master) with ext_linegrid enabled. The Neovim host version is different.

Left: v0.3.7
Right: v0.4.0 (custom build)

I am not sure what the root cause is, but it looks related to light/dark background and default colors. The issues appear for `:colorscheme default`, but not others like `:colorscheme blue`.

I suspect we need more complex logic to pick the correct rendering color when none is provided. The color may need to change depending on version of the neovim host.

justinmk:
in Nvim 0.4 `:help ui-linegrid` note:

```
["default_colors_set", rgb_fg, rgb_bg, rgb_sp, cterm_fg, cterm_bg]      
        The first three arguments set the default foreground, background and
        special colors respectively. cterm_fg and cterm_bg specifies the     
        default color codes to use in a 256-color terminal.
                                                                           
        The RGB values will always be valid colors, by default. If no         
        colors have been set, they will default to black and white, depending 
        on 'background'. By setting the ext_termcolors option, instead       
        -1 will be used for unset colors. This is mostly useful for a TUI
        implementation, where using the terminal builtin ("ANSI") defaults 
        are expected.                                                        
                                                                              
        Note: Unlike the corresponding ui-grid-old events, the screen is not  
        always cleared after sending this event. The UI must repaint the
        screen with changed background color itself.
```

bfredl:
The handling of default colors did not change with `ext_linegrid` in this aspect. The difference is rather that 0.4 will _always_ supply valid colors, while 0.3.x still might use -1 values, which is a backwards compatible change. No extra "complexity" should be needed compared to the previous version, nvim-qt should just be prepared to handle -1 defaults _regardless_ of mode.

bfredl:
More specifically: `handleDefaultColorsSet` needs to check `if (rgb_fg/bg == -1) ` just like `"update_fg/bg"` handler does.

jgehrig:
Just to be clear... Something wrong with my implementation, not `ext_linegrid` itself. 😄

Thanks for the info!

> The difference is rather that 0.4 will always supply valid colors, while 0.3.x still might use -1 values

[...]

Issue #390, opened 2018-01-31
set background not working  in ginit.vim


feluxe:
Hi there!

I have this in `init.vim`:

     set background=dark

and this in `ginit.vim`:

    set background=light

I expected `nvim` (terminal) to run in dark mode and `nvim-qt` (gui) to run in light mode. Unfortunately `nvim-qt` does not switch to light mode.

I have this `Guifont DejaVu Sans Mono:h8` in `ginit.vim` as well, which works fine.

Any idea what this could be?

OS: Arch Linux. Installed nvim-qt from source. `cmake=3.10.1`, `GNU Make 4.2.1`, `Qt=5.10.0`

jgehrig:
Working as of `v0.2.16`.

I think there are also changes in recent versions of `nvim` making this more likely to work on various frontends.

PR #571, opened 2019-08-17
Default colorscheme is broken

jgehrig:
Issue #568 
Issue #390 

Neovim can send color -1 for "default_colors_set", and this is not handled properly. This value should default to black or white based on the value of Neovim setting `background`.

A background setting was added to ShellWidget, which defaults to `dark`. On every call to `default_colors_set` an async `vim_get_option` lookup is performed. Everything renders using the last known (or default) value of background. When a change to this setting is observed, the screen is updated with the new default color settings.


equalsraf:
Great, thanks.

PR #573, opened 2019-08-17
Use SVG icon for better high-dpi scaling

jgehrig:
The ErrorWidget has scaling artifacts for High DPI displays. Converting from PNG to SVG, allows Qt to better render/scale icons images.

**Before:**
image

**After:**
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ jobs:
           command: sudo apt-get update
       - run:
           name: Install dependencies
-          command: sudo apt-get install -y neovim python3-pip python3-dev python3-msgpack python3-jinja2 cmake qt5-default
+          command: sudo apt-get install -y cmake libqt5svg5 libqt5svg5-dev neovim python3-dev python3-jinja2 python3-msgpack python3-pip qt5-default
       - run:
           name: build
           command: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,9 +38,10 @@ include(MacOSXPaths)
 # Qt
 set(CMAKE_AUTOMOC ON)
 find_package(Qt5Core REQUIRED)
-find_package(Qt5Widgets REQUIRED)
 find_package(Qt5Network REQUIRED)
+find_package(Qt5Svg REQUIRED)
 find_package(Qt5Test REQUIRED)
+find_package(Qt5Widgets REQUIRED)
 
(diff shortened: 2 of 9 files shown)


equalsraf:
I wonder, should we keep using the png for the app icon and similar and the svg just for the errorwidget?

jgehrig:
I think SVG is right icon to use everywhere. There is no guarantee what size an icon will be rendered at by the host platform, and SVGs render well at all sizes.

That said, I don't notice any difference in the actual rendering for app icon. I would be happy to use either, let me know.

SVG (80x80 -> 240x240):


PNG (80x80 -> 240x240):

jgehrig:
Regarding the continuous integration failures...

**CircleCI:** This is a real failure, the Qt install is missing SVG support. 
**AppVeyor:** Failure to download nightly, unrelated.
**Travis:** I will investigate, I cannot reproduce on Linux. Mac OS related?

equalsraf:
Sounds good.

My main concern on this one is that we runtime errors if the additional module is missing - need to test the windows builds before we merge this into master.

equalsraf:
I'm seeing the same Circle CI issue in my ubuntu install here - but not in Fedora. The CI for windows is also showing a similar problem.

jgehrig:
The changes run fine for me on Windows.

For Windows, we would want to make sure the correct DLL for Qt5SVG is shipped alongside `nvim-qt.exe`. For Mac/Linux the dependency should be common and easy to handle. 

**CircleCI:** Unrelated config/install failure. Needs a restart, but the issue should be fixed.
**AppVeyor:** Fixed.
**Travis:** Deterministic MacOS only failure, needs investigation.

```
2: FAIL!  : NeovimQt::Test::connectToNeovimTCP() Compared values are not the same
2:    Actual   (c->errorCause())             : NoError
2:    Expected (NeovimConnector::SocketError): SocketError
2:    Loc: [../test/tst_neovimconnector.cpp(65)]
```
I'm not directly touching NeovimConnector, but my changes seem to have an impact on MacOS only... :confused:

jgehrig:
@equalsraf

This has been rebased onto master. The changes from Pull 612 fix the TravisCI issues here too.

I did a quick test, and all images seems to render well. Ready for merge.

equalsraf:
Yet another reminder that linking order matters.

Thanks.

Issue #575, opened 2019-08-25
No bold font rendering at in Plasma - and only in Plasma


roflcopter4:
**OS:** Linux v5.2.9 (ArchLinux)
**QT:** ArchLinux reports version 5.13.0-7
**CC:** gcc 9.1.0

I've given my OS and compiler as per the guidelines but I have seen this issue on numerous systems and builds.

For some reason, and I've hurt my brain trying to figure out why, in Plasma neovim-qt does not render bold fonts. In some cases it doesn't render italic fonts either. However, when run in, for example, Cinnamon font rendering works as expected.

I initially thought this issue was related to my stubborn insistence on using a bitmapped font for programming. It seems few people do these days. Actually neovim-qt is the only editor that's not Extremely Buggy that supports them at all. But I have since noticed that it seems to have the same issue with TTF fonts too, though at least those do render italic. Just not bold.

I really have no idea what to offer here for debugging. I've spent an embarassing number of hours trying to work out the problem to no real avail. All I can say is that by hacking around in the source I've verified that, at least in most of `shell.cpp`, characters that are supposed to be bold are tagged as such. They just are not being rendered that way.

Here is a simple example. The first link is some sample code in neovim-qt. The second link is the same code rendered in a terminal. I could load up cinnamon and render it in neovim-qt there but can't be bothered.

https://i.imgur.com/Zuv479F.png
https://i.imgur.com/WMs0G0I.png

Note: no bold. Why? WHY?

equalsraf:
Can I assume that Plasma means KDE?

I would suspect an issue with KDE or how it setups fontconfig. Anyway these were my first results when searching for this

- https://forum.kde.org/viewtopic.php?f=227&t=139794
- https://www.reddit.com/r/kde/comments/88hzdb/konsole_bold_font/
- https://forum.qt.io/topic/78036/bold-fonts-not-working-for-many-fixed-width-fonts-in-linux
- https://bugs.kde.org/show_bug.cgi?id=378523

There are a couple of workarounds listed in that bug report, as well as KDE versions that are affected, does that help?

roflcopter4:
Yes, I meant KDE. Sorry.

Clearly I was searching for entirely the wrong things. Thank you for the links. That bug report was useful. And interesting too.

[...]

PR #576, opened 2019-08-25
Fix bold fonts not displaying in KDE (#575)

roflcopter4:
I explained the purpose of this tiny PR in the related issue (#575). This fixes it completely for me.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -151,10 +151,20 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 						p.setPen(foreground());
 					}
 
+					// We must specify the style to force bold/italic fonts
 					if (cell.bold || cell.italic) {
 						QFont f = p.font();
-						f.setBold(cell.bold);
-						f.setItalic(cell.italic);
+						if (cell.bold && cell.italic) {
+							f.setBold(true);
+							f.setItalic(true);
+							f.setStyleName(QStringLiteral("BoldItalic"));
+						} else if (cell.bold) {
+							f.setBold(true);
+							f.setStyleName(QStringLiteral("Bold"));
+						} else /* cell.italic */ {
+							f.setItalic(true);
+							f.setStyleName(QStringLiteral("Italic"));
+						}
 						p.setFont(f);
 					} else {
 						p.setFont(font());


equalsraf:
We need to do some further research on this. I'm not sure what this does when not running in KDE. And it seems some fonts are not affected (probably because of the naming).

I have not checked, but would setting the style string to empty also work?

To my knowledge there is no way to determine we are running with KDE's theme (at least without linking against KDE). If there is, that would be ideal since we could apply a special case for KDE.

For testing, it might be possible to do some testing by enable/disabling the use of KDE's theme (again, have not tested) using either CLI arguments or environment variables (e.g. -platformtheme https://doc.qt.io/qt-5/qguiapplication.html)

roflcopter4:
I did a quick check in Cinnamon and can say that it at least seems to work there too. I actually don't even have a Windows machine to check on at the moment.

I'll see if I can look into a special case later. For the moment I just thought I would verify that setting the style to an empty string (or in fact any arbitrary string) works just as well. Nonetheless, on KDE it still must be done. I feel like there's a smarter way that I'm completely missing here.

equalsraf:
> I would like to understand more about the root cause, and why/how this fix works. Why does this work?

Presumably this is preventing a check carried out by KDE that causes the wrong font to be loaded. I honestly dont have a clear picture of the flow between the font request that we make and the KDE platform plugins. Which is why I would prefer to have a KDE specific fix.

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -153,6 +153,7 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 
 					if (cell.bold || cell.italic) {
 						QFont f = p.font();
+						f.setStyleName(QStringLiteral(""));
```
equalsraf: Can we put a comment above this line

```
// #575: Clear style name. The KDE/Plasma theme plugin may set this
// but we want to match the family name with the bold/italic attributes.
```

equalsraf:
Does this also work if we move the code here? https://github.com/equalsraf/neovim-qt/blob/master/src/gui/shellwidget/shellwidget.cpp#L42

jgehrig:
@equalsraf That looks like a better location to me! From a very quick test-pass, it looks good on my Plasma machine...

It looks like moving the fix to `setShellFont()` also resolves the `:GuiFont Liberation\ Mono:h11:b` bug I pointed out in the Issue tracker...

``` patch
--- a/src/gui/shellwidget/shellwidget.cpp       Sat Aug 17 14:20:53 2019 +0100
+++ b/src/gui/shellwidget/shellwidget.cpp       Tue Sep 03 20:33:06 2019 -0400
@@ -40,6 +40,9 @@
 bool ShellWidget::setShellFont(const QString& family, qreal ptSize, int weight, bool italic, bool force)
 {
        QFont f(family, -1, weight, italic);
+       // Issue #575: Clear style name. The KDE/Plasma theme plugin may set this
+       // but we want to match the family name with the bold/italic attributes.
+       f.setStyleName("");
        f.setPointSizeF(ptSize);
        f.setStyleHint(QFont::TypeWriter, QFont::StyleStrategy(QFont::PreferDefault | QFont::ForceIntegerMetrics));
        f.setFixedPitch(true);
```

src/gui/shellwidget/shellwidget.cpp
jgehrig: I don't think `QStringLiteral()` adds anything here? It should be converted to QString by the compiler anyways.

Just `""` should suffice :)

equalsraf:
Ok, last request on my part then. Can you move the fix to the start of ShellWidget::setShellFont?

Issue #549, opened 2019-06-09
Code Cleanup: QColor to HighlightAttribute in Cell and ShellWidget


jgehrig:
With the changes to support "ext_linegrid" in #543 , the object `HighlightAttribute` was introduced. `HighlightAttribute` provides a clean abstraction for storing all of the format/styling options for a given cell.

We should go back and update the old API to use it where possible.

For example, the following change could be made:
``` patch

struct Cell {
 ...
	QChar c;
+
+	HighlightAttribute hl_attr;
-	QColor foregroundColor, backgroundColor, specialColor;
-	bool bold, italic, underline, undercurl;
-	bool valid;
-	bool doubleWidth;
};
```

Callers need to be updated, and there may be more opportunities for improvement than the example above.

It may make sense to handle default color registration as static storage within `HighlightAttribute`?

PR #577, opened 2019-08-28
QColor to HighlightAttribute in Cell and ShellWidget

jgehrig:
I finally got around to making these changes from your comments on Issue #543 :)

I wrote this without inheritance because I think `Cell` and `HighlightAttribute` should be different concepts. However, there would be less code with inheritance.

Closes Issue #549.

Let me know if you think this is worth the effort.


equalsraf:
To be clear, I want to merge this. CircleCI is not being helpful :S

jgehrig:
Ping @equalsraf

Let's merge this or close the PR. It looks like CircleCI is behaving again.

PR #581, opened 2019-09-04
Added argument to disable popup menu ext

paulondc:
I noticed there was no option to disable the popup menu extension. Therefore, this pull request provides a small addition to the args (--no-ext-popupmenu) that can be supplied to the executable (same behavior as no-ext-tabline).


equalsraf:
Thanks @paulondc I needed that.

PR #583, opened 2019-09-11
Clear font style name for new fonts

equalsraf:
The KDE/Plasma theme plugin (or others) may set the font styleName for a
new font. However we want to match the family name with the bold/italic
attributes instead of the stylename.

See #575

------

This is basically a dupe of https://github.com/equalsraf/neovim-qt/pull/576/ the updated message and new location for the fix.

@jgehrig this should be it right? Trying to get this in before tagging a new release.


jgehrig:
@equalsraf This looks correct to me.

If you can wait, I will explicitly test your PR this evening after work.

**nit pick:** You might get smaller/more efficient code with `QStringLiteral("")`? Not sure about the Qt specifics... If not that, then `{}` is usually better than `""` to avoid the implicit strlen calls.

equalsraf:
Ok, i'll make the change too.

Also I've added a fix for the CircleCI issue, so this should unblock the tests on #577.

jgehrig:
Thanks, that all sounds good to me!

Purely curiosity... Any reason we can't we specify like `image: circleci/classic:latest` or `ubuntu-1604:latest`? (No need to change your YAML)

I have no idea why `classic:latest` messed up the Qt install...

equalsraf:
No idea about the image. I just went with a version that other people reported had no issues with apt. Also running with an older version is not too bad, it keeps us from using bleeding edge Qt apis that are not supported in older versions.

jgehrig:
Thanks.

Definitely +1 to this!
> it keeps us from using bleeding edge Qt apis that are not supported in older versions.

jgehrig:
I just tested, and this works on my laptop with Plasma/KDE... Looks good! :+1: 

Both the status bar, and something like `:GuiFont Hack:h11:b` display properly.

PR #586, opened 2019-09-17
Add support for Neovim ext_cmdline

jgehrig:
A first pass at adding support for the Neovim external commandline.

image

**Outstanding Issues:**
- [ ] Test for Mac OS X.
- [ ] Test for Windows.
- [x] Test For Linux


jgehrig:
I can't explain the `...` characters being added to the bottom of the screen. Interestingly, nvim is explicitly sending events to paint these characters on the grid?

The behavior is dropped in nvim 0.4.0, so maybe we should do an API check to disable `ext_cmdline` for nvim < v0.4.0?

jgehrig:
image

Currently the `ext_popupmenu` integration is not well supported for v0.4.0. For earlier versions the TUI wildmenu is used. I will address this once we think the commanline is in a reasonable state.

justinmk:
Nice. This would be welcome for users who want top placement, or want the cmdline to overwite the statusline. I assume that will be configurable somehow?

jgehrig:
> Nice. This would be welcome for users who want top placement, or want the cmdline to overwite the statusline. I assume that will be configurable somehow?

Currently it is pinned to the center of the screen, but I can look into a configuration scheme. Making it more configurable would be good...

I would also like something similar to the TUI look/feel too. I can see users wanting ext_messages, without relocating the cmdline.

jgehrig:
@justinmk Do you know anything about the `...` behavior?

Shatur95:
Would be nice to have such feature. It's already implemented in [gonvim](https://github.com/akiyosi/goneovim).

jgehrig:
Yes, I will loop back to this feature eventually. The GUI still needs work, and ext_popupmenu lacks a few options to make this work well.

Thanks, that means the `...` behavior is almost certainly a client-side issue. We can set a minimum API version such that this issue does not appear.

svermeulen:
I tried testing this on OSX since I see that is one of the TODOs listed here.  After building neovim-qt from the branch here, the external command line does not show up.  Is there something I need to do to enable it?

jgehrig:
@svermeulen

This PR still needs work... The cmdline functionality should be reasonable, but the widget needs work. It will have display issues.

You need to run `:GuiCmdline 1` to enable the feature.

Issue #585, opened 2019-09-16
Unknown font: on latest nvim release


cmaughan:
I updated to Neovim 4.2, released yesterday.  When running nvim-qt I now get a startup error:
```
Unknown font:
```
I'm on windows.  I've tried it with NORC.  Same behavior.  nvim is fine, but nvim-qt is not.

jgehrig:
Thanks @cmaughan!

I would guess this is related to #575 and #583 . I can investigate on a Windows machine later today.

If you can compile and test, try deleting the following lines from `shellwidget.cpp`:
```C++
// Issue #575: Clear style name. The KDE/Plasma theme plugin may set this
// but we want to match the family name with the bold/italic attributes.
f.setStyleName("");
```

jgehrig:
I can't reproduce the issue (Windows or Linux).

Can you provide any additional details? How did you install?

Can you remove the line above, compile, and test?

cmaughan:
I'm not setup to compile this project, but I can try; I installed by just taking the latest neovim release and extracting it.
Somebody told me yesterday that on the reddit release thread there was a comment that the windows version needed more 'finessing'!  I don't have the link though.

cmaughan:
Note, I'm on Windows 
Version	10.0.17763 Build 17763

blueyed:
Another report: https://github.com/neovim/neovim/issues/11037

jgehrig:
I still cannot reproduce the issue on Windows 10, 10.0.18362 Build 18362 x64.

An explicit set of steps would be helpful: specific links, install location, every step/action, etc. 

@cmaughan 
 > I installed by just taking the latest neovim release and extracting it.

This is good to know, I installed from a different source. Do you have any more details?

@the42
Any additional details you can provide would be helpful.

cmaughan:
Steps:
1. Go to: https://github.com/neovim/neovim/releases
2. Download the 0.4.2 release (nvim-win64.zip)
3. Using WinZip, extract to a folder on your drive (for me this is c:/tools/neovim).
4. Open a windows command prompt.  
5. Navigate to c:/tools/neovim (if not in your path)
6. Run nvim-qt.exe

somini:
I can reproduce this on Linux (ArchLinux; nvim v0.4.2, nvim-qt v0.2.14-1).

somini:
> I can reproduce this on Linux (ArchLinux; nvim v0.4.2, nvim-qt v0.2.14-1).

[...]

Issue #584, opened 2019-09-15
List available fonts using GuiFont


nico-abram:
Similar to `set guifont=*`

jgehrig:
Does the command `:set guifont=*` list fonts with nvim? I don't get a list back (0.4.0 or 0.3.8)...

You could probably use [QFontDatabase](https://doc.qt.io/qt-5/qfontdatabase.html) to look up the available fonts, and return that information to neovim. You could either create a new function to do this in `nvim_gui_shim.vim`, or expand upon `:GuiFont`.

For GuiFont, you could call out the `fdesc == "*"` case in `bool Shell::setGuiFont(const QString& fdesc, bool force)`, and then issue an `:echo` with the list of fonts via `m_nvim->api0()->vim_command(...)` or similar. The approach would be similar for a new function too.

justinmk:
> Does the command `:set guifont=*` list fonts with nvim? I don't get a list back (0.4.0 or 0.3.8)...

I assume that's a gVim feature.

jgehrig:
You are correct, this is a gVim feature.

![image](https://user-images.githubusercontent.com/11207308/65055791-d158c900-d93d-11e9-891d-2302932931c4.png)

jgehrig:
Closing as merged.

See `:set guifont=*` or `:GuiFont *`.

Issue #491, opened 2018-12-31
Implement GuiFont *


gagbo:
Navigating Qt documentation I found this snippet : 
```cpp
void Notepad::on_actionFont_triggered()
{
    bool fontSelected;
    QFont font = QFontDialog::getFont(&fontSelected, this);
    if (fontSelected)
        ui->textEdit->setFont(font);
}
```

Maybe it could be used to implement a `:GuiFont *` like gvim does.

I'm just leaving this here since I'm trying to do something else, this is probably going to be a small PR soon. I also think it should be easy enough to add/test the snippet

equalsraf:
in vim-qt there is custom widget for this very purpose - mostly to filter monospace fonts https://github.com/equalsraf/vim-qt/blob/master/src/qt/fontdialog.cpp

gagbo:
Is all of this necessary with [FontDialogOptions](http://doc.qt.io/qt-5/qfontdialog.html#FontDialogOption-enum) ?

equalsraf:
Maybe not :D it was way back in Qt4

equalsraf:
Closing duplicate, see https://github.com/equalsraf/neovim-qt/issues/584

PR #588, opened 2019-09-18
Various Font Related Fixes

jgehrig:
This PR contains a few font-related fixes/improvements :)

I modified the `bool Shell::setGuiFont(...)` logic so users can view font descriptions with either the `:GuiFont` or `:set guifont` commands. Before option guifont would update when set from itself, but not when set from GuiFont.

**Issue #585:** Error message "Unknown font:" displayed for recent versions of neovim (0.4.2).

**Issue #584:** Add QFontDialog similar to gVim for `:GuiFont *` and `:set guifont=*`

image

Also... I'm sneaking in an empty implementation for `hl_group_set`. This feature is rather noisy on startup for Debug builds.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -1,5 +1,6 @@
 #include "shell.h"
 
+#include <QFontDialog>
 #include <QPainter>
 #include <QPaintEvent>
 #include <QDebug>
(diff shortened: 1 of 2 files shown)


jgehrig:
I have not tested this yet on Mac or Windows. Pending...

src/gui/shell.cpp
```diff
@@ -101,39 +102,61 @@ bool Shell::setGuiFont(const QString& fdesc, bool force)
...
-			bool ok = false;
-			qreal height = attr.mid(1).toFloat(&ok);
-			if (!ok) {
-				m_nvim->api0()->vim_report_error("Invalid font height");
-				return false;
+		qreal pointSize = font().pointSizeF();
+		int weight = -1;
+		bool italic = false;
+		for (const auto& attr : attrs) {
+			if (attr.size() >= 2 && attr[0] == "h") {
```
oczekp: I think small typo here, instead of "h" should be as previously 'h'.
I've tested this fix against https://github.com/equalsraf/neovim-qt/issues/585 issue on my Windows machine and it looks fine:)

jgehrig: You are correct... Nice catch!

It still works, but the typing is wrong, better to have `'h'`

jgehrig:
Related Issue #491

src/gui/shell.cpp
```diff
@@ -93,44 +94,69 @@ QString Shell::fontDesc()
...
-			bool ok = false;
-			qreal height = attr.mid(1).toFloat(&ok);
-			if (!ok) {
-				m_nvim->api0()->vim_report_error("Invalid font height");
-				return false;
+	bool setShellFontSuccess{ false };
+
+	if (fdesc == "*") {
+		bool fontDialogSuccess = false;
+		QFont font{ QFontDialog::getFont(&fontDialogSuccess, QWidget::font(), this, {},
```
equalsraf: This is the modal variant right?

I assume this blocks all event handling (resizes and painting).

equalsraf: QWidget::font() could be the current shell font()

equalsraf:
Very cool, this PR brings back [memories](https://github.com/equalsraf/vim-qt/commit/c241c62f98f715f111459158d0f1c68df8c61feb)

src/gui/shell.cpp
jgehrig: I think `QWidget::font()` gets the current shell font... Or is there a better way?

jgehrig: Interesting, Qt says `QFontDialog` is modal... However the user can still return focus to MainWindow before closing the dialog?

This results in weird behavior, and should be fixed...
image

equalsraf:
I'm using i3 here, and windows cannot be resized. This branch seems to be triggering some kind infinite resize loop - where neovim triggers a resize, but the window cannot resize which triggers a resize ad nauseum. This happens as soon as nvim starts up.

I dont see any obvious issue with the code though.

src/gui/shell.cpp
equalsraf: > I think `QWidget::font()` gets the current shell font... Or is there a better way?

I think my brain is suffering from selective C++ amnesia. I was expecting to just see font() there.

jgehrig:
> I'm using i3 here, and windows cannot be resized.

Interesting, I don't see this behavior with KDE. I assume this is a regular resize, and is unrelated to the odd `GuiFont *` resize behavior?

The changes for 5c9ee68 could trigger a loop without the following?
```
// Exit early if the font description does not change, prevent loops.
if (fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0)
```

I probably makes sense to split this up... The 'Unknown font' issue is impacting users, but everything else is lower impact.

equalsraf:
Adding some debug statement, the loop keeps setting two different font sizes

```
"DejaVu Sans Mono:h14"
"DejaVu Sans Mono:h11"
"DejaVu Sans Mono:h14"
"DejaVu Sans Mono:h11"
...
```

This DOES NOT happen when loading with `-u NORC`.

To reproduce it is enough to have an init.vim with

```
set guifont=DejaVu\ Sans\ Mono:h14
```

It seems the updateGuiFontVariables setting `guifont is causing this issue`. And commenting the guifont line makes it stop.

In this case the recursion check always fails because of the size difference.

alfredgeo:
On macosX I didn't get the popup window with this pull-request, should I add some specific compiliation flag ?

laurent-george:
On macos I didn't get the popup menu with this pull request. Should I use some specific flag during compilation to get it ? Do you need help in order to merge this pull request ?

jgehrig:
Hey @laurent-george

Sorry for the (very) slow response...

My only concern was the resize behavior from the screenshot above. After some additional testing on Mac/Windows, this is a Linux only issue (and probably only for certain configurations). The issue here is minor, and probably an issue with Qt's Linux implementation.

I just tested this change on Windows/Mac, and everything looks good. No resize issues.

Let's get to the bottom of your issue, and try to merge this change!

1) How are you triggering the font dialog? `:GuiFont *` or `:set guifont=*`
2) Do you have NVIM_QT_RUNTIME_PATH set? Do you have `:GuiFont` as a command?

A quick sanity check on someone else's PC before merge would be appreciated! :+1:

jgehrig:
@equalsraf I think this one is ready.

I have tested on Mac/Windows and the feature looks good.

On Linux everything looks fine, but the QFontDialog resize artifacts from above are still present.

The resize artifact is specific to my configuration (KDE w/Qt 5.13), and doesn't happen on all Linux machines. It smells like a Qt Bug... I've investigated a few workarounds, and there isn't much we can do. QFontDialog cannot be made asynchronous, and the window SizePolicy/FocusPolicy controls are not fine enough to disable resize temporarily without visible side-effect. IMO we should just let this be, it is only a small visual bug on limited configurations anyways.

equalsraf:
Thanks @jgehrig.

Mte90:
Seems that with the latest master my ginit.vim is not executed.
My ginit.vim content is:
```
if exists('g:GuiLoaded')
    Guifont! DroidSansMono\ Nerd\ Font:h10
    GuiTabline 0
    call GuiWindowMaximized(1)
endif
```
I see that the font is not changed, if I run manually the font change is executed.
So on my testing I reverted some commits before the latest master and found if that the commit is https://github.com/equalsraf/neovim-qt/commit/06b40dda38518a827f0e870d7768fcf36c453a4a  if I set the commit before that the font is configured.

jgehrig:
Sorry for the break @Mte90... I will take a look at this issue tonight. Thanks!

Mte90:
No problem it was an excuse to navigate a bit in the code of this software :-D

Issue #587, opened 2019-09-17
`QCommandLineParser: option not defined: "qwindowgeometry"` when starting nvim-qt 0.2.14


YorickPeterse:
As of version 0.2.14, starting `nvim-qt` on Arch Linux produces the the following warning:

    QCommandLineParser: option not defined: "qwindowgeometry"

This does not seem to prevent running of neovim-qt, but it seems to suggest that commandline arguments might not be parsed properly.

jgehrig:
This is probably related to #551. Are you running a Debug or Release build?

It is a harmless message, but we don't want it to display for Release builds.

YorickPeterse:
@jgehrig This is using a release build (see the [PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/neovim-qt#n22) for Arch Linux).

jgehrig:
Thanks for the confirmation, this should definitely be fixed.

https://github.com/equalsraf/neovim-qt/blob/fc27a4e8bfdc6efa095fa79cc7db6801bc111e50/src/gui/app.cpp#L120-L122

`QCommandlineParser` leaves something to be desired... Only one of the last two options will ever be set. Looking at the Qt source, `isSet` will trigger qWarnings for unkown arguments; we have no control.

I have a few ideas, and will post a fix shortly.

jgehrig:
Fix merged. Marking as closed.

Issue #565, opened 2019-08-09
Windows/Mac qwindowgeometry arg not listed in help message


jgehrig:
On Apple/Windows platforms, `--geometry` is listed as an argument, but it has no effect The argument `--qwindowgeometry` should be used instead.

See `nvim-qt --help` for details.

jgehrig:
I will create a PR for this, easy fix :)

If it is okay, I would also like to remove some platform specific ifdefs while here.

PR #590, opened 2019-09-19
QCommandLineParser geometry argument improvements

jgehrig:
**Issue #587:** QCommandLineParser issues a qWarning when isSet is called for an
argument that has not been defined. This is fixed by adding platform specific
logic to only check the argument which exists --geometry or --qwindowgeometry.

**Issue #565:** The help message on some platforms displays the wrong --geometry
argument on Windows/Apple platforms. The fix for Issue 587 is used to add the
correct help message.


equalsraf:
Merging. Thanks @jgehrig.

That is one strange piece of behaviour.

PR #592, opened 2019-09-26
Issue 585 Unknown font error neovim 0.4.2+

jgehrig:
Fixes **Issue #585**

Let's split up **Pull Request #588**: this fix is higher priority than the other fixes in that bundle. We shouldn't let those features delay this fix further.

**Issue Description:**
This scenario has always been hit, but recent versions of neovim display the
error messages more aggressively. Neovim sends a "set guifont" command at
startup with a blank font family. It is safe to ignore this because Neovim-Qt
sets the default font before attachment.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -49,14 +49,22 @@ bool ShellWidget::setShellFont(const QString& family, qreal ptSize, int weight, 
 	f.setFixedPitch(true);
 	f.setKerning(false);
 
+	// Issue #585 Error message "Unknown font:" for Neovim 0.4.2+.
+	// This case has always been hit, but results in user visible error messages for recent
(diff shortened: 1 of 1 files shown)


YorickPeterse:
I tried this patch using neovim-qt `master` and it resolves the problem described in https://github.com/equalsraf/neovim-qt/issues/585 nicely.

equalsraf:
@jgehrig thanks for the split

somini:
Any chance of cutting a new release including this?

https://github.com/equalsraf/neovim-qt/compare/v0.2.14...master

There's only 3 commits on `master`, and two of them are for issues affecting many people every time they open nvim-qt, with known fixes. Please take this as a friendly nudge, I'm a simple user lurking on the issue tracker. :grin:

equalsraf:
@somini I spend very little time in github these days or code - the friendly nudges are highly appreciated. I have tagged v0.2.15.

PR #602, opened 2019-10-23
Use cmake's Python package to find the interperter

jamessan:
As of cmake 3.12, the PythonInterp package is deprecated in favor of the Python package.

The Python package has existed for a long time, so this shouldn't affect cmake compatibility.


jamessan:
> The Python package has existed for a long time, so this shouldn't affect cmake compatibility.

Or I was completely wrong here, and it did get introduced in 3.12. :)

PR #603, opened 2019-10-23
circleci: Generate bindings during the build

jamessan:


jamessan:
Meh, I fail at getting circleci to use a different Python version.

equalsraf:
Any particular reason for this? Currently nvim comes from the unstable ppa, so we bump into regressions i suppose. But the code would not be used since we always use the oldest supported apis.

equalsraf:
The bang line on the script uses **python** maybe that is always python 2 in this ubuntu image.

equalsraf:
Manual override works


```diff
diff --git a/.circleci/config.yml b/.circleci/config.yml
index a17f8d0..ea2d2ef 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
             pyenv global system
             mkdir build
             cd build
-            cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_GCOV=ON ..
+            cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_GCOV=ON -DPYTHON_EXECUTABLE=/usr/bin/python3 ..
             cmake --build . --target bindings -- VERBOSE=1
             cmake --build . -- VERBOSE=1
       - run:

```

equalsraf:
I've merged this on the side, thanks @jamessan.

jamessan:
> Any particular reason for this?

Figured it'd be good to get _some_ coverage of building the bindings, so it would be detected if the process breaks for some reason.

Not super important, but thanks for massaging it into working form. :)

Issue #607, opened 2019-11-03
"vnoremap < <gv" doesn't work


aj3423:
I use two mapping to keep selection after indenting

    vnoremap > >gv
    vnoremap < <gv

The first line works fine, but the second line doesn't work. I tried the console *nvim.exe*, it works fine for both lines.  

Environment:  win10-64bit  nvim-qt 0.5.0

rsheasby:
Was just about to open a similar issue for this. I'm unable to map `<`.
`nnoremap < <<` doesn't work either. Works perfectly when using nvim directly from terminal though.

justinmk:
Maybe `<` needs to be sent as `<lt>`

rsheasby:
I’ve already tried that. No luck. 

Besides, if that were the case why would it work correctly in regular nvim?

justinmk:
? This is something nvim-qt needs to do in it's input routine.

alexeevit:
ubuntu 18.04, the same problem

jhimcras:
manjaro 18.1.1, KDE 5.17.3, kernel 5.3.12-1
Same problem.
```
void Shell::keyPressEvent(QKeyEvent *ev) {
//...
    QString inp = Input.convertKey(ev->text(), ev->key(), ev->modifiers());
    std::cout << insp.toStdString().c_str() << "\n";
//...
}
```
intput: <
output: <s-lt>

It works with `vnoremap <s-lt> <gv`.

jgehrig:
This looks like a bug in neovim-qt to me...

@justinmk When the user types `<` they are holding down the keys `Shift + ,<`. Should the client send `<S-lt>` or just `<lt>`?

I assume the client is supposed to send only `<lt>`, and neovim-qt needs something like the following:
``` C++
	const auto modNoShift { mod & ~Qt::KeyboardModifier::ShiftModifier };
	return QString{ "<%1%2>"}.arg(GetModifierPrefix(modNoShift)).arg("lt");
```

justinmk:
> @justinmk When the user types `<` they are holding down the keys `Shift + ,<`. Should the client send `<S-lt>` or just `<lt>`?

Just `<lt>`. Otherwise users would need to map `<s-lt>` instead of `<lt>`. Nvim does not alias the two, otherwise the `s-` prefix would be pointless.

You can test it like this (in any Nvim UI, including terminal Nvim):

```
:nn <s-lt> :echo 's-lt'<cr>
[...]

Issue #593, opened 2019-09-27
Pressing Control + Super inserts ^S


beelzebielsk:
When I press the Control and Super key simultaneously, it is as if ^S (Control + S) was sent to neovim-qt.

I'm using version 0.2.13 on arch linux. I tried this both with my normal rc file and with no rc file whatsoever. Same behavior.

jgehrig:
I cannot reproduce this behavior... Could you please elaborate?

To make sure we are on the same page... I assume Super/Meta/Windows are all the same key? :)

**Example:**
1. Using KDE/Plasma with nvim-qt (version 0.2.12; nvim 0.3.7).
2. Open nvim-qt with command `nvim-qt -- -u NORC`.
3. In `Insert` mode, press `<Ctrl + Windows Key>`.
4. Notice nothing is inserted.
5. Map `<C-S>` so we can observe when `<C-S>` is pressed, `:map <C-S> :ehco('<C-S>' pressed!')<CR>`
6. Press `<Ctrl + Windows Key>`.
7. Notice no message displays.

beelzebielsk:
Hi, thanks for getting back to me.

I believe they are all the same key, yes. Super is the Windows key on my keyboard.

1. Using i3 with nvim-qt (version 0.2.14, nvim 0.4.2).
2. Open nvim-qt command `nvim-qt -- -u NORC`.
3. In `Insert` mode, press <Ctrel + Windows Key>.
4. The characters `^S` appear in the buffer.

That's what happens for me.

I'm on Arch Linux if that helps.

jgehrig:
No worries, I'm glad to help!

> I believe they are all the same key, yes. Super is the Windows key on my keyboard.

Everyone has a different name for that key :)

> 1. Using i3 with nvim-qt (version 0.2.14, nvim 0.4.2).
> 2. Open nvim-qt command nvim-qt -- -u NORC.
> 3. In Insert mode, press <Ctrl + Windows Key>.
> 4. The character (characters?) ^S appear in the buffer.

I do not see the `^S` behavior for these steps. Perhaps Desktop Manager (KDE vs i3) has an impact?

**Some Debugging Ideas:**
Do you have the ability to easily boot into another Desktop Manager? If so, what happens?

Try building with the following `Shell::keyPressEvent()` patch applied:
``` patch
diff -r 5556b921b590 src/gui/shell.cpp
--- a/src/gui/shell.cpp Wed Sep 18 18:06:11 2019 -0400
[...]

Issue #579, opened 2019-09-01
cannot map <A-...> on macos


zgpio:
# Environment
* os: macos 10.14
* nvim-qt version: 5979623
* nvim version:  v0.4.0-1860-g0ec80d5f6

# Reproduce Step
1. `/Applications/nvim-qt.app/Contents/MacOS/nvim-qt -- -u NONE`
2. `:map <A-k> :echo "hello"<CR>`
3. try to input `<A-k>`

By the way, I found `<c-a-k>` can be mapped.

zgpio:
https://github.com/equalsraf/neovim-qt/issues/510#issuecomment-456190038 He also mentioned such problems.

zgpio:
I found that <A-k> was internally converted to <A- ̊>.

justinmk:
> I found that was internally converted to <A- ̊>.

That is the macOS default behavior. Though I would expect Qt to give nvim-qt the proper flags, and then nvim-qt just composes a `<A-x>` string as usual, rather than the default macOS behavior.

jgehrig:
This is interesting behavior from MacOS...

From `qDebug()` statements injected into `shell.cpp`:
```
QKeyEvent: QKeyEvent(KeyPress, Key_K, AltModifier, text="˚")
  convertKey: "<A-˚>"
QKeyEvent: QKeyEvent(KeyPress, Key_K, text="k")
  convertKey: "k"
QKeyEvent: QKeyEvent(KeyPress, Key_S, AltModifier, text="ß")
  convertKey: "<A-ß>"
```

The input is handled this way because MacOS generates `˚`, and Qt passes this along in `text` to our input layer.

I'm not sure it is possible to handle this case. How can we identify the user's intent as `˚` or `<A-k>`?

You could change the keyboard layout to Unicode Hex Input:
https://apple.stackexchange.com/a/350065

However, even with the alternate layout MacOS still generates odd key events
```
MacOS:
  QKeyEvent: QKeyEvent(KeyPress, Key_K)
    convertKey: ""

Linux:
  QKeyEvent: QKeyEvent(KeyPress, Key_K, AltModifier, text="k")
    convertKey: "<A-k>"
```

jgehrig:
@mikew Moving the discussion from Pull Request 612 here.

I am still somewhat hesitant on this issue... The current behavior isn't strictly invalid. I wonder if there is a good way to make the `Unicode Hex Input` keyboard work and preserve the Alt - Unicode Symbol input?

[...]

Issue #510, opened 2019-01-21
Alt Gr does not work


Hofsmo:
Hi,
When I push the Alt Gr button neovim-qt does not seem to recognize it. Neovim on the other hand recognizes it. I need Alt Gr for ~, $ and other characters since I have a Norwegian keyboard. Does anyone have a fix for this?

Hofsmo:
Hi, thanks for the reply.

I normally use vim in the gnome terminal. However, at work I have to use windows, so since I anyway was trying something new I decided to give neovim a shot. 

For the terminal emulator I testsed neovim with cmd and I could use Alt Gr to write ~,$, etc.

I don't have access to my work computer right now, but I will check tomorrow exactly which version of windows it is running and I will install some other Qt editor to check if Alt Gr works there.

Hofsmo:
@gagbo your solution kind of works. It works for some letters which I write using AltGr+key. The normal way to write ~ is a bit more complicated as it is normally written using AltGr+key then followed by space. Now I can write ~ by typing Ctrl+v AltGr+key+(any letter) then I have to delete the letter I made instead of pushing space.

gagbo:
It was solely for debugging purposes. `^V` writes the raw input neovim got, so it's just something to help debug the issue :)

So when you do ^V and type `AltGr+tilde_key+<Space>` nothing happens ? You have to type another letter ? When you type another letter, do you have the ~ on the letter or do you have 2 glyphs like `~n` (if you go `^V AltGr+tilde_key+n` for example)

ravicious:
I tried the Ctrl+v trick and it doesn't work for me, maybe I misunderstood something but I press `Ctrl + v` in insert mode and then `Alt + o`, but instead of getting `ó` I just get nothing, the editor is stuck in displaying highlighted `^`, as if it was awaiting a key.

Hofsmo:
> 
> 
> I tried the Ctrl+v trick and it doesn't work for me, maybe I misunderstood something but I press `Ctrl + v` in insert mode and then `Alt + o`, but instead of getting `ó` I just get nothing, the editor is stuck in displaying highlighted `^`, as if it was awaiting a key.

Hi, I think the combination you wrote after `Ctrl + v` maybe doesn't do anything. At least if I do `Ctrl + v` and something that doesn't do anything I get the behavior you describe. Have you tried some other combinations?

ravicious:
[...]

PR #612, opened 2019-12-08
Refactor Keyboard Input Code, Improve Test Coverage, Fix Bugs

jgehrig:
Several outstanding bugs are related to the way key input is handled. This code is difficult to modify and validate due to the usage of #ifdefs. Each platform has different behavior, and Unit Tests only validate the active platform.

This change does the following:
1) Get rid of all #ifdefs, to make the code easier to read.
2) Split platform specific behavior into separate compilation units.
3) Add UT coverage for all platforms, regardless of active platform.

I've also added fixes for Issue #593 and Issue #607.

**Issue 607:** The QKeyEvent for "<" contains a ShiftModifier. We need to remove this and send `<lt>` instead of `<S-lt>`.
**Issue 593:** Some versions of Qt can send `Key_Super_L`. We should ignore this event along with other modifier-only key events.
**Issue 579:** Cannot map <A-...> on MacOS. Keyboard layout "UnicodeHexInput" required.
**Issue 510:** Alt special key input does not work on MacOS. We are currently sending `<A-...>` instead of `...`.


jgehrig:
I have seen this same issue before in Pull Request #573. Perhaps this is an unreliable UnitTest?

```
FAIL!  : NeovimQt::Test::connectToNeovimTCP() Compared values are not the same
   Actual   (c->errorCause())             : NoError
   Expected (NeovimConnector::SocketError): SocketError
   Loc: [/Users/travis/build/equalsraf/neovim-qt/test/tst_neovimconnector.cpp(65)]
```

Both changes trigger the issue, but don't touch neovimconnector in any meaningful way. Neither change shows any observable build or runtime issues. I am not sure what is happening here...

src/gui/input.cpp
```diff
@@ -164,8 +164,11 @@ QString convertKey(const QString& text, int key, Qt::KeyboardModifiers mod) noex
 		return ToKeyString(GetModifierPrefix(mod), specialKeys.value(key));
 	}
 
+	// The key "<" should be sent as "<lt>"
+	// Issue#607: Remove ShiftModifier from "<", shift is implied.
 	if (text == "<") {
```
equalsraf: I wonder if should apply to other modifiers too. Alt/Cmd/Ctrl.

jgehrig: I'm not sure I follow... Could you elaborate?

equalsraf: Seems like pressing Alt+<, or Ctrl+<, also generate \<lt\> in terminal.

Having said that these things are keyboard layout specific, e.g. in some keyboard layouts < does not require a Shift key but \<gt\> does.

jgehrig: Ahhhh, I understand.

I can get `Alt + <` to map in the TUI:
```
:nn <A-lt> :echo 'Alt + "<"'<CR>
{ Press Shift + Alt + ,< }
{ Displays: Alt + "<"}
```

It sounds like `<S-lt>` is the only combination that should not appear.

equalsraf: > I can get Alt + < to map in the TUI:

Interesting, I cant. Carry on then.

equalsraf:
> I have seen this same issue before in Pull Request #573. Perhaps this is an unreliable UnitTest?
> 
> ```
> FAIL!  : NeovimQt::Test::connectToNeovimTCP() Compared values are not the same
>    Actual   (c->errorCause())             : NoError
>    Expected (NeovimConnector::SocketError): SocketError
>    Loc: [/Users/travis/build/equalsraf/neovim-qt/test/tst_neovimconnector.cpp(65)]
> ```
> 
> Both changes trigger the issue, but don't touch neovimconnector in any meaningful way. Neither change shows any observable build or runtime issues. I am not sure what is happening here...

Reran a couple of times locally and in TravisCI:

- locally in my Linux laptop it works
- for this PR it keeps failing
- on the master branch it succeeded

Aah the fun this will be ...

equalsraf:
From https://github.com/equalsraf/neovim-qt/issues/134 it seems that on Mac OS X I was seeing  key: < with modifier Shift and text: <. This was a while back and probably in a pt_PT layout though.

jgehrig:
Gahhh! The failure was deterministic for the SVG changes too, but only on MacOS. Very Strange.

I have MacOS VM that I can debug on. This will be interesting indeed...

equalsraf:
Mmmm, there is also the possibility that there is some issue with CI. I have enabled daily builds for the master branch. I'll try to test it a bit too in https://github.com/equalsraf/neovim-qt/pull/615

jgehrig:
With the SVG changes, I could reproduce the issue locally on MacOS every time.

jgehrig:
Based on looking at these failures in the debugger, my best guess is unreliable timeout/wait logic causes these failures.

In all cases `NeovimConnector::connectToHost(...)` is called, and executes correctly.

In the passing tests, a breakpoint at `NeovimConnector::setError(...)` is reached. In the failing tests, breakpoints set at `NeovimConnector::setError(...)` and `NeovimConnector::discoverMetadata(...)` are never reached.

I can make `tst_neovimconnector` pass with either hack below:
``` diff
NeovimConnector* NeovimConnector::connectToHost(const QString& host, int port)
...
	s->connectToHost(host, port);
+	s->waitForConnected();
	return c;
}
```

```diff
    void connectToNeovimTCP() {
        ...
        // The signal might be emited before we get to connect
-        SPYWAIT(onError);
+        SPYWAIT(onError, 50000 /*msec*/);

        QCOMPARE(c->errorCause(), NeovimConnector::SocketError);
        c->deleteLater();
    }
```

Maybe we can improve the test reliability by adjusting timeout values...

There are also a few very slow tests; maybe this is a good opportunity to improve the test performance too?

jgehrig:
@equalsraf

The last commit fixes the Travis error. AppVeyor is failing to download a nightly release of Neovim.

For Travies, MacOS takes ~50 seconds to timeout for the TCP socket sent to `127.0.0.1:64999` in some cases. Based on the debugger, we aren't touching anything in the execution path of this test.

Adding a 2-minute timeout resolves the issue, `SPYWAIT(onError, 120000 /*msec*/);`.

Do you have any opinions on such a solution?

Other that the long timeout, I don't like how `Neovimconnect::m_error` is initialized as `NoError`. It might be better to initialize it as `NotConnected`, and set `QAbstractSocket::connected()` to positively set `NoError`. Not sure how involved such a change would be...

The test could then become:
```C++
SPYWAIT(onError, 2500 /*msec*/);

QVERIFY(c->errorCause() == NeovimConnector::SocketError ||
    c->errorCause() == NeovimConnector::NotConnected);
```

jgehrig:
I also improved the test performance. Only a handful of SPYWAITS were causing most of the spinlock. Manually commenting and lowering the timeout value for these cases seems worthwhile.

I can spin off these changes, or include them in their own commit with this Pull Request.

equalsraf:
ACK I'm good with merging the fix along with this CR.

jgehrig:
@equalsraf 

That's all I've got for this PR, I think it is ready... Let me know if you feel it needs anything else :)

mikew:
No control / command / meta combos are working for me with this PR. On macOS Catalina, running ...

```
nvim-qt -- -u NONE
:split
^w^w
```

... and the focus doesn't move between splits.

src/gui/input.cpp
```diff
@@ -1,263 +1,224 @@
...
+			|| key == Qt::Key::Key_Super_R) {
+			return {};
+		}
+
+		// Issue #344: Ignore Ctrl-Shift, C-S- being treated as C-Space
+		if (mod & Qt::KeyboardModifier::AltModifier
+			|| mod & Qt::KeyboardModifier::ControlModifier
+			|| mod & Qt::KeyboardModifier::MetaModifier
+			|| mod & Qt::KeyboardModifier::ShiftModifier) {
+			return {};
```
mikew: This is breaking things like `^w` on macOS Catalina. For some reason, when `^w` is pressed (or most any key with control or command), `text` is empty, but `key` is still `87` (the valid ASCII code). Adding a check for `if (key > 255 && ...` fixes the issue without re-introducing #344, but I've no clue how portable that solution is.

jgehrig: I don't think the `key > 255` is safe for international keyboard layouts. I think it is possible to have Ctrl + {Non English Character} combinations, but I am not 100% sure.

Can you re-test if I revert this snippet?

jgehrig:
Thanks @mikew!

MacOS is a difficult platform for me to debug/test on, since I don't own a Mac...

I would like to write a unit test for this scenario, so it isn't regressed in the future. Could you provide some debug output for a few of the key sequences in question?

This patch will provide debug output for the `QKeyEvents` I need to fix the issue and create UTs:
```patch
diff -r a4b43d295f61 src/gui/shell.cpp
--- a/src/gui/shell.cpp Mon Dec 16 02:23:57 2019 -0500
+++ b/src/gui/shell.cpp Thu Dec 26 10:58:19 2019 -0500
@@ -1070,6 +1070,8 @@
                this->setCursor(Qt::BlankCursor);
        }
 
+       qDebug() << "QKeyEvent ev:" << ev;
+
        const QString inp = Input::convertKey(ev->text(), ev->key(), ev->modifiers());
        if (inp.isEmpty()) {
                QWidget::keyPressEvent(ev);
```

src/gui/input.cpp
jgehrig: Ready for re-test :)

If you can get me the relevant QKeyEvents too, that would be most appreciated!

mikew: So this gets basic combos back, but now any keypress involving shift gets ignored:

```
// w key down
QKeyEvent ev: QKeyEvent(KeyPress, Key_W, text="w")
sent to vim: 'w'
// ctrl key down
QKeyEvent ev: QKeyEvent(KeyPress, Key_Meta, MetaModifier)
sent to vim: ''
// ctrl+w
QKeyEvent ev: QKeyEvent(KeyPress, Key_W, MetaModifier)
sent to vim: '<C-W>'
// ctrl key down
QKeyEvent ev: QKeyEvent(KeyPress, Key_Meta, MetaModifier)
sent to vim: ''
// shift key down
QKeyEvent ev: QKeyEvent(KeyPress, Key_Shift, ShiftModifier|MetaModifier)
sent to vim: ''
// ctrl+shift+w
QKeyEvent ev: QKeyEvent(KeyPress, Key_W, ShiftModifier|MetaModifier)
sent to vim: ''
```

I wonder if even ever using `ev->text()` is the right thing to do. On macOS, `alt+x` gives the unicode symbol `≈`, so they get sent to vim as `<A-≈>`. Whether it should be sent as simply `≈` or `<A-x>` is debatable.

mikew: And if it helps, here's what `ctrl+shift` looks like (I removed the `if` block checking just to show what would naively be sent to vim)

```
// ctrl key down
QKeyEvent ev: QKeyEvent(KeyPress, Key_Meta, MetaModifier)
sent to vim: ''
// ctrl+shift
QKeyEvent ev: QKeyEvent(KeyPress, Key_Shift, ShiftModifier|MetaModifier)
sent to vim: '<C- >'
// shift key down
QKeyEvent ev: QKeyEvent(KeyPress, Key_Shift, ShiftModifier)
sent to vim: ''
// shift+ctrl
QKeyEvent ev: QKeyEvent(KeyPress, Key_Meta, ShiftModifier|MetaModifier)
sent to vim: ''
```

looks like there's a difference somehow, depending on which key is pressed first.

mikew: Also seems like this is a known thing in Qt:

https://doc.qt.io/qt-5/qkeyevent.html#text

> Return values when modifier keys such as Shift, Control, Alt, and Meta are pressed differ among platforms and could return an empty string.

jgehrig: > I wonder if even ever using ev->text() is the right thing to do. On macOS, alt+x gives the unicode symbol ≈, so they get sent to vim as <A-≈>. Whether it should be sent as simply ≈ or <A-x> is debatable.

The scenario you mention is essentially the same as #579 :)

I would lean towards letting the platform escape the character... You can disable this behavior by setting you keyboard to Unicode Hex in MacOS. I think there is a good argument for either behavior. How can we identify user intent?

jgehrig: > So this gets basic combos back, but now any keypress involving shift gets ignored:
...
// ctrl+shift+w
QKeyEvent ev: QKeyEvent(KeyPress, Key_W, ShiftModifier|MetaModifier)
sent to vim: ''

Does this work in master? Based on a read of the code in master, I suspect no. The code for `text.isEmpty()` should be equivalent now... Could be a mistake on my part to.

If this works in master, we have a regression. If this doesn't, we have a long-standing bug. We can fix it either way.

Sorry, will need some help. I can't test natively on MacOS for a while...

jgehrig: IMO... This  is all too difficult to reason through manually. There are just too many corner cases based on platform, localization, and keyboard layout. That is why there are so many input bugs right now.

If we can write reliable unit tests for all of these scenarios, we ensure that each case is handled properly and we don't regress things.

mikew: > I think there is a good argument for either behavior. How can we identify user intent?

I'm really not sure here. I can see arguments for both ways: vim heavy users might expect `<M-x>` to come through, while "a good macOS citizen" might expect `≈`. Either way, what gets sent from nvim-qt (`<M-≈>`), seems to be the oddest.

For the record:

- macvim sends `≈`
- master sends `<M-≈>`
- I'd prefer it send `<M-x>`. While it's less "macOS", I think it's more intuitive, as `<M-x>` is `<M-≈>` while `<M-S-X>` is `<M-˛>`

> If this works in master, we have a regression. If this doesn't, we have a long-standing bug.

Control-shift / command-shift mappings don't work in master either, it's a long-standing issue. This PR made it much easier for me to track down, and that's appreciated.

> Sorry, will need some help.

No worries, I am happy to help out where I can. It's been a little hectic here due to holidays, but that's mostly over for now.

mikew: What about:

```cpp
		if (
			(
				key == Qt::Key::Key_Shift
				|| key == Key_Control()
			)
			&& (mod & Qt::KeyboardModifier::ShiftModifier)
			&& (mod & ControlModifier())
		) {
                }
```

That seems to stop #344 from happening on macOS, while not blocking anything else from going through.

I can't test on Windows, which sucks cause that was the platform of the original issue, but I can test on Linux.

As for the other issue I mentioned, I don't think it should hold this work back.

jgehrig: Thanks for the help, and happy holidays! Likewise, I may be in and out over the next few days...

I can handle Linux/Windows testing. But, MacOS is much less familiar and available to me.

With the current changes 344 is working again, no?

The simplest fix for 344 is probably:
``` patch
                if (key == Qt::Key::Key_Alt
                        || key == Qt::Key::Key_Control
                        || key == Qt::Key::Key_Meta
+                      || key == Qt::Key::Key_Shift
```

This should be functionally equivalent because:
`Key_Shift` implies mod & Qt::ShiftModifier
`Key_Control` implies mod & ControlModifier()

Regarding 579, more work would be required. I'm also not sure how this would interact with Issue 510 (pinging affected users).

I suspect to properly fix the issue, we would need to ignore `QKeyEvent::text()` and do something similar to `GetSpecialKeysMap()` to map a `Qt::Key` to a corresponding string.

mikew: > With the current changes 344 is working again, no?

Correct. Sorry if I what I said was confusing. I should have clarified that #344 is *still* fixed, in addition to other mappings now working on mac.

> The simplest fix for 344 is probably

Can confirm this works. Now, given that it's only checking `key`, can this then be moved near the top of `Input::convertKey`?

> I suspect to properly fix the issue, we would need to ignore QKeyEvent::text() and do something similar to GetSpecialKeysMap() to map a Qt::Key to a corresponding string.

That's what I'm thinking. I've tried a couple of things, both work:

- platform specific `QKeyEvent Input::NormalizeQKeyEvent(QKeyEvent *ev)`
- platform specific `QString Input::GetTextFromQKeyEvent(QKeyEvent *ev)`

They both boil down to the same thing on macos:

```cpp
QString GetTextFromQKeyEvent(QKeyEvent *ev) noexcept
{
	const QString text = ev->text();
	const int key = ev->key();
	const Qt::KeyboardModifiers modifiers = ev->modifiers();

        // Convert macOS Alt+ mappings back to their normal characters.
	if (
		!text.isEmpty()
		&& modifiers & Qt::KeyboardModifier::AltModifier
		&& !(
			modifiers & CmdModifier()
			&& modifiers & ControlModifier()
		)
	) {
		return QString(key);
	}

        // On macOS, QKeyEvent::text() is empty when pressing control or command.
        // https://doc.qt.io/qt-5/qkeyevent.html#text
	if (
		text.isEmpty()
		&& (
			modifiers & ControlModifier()
			|| modifiers & CmdModifier()
		)
	) {
		return QString(key);
	}

	return text;
}
```

Calling that where appropriate in `shell.cpp` fixes two oddities on macOS:

- Things like `Alt+X` get mapped to `M+X` now, instead of `M-≈`
- `text` is no longer empty when command or control are pressed on macos. That seems unexpected, and what was causing these issues with mapping `shift` on mac.

But honestly, this works as well:

```cpp
QString GetTextFromQKeyEvent(QKeyEvent *ev) noexcept
{
	const QChar c = QChar(ev->key());

	if (c.isPrint()) {
		if (!(ev->modifiers() & Qt::KeyboardModifier::ShiftModifier)) {
			return c.toLower();
		}

		return c;
	}

	return "";
}
```

Which just makes me think `QKeyEvent::text()` is dubious.

---

So there's a few ways of solving the `Alt+` mappings on macOS. Both the long and short versions work the same on all internation keyboards I tested (U.S, French, German - Standard, Belarusian).

I've no idea about AltGr, so can't weigh in on #510. But I do know things get extra hairy: nvim-qt will catch `Alt+E` and people can map it, but that puts macOS into accent aigu mode. So if the *does* map it, then they next type `a`, they'll get `á`.

jgehrig: Since 344 works again, I am going to resolve this thread.

Let's move the MacOS input normalization discussion over to #579. I think the fix will just complicate and slow down this Pull Request. It will be easier to review and validate small incremental changes.

I am still a little hesitant on this, `M-≈` is technically valid input according to the neovim input specifications.

> Can confirm this works.

Let's go with the simpler fix. I will force-push the `Qt::Key::Key_Shift` changes.

> Now, given that it's only checking key, can this then be moved near the top of Input::convertKey?

I would prefer to leave as-is. I think it is a good idea to keep the code similar to what was there before. This makes review and the diff/history easier to read. If I were writing from scratch, I would definitely do this.

>That's what I'm thinking. I've tried a couple of things, both work:
>
> - platform specific QKeyEvent Input::NormalizeQKeyEvent(QKeyEvent *ev)
> - platform specific QString Input::GetTextFromQKeyEvent(QKeyEvent *ev)
I prefer a `GetTextFromQKeyEvent` type solution. I don't think we should modify the QKeyEvent sent by the platform.

Of the two functions you provided, I prefer the first example.

I think `toLower()` is a code smell, and I prefer to avoid it. While safe for English, not all language characters map cleanly between cases. This can lead to some unexpected bugs...

jgehrig:
#510 should be fixed. There is now some MacOS specific logic to remove the `<A-...>`, which is incorrectly sent with special character sequences.

#579 should be fixed now if the user selects a keyboard layout that does not conflict with the bindings, such as Unicode Hex Input.

@mikew Are you still able to look the code over, and give everything a quick test?

jgehrig:
@equalsraf This should be ready to go.

Everything I have tested looks reasonable on Mac/Linux/Windows. I am not aware of any outstanding issues.

equalsraf:
Merged. Thanks @jgehrig. This is a tremendous amount of work.

jgehrig:
Thanks for the kind words! :)

equalsraf:
https://github.com/equalsraf/neovim-qt/graphs/contributors :clap:

Issue #599, opened 2019-10-14
Can't overwrite Cursor color over parenthesis, and can't disable reverse color


AgentCosmic:
As you can see on the image, when the cursor in on a parenthesis there is a purple background which I can't disable. The correct color should be white. Another problem is that the guifg will invert everywhere even though I explicitly set a color.

Is there a highlight group or option that I should be adding?

![image](https://user-images.githubusercontent.com/1035714/66758820-a4aeb780-eed1-11e9-8752-7ea43171cdfb.png)

equalsraf:
No. This is currently a limitation in neovim-qt, it always paints the cursor by inverting the colors instead of using the correct color

https://github.com/equalsraf/neovim-qt/blob/9875617f1f471031ee48cb4f51757ba73e08833b/src/gui/shell.cpp#L1038-L1052

AgentCosmic:
Oh, will it ever be supported? Cos I always get confused where my cursor's at, especially without the blinking option. Thanks.

jgehrig:
Maybe Pull Request #520 would resolve (or at least help with) this issue?

![peekV43L9Z](https://user-images.githubusercontent.com/11207308/67357314-e6bba700-f52a-11e9-921c-e62eb24bb724.gif)

Curious... What is the "correct" way to handle bracketing cursor color?

@equalsraf
We now have support for light/dark fg/bg issue last mentioned in the thread. Do you think the feature is worth merging if that code is finished?

jgehrig:
XOR cursor caused this issue. With #640, the logic is no longer present.

With the new `:set guicursor` implementation, this issue is not present.

Marking as closed.

Issue #619, opened 2019-12-19
scrolling artifacts on Windows after some idle time


jdhao:
I have got scrolling artifacts on the screen on Windows after nvim-qt has been idle for some time. 
See the white vertical line in the blow image.

![image](https://user-images.githubusercontent.com/16662357/71157314-bf5c2e80-227c-11ea-88d0-17e1f2150d0c.png)

The nvim-qt is bundled with nvim nightly version: NVIM v0.5.0-172-g1ff5b60cb

I have not observed this for previous versions of nvim-qt. This happens only recently.

jgehrig:
These look similar to the artifacts I see on Linux as a result of display scaling.

Is display scaling set to 100%? Does this issue persist with 100% display scaling?

I haven't seen this issue, but I will keep an eye out. If you can provide deterministic steps to reproduce the issue, that would be helpful.

jdhao:
Yeah, the Windows display scale is 100% and the screen resolution is set to `1920 x 1080`.

jdhao:
@jgehrig I think I can reproduce this issue now. It has something to do with the mouse wheel setting.

+ open nvim-qt with the following cofing (nvim-qt -- -u init.vim)
```
set mouse=nic
nnoremap <silent> <ScrollWheelDown> j
nnoremap <silent> <ScrollWheelUp>   k
```

+ Open a file
+ click other window or places to make the focus lost for nvim-qt 
+ Move the mouse cursor back to nvim-qt but not click the current buffer
+ scroll the mouse wheel down or up, and you will see those artifacts

![image](https://user-images.githubusercontent.com/16662357/71302248-07e22c00-23e4-11ea-909f-269687992c3f.png)

I am also confused that placing the mouse cursor on the nvim-qt window will activate it so that I can use mouse scroll to move the cursor inside nvim-qt.  I thought that only after clicking the nvim-qt window, will we be able to activate it and use the mouse wheel to scroll.

jgehrig:
I can't quite reproduce the exact scenario, but I see similar issues when the cursor "bounces" off the screen edge in an inactive state. The issue appears on Linux/Windows with 100% scaling.

The cursor outline disappears only after a "bounce" off the screen edge. Recorded with [ScreenToGif](https://www.screentogif.com/).

![output](https://user-images.githubusercontent.com/11207308/71312408-7e892480-23f8-11ea-9db7-70257237acfb.gif)

[...]

PR #623, opened 2020-01-14
Add GuiCursor Support

jgehrig:
Based on #520

Thanks @b-r-o-c-k, your Pull Request was instrumental in putting this together and understanding the various guicursor options. :smile: 

I have moved some of the cursor code from `Shell` into `ShellWidget`, so we can have a clean abstraction. Moving the cursor there allows us to keep `Cell` immutable, and avoid `const_cast`.  I have also integrated guicursor styling with the new `HighlightAttribute` map from work on ext_linegrid.

I think this still needs a minimum neovim API version, I am not yet sure which one we should use due to the highlight map and `hl_id` vs `attr_id`. I will investigate...

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -57,7 +57,7 @@ const QMap<int, QString>& GetSpecialKeysMap() noexcept
 
 static QVariant GetButtonName(
 	Qt::MouseButton bt,
-	int clickCount) noexcept
+	uint8_t clickCount) noexcept
 {
 	// NOTE: In practice Neovim only supports the clickcount for Left
 	// mouse presses, even if our shell can support other buttons
@@ -109,7 +109,7 @@ QString convertMouse(
 	QEvent::Type type,
 	Qt::KeyboardModifiers mod,
 	QPoint pos,
-	int clickCount) noexcept
+	uint8_t clickCount) noexcept
 {
 	const QVariant varButtonName{ GetButtonName(bt, clickCount) };
 	if (!varButtonName.isValid() || !varButtonName.canConvert<QString>()) {
--- a/src/gui/input.h
+++ b/src/gui/input.h
@@ -34,7 +34,7 @@ QString convertMouse(
 	QEvent::Type type,
 	Qt::KeyboardModifiers mod,
 	QPoint pos,
-	int clicksCount) noexcept;
+	uint8_t clicksCount) noexcept;
 
 /// Platform specific Qt key modifier bitmask for 'Control'.
 Qt::KeyboardModifiers ControlModifier() noexcept;
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -463,12 +463,7 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
 	} else if (name == "mouse_off"){
 		handleMouse(false);
 	} else if (name == "mode_change"){
-		if (opargs.size() < 1 || !opargs.at(0).canConvert<QByteArray>()) {
-			qWarning() << "Unexpected argument for change_mode:" << opargs;
-			return;
-		}
-		QString mode = m_nvim->decode(opargs.at(0).toByteArray());
-		handleModeChange(mode);
+		handleModeChange(opargs);
 	} else if (name == "cursor_on"){
 	} else if (name == "set_title"){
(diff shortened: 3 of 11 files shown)


jgehrig:
@equalsraf I am curious if you have any opinions on how `guicursor` should enabled/disabled for the GUI.

I had planned on adding an options-check in `shell.cpp` here:
``` C++
	if (true) // :GuiCursor? Setting to enable/disable?
	{
		ShellWidget::paintEvent(ev);
		return;
	}
```
We could do a few things:
1. Always on for clients that support `guicursor`
2. Add an option, `:GuiCursor` to this shim; default to on.
3. Add an option, `:GuiCursor` to this shim; default to off.
4. Add a temporary undocumented setting to `QSettings` as a fallback in case we break something.

I was thinking we could add an on-by-default setting `:GuiCursor` to the shim, and check `ginit.vim`. Any thoughts?

jgehrig:
Changes should fix Issue #599.

We should also investigate Issue #619, to see if it can be corrected while making these changes.

justinmk:
> any opinions on how `guicursor` should enabled/disabled for the GUI.

Why is `GuiCursor` command/function needed? Why not let it be controlled by 'guicursor' option? That's the purpose of Nvim emitting `option_set`--so that the options can work like gvim.

jgehrig:
@justinmk Good point. I suppose I didn't fully understand `cursor_style_enabled` from `mode_info_set`.

It is nice to have a fallback for disabling new features, in case something is broken. It looks like `set guicursor=` would be adequate in this case.

jgehrig:
The option `guicursor` will be enabled by default. Users could still fall-back to the old cursor via neovim's `:set guicursor=`.

I still need to check how well this behaves with older versions of `nvim`. I suspect very old versions will be fine, but versions in between may have issues with `hl_id` vs `attr_id`. Tests pending...

equalsraf:
> @equalsraf I am curious if you have any opinions on how guicursor should enabled/disabled for the GUI.

Enable by default since the existing implementation is a bit of an hack in any case. I dont think we need any specific commands for this.

jgehrig:
@equalsraf

Interesting... It looks like nvim-qt does not currently work with `nvim <= 0.3.1`.

The lowest working version is `nvim 0.3.2`, which still uses `attr_id`. So, this PR works fine with every working combination of nvim + nvim-qt.

 >Enable by default since the existing implementation is a bit of an hack in any case. I dont think we need any specific commands for this. We can fallback to current behaviour if something weird happens (like a missing option).

This sounds good to me. Let's get `guicursor` implemented, and keep the existing code as a fallback for the short term. Eventually, we can replace the old cursor implementation with `guicurosr` and some default styling options.

Once this is merged, I will create an Issue to track removal of the old cursor code.

jgehrig:
The current iteration works down to `v0.2.0`, however cursor styling is only applied when `ext_linegrid` is supported and enabled.

The fallback paths from `attr_id` to `hl_id` get convoluted. I will try to write some code than cleanly switches between the two. Alternatively, it may be easier to just require `ext_linegrid` for cursor styling due to the simpler code. The support would still be better than what exists today...

equalsraf:
> Alternatively, it may be easier to just require ext_linegrid for cursor styling due to the simpler code. The support would still be better than what exists today...

I'm ok with this :+1:

jgehrig:
Let's make `ext_linegrid` a requirement for cursor styling. Keep it Simple! :+1: 

A user that gets this `nvim-qt` patch is also likely to have a recent version of `nvim` too.

Supporting the old highlight API requires more complicated async code, and won't benefit many users. Attached is a (rough) patch to support `Neovim API 3` cursor styling.

[patch_legacy_grid_cursor.txt](https://github.com/equalsraf/neovim-qt/files/4179128/patch_legacy_grid_cursor.txt)

jgehrig:
@equalsraf 

Done. This should be ready for merge.

I have tested the basic features (blink, styling, shape/size), and everything looks good for new and old versions of `nvim`. We're going with the no cursor coloring without `ext_linegrid` approach.

A quick test and proof read of the code on someone else's machine wouldn't hurt... There is a decent amount of new and refactored code in this change :)

Thanks!

jgehrig:
@equalsraf

The merge with #643 has been handled. I did some quick re-testing, and everything still looks good.

Issue #626, opened 2020-01-23
GUI Treeview Not Working


jgehrig:
The `:GuiTreeviewShow` command is not working.

No directory contents are displayed:
![image](https://user-images.githubusercontent.com/11207308/72953730-106fae80-3d64-11ea-89ab-a270274ec060.png)

Based on a quick debug of the scenario, it looks like `TreeView::setDirectory(...)` is never being called. This causes the directory contents to remain un-populated.

This can be confirmed with a quick hack:
``` patch
diff -r 74e9449bd4ab src/gui/treeview.cpp
--- a/src/gui/treeview.cpp      Wed Jan 15 15:01:23 2020 -0500
+++ b/src/gui/treeview.cpp      Wed Jan 22 22:13:13 2020 -0500
@@ -61,6 +61,7 @@
 
 void TreeView::handleNeovimNotification(const QByteArray &name,
                                        const QVariantList &args) {
+       setDirectory("/some/valid/path");
        if (name == "Dir" && args.size() >= 0) {
                setDirectory(m_nvim->decode(args.at(0).toByteArray()), false);
        } else if (name == "Gui"
```

jgehrig:
The initial value for the current working directory is never set.

Something like this is required in the CTOR:
``` C++
	auto workingDir{ m_nvim->api0()->vim_call_function("getcwd", {}) };

	auto setWorkingDir = [this](uint32_t msgid, uint64_t f, const QVariant& r) noexcept
	{
		qDebug() << msgid << f << r;
		this->setDirectory(r.toString());
	};

	connect(workingDir, &MsgpackRequest::finished, m_nvim, setWorkingDir);
```

Some type checking and error handling is probably warranted. I am also not sure if this is thread safe?

equalsraf:
Interesting it does start empty, but in my case issuing a call to `:cd ` or `:cd /some/path` updates the view.

I'm running nvim 0.4.0 here, what version are you on?

On my end it just seems it is enough to initialize the treeview to the current tab directory when starting. I'm not sure if there is a way to force this event to be triggered from the nvim side.

jgehrig:
Yes, I observe the same behavior, and for multiple versions of neovim. My installed version is 0.4.1.

The snippet above works to trigger the initialization from C++.

Alternatively, we may be able to add something like `call rpcnotify(0, 'Dir', getcwd())` to the shim. It might work in `:GuiTreeviewShow` or on load of the shim?

[...]

PR #627, opened 2020-01-24
Issue 626: GuiTreeview empty on startup

jgehrig:
Fixes Issue #626 

The `autocmd WinEnter` does not trigger for the first window.

We can add an `autocmd VimEnter` to trigger the first initialization for the first window.


jgehrig:
@equalsraf

This one is ready for merge, and a simple fix. Tested working.

Issue #628, opened 2020-01-24
GuiFont! fails for bad monospace fonts


jgehrig:
@Mte90 is quoted below from #588:

Seems that with the latest master my ginit.vim is not executed.

My ginit.vim content is:
```
if exists('g:GuiLoaded')
    Guifont! DroidSansMono\ Nerd\ Font:h10
    GuiTabline 0
    call GuiWindowMaximized(1)
endif
```
I see that the font is not changed, if I run manually the font change is executed.
So on my testing I reverted some commits before the latest master and found if that the commit is 06b40dd if I set the commit before that the font is configured.

jgehrig:
This is an interesting bug...

It occurs due to `m_nvim->api0()->vim_set_option("guifont", fontDesc());` in `setAttached`.

Here is the resulting call order:
```
guifont ""
setAttached
GuiFont "Liberation Sans:h14"
guifont "Liberation Mono:h11"
guifont "Liberation Sans:h14"
```

The call to `guifont "Liberation Mono:h11"` is a result of trying to update `guifont` so that it reflects the displayed font in `setAttached`. Since `set guifont` does not support an option for forcing bad monospace fonts, the default font (Liberation Mono) overrides the user-set font (Liberation Sans).

I think we can safely get rid of the problematic line... We will still update the value of `guifont` when `:GuiFont` is called. This behavior seems similar to other gui options.

jgehrig:
Fixed in Pull #629, commit e2b0ac5d0736020573d125d094af0ba06c5bc362.

PR #629, opened 2020-01-24
Issue 628: GuiFont! fails for bad monospace fonts

jgehrig:
Issue #628

The line following line from setAttaced is invalid:
	`m_nvim->api0()->vim_set_option("guifont", fontDesc());`

This works for well formed monospace fonts, but does not work for bad
monospace fonts. This line updates the value of `guifont` with the default
font (valid monospace) on initialization.

If the user then sets a bad monospace font with GuiFont!, in some cases timing
of the neovim guifont events can override the user defined font.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -165,7 +165,6 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
 	// Updating guifont when the user has already called 'set guifont=...' may cause
 	// unwanted recursion. Only update this option for ':GuiFont', and dialog calls.
 	if (isGuiDialogRequest || updateOption) {
-		qDebug() << "Update guifont option!";
 		m_nvim->api0()->vim_set_option("guifont", fontDesc());
 	}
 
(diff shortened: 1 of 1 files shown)


Mte90:
Confirmed that works :-)

src/gui/shell.cpp
```diff
@@ -852,14 +850,21 @@ void Shell::handleMouse(bool enabled)
...
 {
-	if (args.size() == 2) {
-		QString fdesc = m_nvim->decode(args.at(1).toByteArray());
-		setGuiFont(fdesc, false /*force*/, true /*setOption*/);
-	} else if (args.size() == 3) {
-		QString fdesc = m_nvim->decode(args.at(1).toByteArray());
-		const bool force = args.at(2) == 1;
-		setGuiFont(fdesc, force, true /*setOption*/);
+	if (args.size() < 2
+		|| !args.at(0).canConvert<QByteArray>())
```
jgehrig: This should be `!args.at(1).canConvert<QByteArray>()`

jgehrig:
@equalsraf 

Ready for merge. Let's get this checked-in to unblock the issue.

equalsraf:
Merged, thanks @jgehrig

PR #630, opened 2020-01-24
Fix touchpad scroll speed

Finnerale:
This is a rebased and single commit version of #285 
Credit goes to @koalefant 
This is tested on X11 and Wayland with mouse and touchpad.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -1,5 +1,6 @@
 #include "shell.h"
 
+#include <cmath>
 #include <QFontDialog>
 #include <QPainter>
 #include <QPaintEvent>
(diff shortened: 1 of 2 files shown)


src/gui/shell.cpp
```diff
@@ -24,7 +24,8 @@ Shell::Shell(NeovimConnector *nvim, ShellOptions opts, QWidget *parent)
 	m_hg_foreground(Qt::black), m_hg_background(Qt::white), m_hg_special(QColor()),
 	m_cursor_color(Qt::white), m_cursor_pos(0,0), m_insertMode(false),
 	m_resizing(false),
-	m_mouse_wheel_delta_fraction(0, 0),
+	m_scroll_remainder(0.0f, 0.0f),
+	m_scroll_last_direction(0, 0),
```
jgehrig: Not necessary.

The default constructors create these values anyways.

src/gui/shell.cpp
```diff
@@ -1217,29 +1218,45 @@ void Shell::wheelEvent(QWheelEvent *ev)
...
+	// Scroll distance considering previous events
+	QPointF total_delta = scroll_delta_f + m_scroll_remainder;
+	QPoint scroll_delta = QPoint(int(total_delta.x()), int(total_delta.y()));
+
+	// Store remainder so we can accumulate fine scroll events
+	m_scroll_remainder = total_delta - QPointF(scroll_delta);
+
+	int horiz = scroll_delta.x();
+	int vert = scroll_delta.y();
+
```
jgehrig: No need for `horiz` or `vert`.

`if (scroll_delta.isNull()) {`

src/gui/shell.cpp
```diff
@@ -1217,29 +1218,45 @@ void Shell::wheelEvent(QWheelEvent *ev)
...
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 7, 0))
+	if (ev->inverted()) {
+		scroll_delta_f = -scroll_delta_f;
+	}
+#endif
+
+	// Reset scroll remainder if we change scroll direction;
+	int direction_x = int(scroll_delta_f.x() < 0.0f ? -1 :
+						  scroll_delta_f.x() > 0.0f ? 1 : m_scroll_last_direction.x());
```
jgehrig: You should remove the C-Style cast here, it will only suppress valid compile-time issues.

Also, this line is hard to follow.  A small `static` helper function might make this easier to read given the line below is almost identical.

Finnerale: I've moved it into a `normalize` function

src/gui/shell.cpp
```diff
@@ -1212,52 +1211,70 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
...
+	}
+	int direction_y = normalize(scroll_delta_f.y());
+	if (direction_y != m_scroll_last_direction.y()) {
+		m_scroll_remainder.setY(0.0f);
+		m_scroll_last_direction.setY(direction_y);
+	}
+
+	// Scroll distance considering previous events
+	QPointF total_delta = scroll_delta_f + m_scroll_remainder;
+	QPoint scroll_delta = QPoint(int(total_delta.x()), int(total_delta.y()));
```
jgehrig: This may result in an unnecessary copy/move. You should call the constructor directly instead:

`QPoint scroll_delta{ int(total_delta.x()), int(total_delta.y()) };`

Or better yet, maybe something like this would work:
```
const QPointF scroll_delta{ std::floor(total_delta.x()), std::floor(total_delta.y()) };

// Store remainder so we can accumulate fine scroll events
m_scroll_remainder = total_delta - scroll_delta;
```

No C-Style cast, so slightly safer. More clear the rounding is intentional. There are 2 less copies.

src/gui/shell.cpp
```diff
@@ -1212,52 +1211,70 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
...
+	if (ev->inverted()) {
+		scroll_delta_f = -scroll_delta_f;
+	}
 #endif
 
-	if (horiz == 0 && vert == 0) {
+	// Reset scroll remainder if we change scroll direction;
+	int direction_x = normalize(scroll_delta_f.x());
+	if (direction_x != m_scroll_last_direction.x()) {
+		m_scroll_remainder.setX(0.0f);
```
jgehrig: nitpick: `0.0f` is redundant. `0f` or `0.0`.

src/gui/shell.cpp
```diff
@@ -1212,52 +1211,70 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
...
+#if (QT_VERSION < QT_VERSION_CHECK(5, 5, 0))
+	int defaultDeltasPerStep = 120;
 #else
-	int horiz, vert;
-	horiz = ev->angleDelta().x();
-	vert = ev->angleDelta().y();
+	int defaultDeltasPerStep = QWheelEvent::DefaultDeltasPerStep;
+#endif
+
+	QPointF scroll_delta_f(float(ev->angleDelta().x()) / defaultDeltasPerStep,
```
jgehrig: Is the float cast here unnecessary?

I think there is an implicit cast based on the constructor's type of `qreal`. Loss of precision also occurs, `qreal` is usually `double`.

src/gui/shell.cpp
```diff
@@ -1212,52 +1211,70 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
...
-	QPoint total_delta = m_mouse_wheel_delta_fraction + ev->pixelDelta();
-	// Delta rounded to a minimal scroll step
-	QPoint cell_delta(total_delta.x() / scroll_step_x, total_delta.y() / scroll_step_y);
-	// Save remainder for future events
-	m_mouse_wheel_delta_fraction = total_delta - QPoint(cell_delta.x() * scroll_step_x, cell_delta.y() * scroll_step_y);
-
-	int horiz = cell_delta.x();
-	int vert = cell_delta.y();
+#if (QT_VERSION < QT_VERSION_CHECK(5, 5, 0))
+	int defaultDeltasPerStep = 120;
```
jgehrig: This should be `const`, it helps the compiler optimize the code.

You should mark any new variables below const too (wherever possible).

src/gui/shell.cpp
```diff
@@ -1212,52 +1211,70 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
 	}
 }
 
+static int normalize(float x)
+{
+    if (x == 0.0f) return 0;
+    else if (x < 0) return -1;
+    else return 1;
```
jgehrig: nitpick: Add `{}` braces, no need for else/elseif.

jgehrig:
@Finnerale 

A few more comments. Thank you for picking up the change, and addressing everything!

I'm not sure I have a computer to test this on... However, the modifications look reasonable to me.

src/gui/shell.cpp
Finnerale: Was necessary as angleDelta() returns a QPoint but I wrote it now by doing the division after scroll_delta_f was constructed which yields the same result.

equalsraf:
Looks good on my system but I dont have any device that supports this, maybe in a few days i can get one.

jgehrig:
@Finnerale 

So sorry for the delay... I now have a device to test this on. Wow, without this patch, scrolling is completely unusable on high-resolution touchpads!

I don't see any issues with the patch applied. I tested high-res and low-res scroll events.

We should merge this... Can you drop `40c5691` and `6d544a6`, and then rebase onto master?

jgehrig:
@equalsraf 

Are there any scenarios in particular you would like to see tested?

After a re-base, we should merge this...

Finnerale:
@jgehrig I've re-based it :+1:

equalsraf:
> Are there any scenarios in particular you would like to see tested?

cant say i do. i dont have a way to test this one.

I've merged this, thanks @Finnerale

Issue #199, opened 2016-10-16
Ctrl+Caps Lock inserts dollar sign


unclechu:
Why `Ctrl`+`Caps Lock` inserts dollar sign?

I'm actually using keyboard hack, and real `Caps Lock` enables by pressing both controls, so it means when `Caps Lock` is pressed `Control` keys isn't released yet. I could add release-events triggering before triggering caps lock to my hack app, but I guess this behavior of **neovim-qt** isn't right, don't you think?

P.S. I checked this without my app, just pressing really mapped caps lock and control key.

equalsraf:
Confirmed Ctrl+Caps inserts **$**, more acurately **\<C-$\>** is being sent, even though the key event text is empty.

A wrong assumption seems to be made in input.cpp about the key code, e.g. for this particular case

```diff
diff --git a/src/gui/input.cpp b/src/gui/input.cpp
index c5b4d6f..e584b47 100644
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -167,6 +167,8 @@ QString InputConv::convertKey(const QString& text, in
t k, Qt::KeyboardModifiers
                        QList<Qt::Key> keys = { Key_Control, Key_Alt, Key
_Cmd };
                        if (keys.contains((Qt::Key)k)) {
                                return QString();
+                       } else if (k == Qt::Key_CapsLock){
+                               return QString();
                        } else {
                                // key code will be the value of the char (hopefully)
                                c = QChar(k);
```

But we should probably check whats happening on Mac OS first.

shlomif:
Now it seems that in `nvim-qt -- -u NONE` on xfce on fedora 31 x86-64 capslock without a control also inserts a dollar sign. kate, and nvim handle it fine.

shlomif:
The bad symptoms disappear after I apply this forward ported patch of the one in an earlier comment:

```diff
diff --git a/src/gui/input.cpp b/src/gui/input.cpp
index 5e44293..7858a69 100644
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -166,6 +166,8 @@ QString convertKey(const QKeyEvent& ev) noexcept
 
 	if (specialKeys.contains(key)) {
 		return ToKeyString(GetModifierPrefix(mod), specialKeys.value(key));
+	} else if (key == Qt::Key_CapsLock) {
+		return QString();
 	}
 
 	// The key "<" should be sent as "<lt>"
```

[...]

PR #631, opened 2020-01-28
Issue 199: Control + CapsLock inserts $

jgehrig:
Issue #199 

When CapsLock is pressed Qt sends events such as:
	`QKeyEvent(KeyPress, Qt::Key_CapsLock, Qt::NoModifier, "")`

These key events do not get rejected as special keys, and the lack of a text
field triggers the conversion to '$'. These events should be ignored.


equalsraf:
Alright, thanks.

PR #637, opened 2020-02-01
Avoid printing Char when AltGr is pressed

Iazel:
# System specs
Debian 9
QMake version 3.0
Using Qt version 5.7.1 in /usr/lib/x86_64-linux-gnu
Input source: Italian
 
# Problem
Every time I press `AltGr` a weird character is immediately inserted. This behaviour only happens when using nvim-qt, nvim in terminal properly handles this case.

The debug output is:
```
QKeyEvent ev: QKeyEvent(KeyPress, Key_AltGr, GroupSwitchModifier)
   "ᄃ"
QKeyEvent ev: QKeyEvent(KeyPress, Key_AsciiTilde, GroupSwitchModifier, text="~")
   "~"
```
The key combination for tilde is `AltGr+ì` where `ì` is a button on the left of backspace.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -185,6 +185,7 @@ QString convertKey(const QKeyEvent& ev) noexcept
 		//   Issue#593: Pressing Control + Super inserts ^S
 		//   Issue#199: Pressing Control + CapsLock inserts $
 		if (key == Qt::Key::Key_Alt
+			|| key == Qt::Key::Key_AltGr
 			|| key == Qt::Key::Key_CapsLock
 			|| key == Qt::Key::Key_Control
 			|| key == Qt::Key::Key_Meta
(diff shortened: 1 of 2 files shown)


jgehrig:
Issue #510

jgehrig:
The fix looks good to me.

Can you please add a unit test in `test/tst_input_common.cpp` with the `QKeyEvent`s you provided in the issue?

An issue reference comment in the tests is also helpful:
`// Issue#510: Some Description of specific event case`

The tests help ensure this scenario is not regressed in the future :)

Iazel:
Sure, I'm not familiar with cpp, but I'll try to write a test for it :+1:

Iazel:
Ok, I've added tests and looks like are working properly. I've also squashed the branch, so it should be ready for merging :)

jgehrig:
Looks good, thanks!

One quick comment and it is ready... I'm not familiar with this keyboard layout, so my ability to test is limited.

If the unit tests pass, I am comfortable with the changes.

jgehrig:
Once squashed, this looks good to me.

Thanks for the fix! :+1:

Iazel:
Squashed, now it should be ready to go :)

jgehrig:
Thanks for the fix! 👍 

@equalsraf Looks good to me, ready to merge.

Issue #597, opened 2019-10-11
Other font weight for GuiFont


hosxy:
Could you please add other font weight (eg: semibold,) or add an arguments to provide font weight  for GuiFont  ?

equalsraf:
Is there any expected unit of weight for this?

Qt uses a scale of 0-99 for font weights, where 50 is Normal and 75 is Bold. The only intermediate value listed is 63 (called DemiBold). Anyway these seem consistent with vim-gtk:

- https://doc.qt.io/qt-5/qfont.html#Weight-enum
- https://vim.fandom.com/wiki/Setting_the_font_in_the_GUI

Not sure how this will play out with font metrics, but otherwise I dont see a reason not to do it.

It sounds like the implementation would be similar to the font height **hXX** (except this one is an integer). I do not want to use **wXX** here as it clashes with the width (which we dont support but vim in Mac OS does).

https://github.com/equalsraf/neovim-qt/blob/9875617f1f471031ee48cb4f51757ba73e08833b/src/gui/shell.cpp#L111-L118

jgehrig:
Closing. Since Pull #651, arbitrary font weight can now be set:

See `:help GuiFont`:
```
The following attributes are available

	hXX - height is XX in points (can be floating-point)
	wXX - weight is XX
	b   - bold weight
	sb  - semibold weight
	l   - light weight
	i   - italic
```

Under the hood, we're using [QFont::Weight](https://doc.qt.io/qt-5/qfont.html#Weight-enum). The range set of valid ranges [0-99].

NOTE: Rendering is handled by Qt, and you may not get full granularity on all platforms. Similar to #650.

Issue #532, opened 2019-04-03
arbitrary font weights


nilsbecker:
#136 added support for the "light" font weight. however, at least on os x there are multiple other font weights, including 'book', 'semibold', etc. in particular there is also the weight 'retina' which designates a font specifically optimized for rendering on a high-res display. it would be great to have some workaround to use all of these. for instance, being able to give the postscript name directly, e.g. `FiraCode-Retina:h12` etc.

jgehrig:
Support for arbitrary font weight was added in `v0.2.16`.

For example Hack; 11pt; 90% weight:
`:GuiFont Hack:h11:w90`

See `:help GuiFont` for more details.
```
The following attributes are available

	hXX - height is XX in points (can be floating-point)
	wXX - weight is XX
	b   - bold weight
	sb  - semibold weight
	l   - light weight
	i   - italic
```

Marking as closed.

PR #638, opened 2020-02-02
Support demi-bold font weight

NeilGirdhar:


jgehrig:
Technically this isn't supported by `guifont`. However, we already have light support and this seems useful.

You should also update the documentation for `GuiFont` so the feature is discoverable:
`src/gui/runtime/doc/nvim_gui_shim.txt`

NeilGirdhar:
@jgehrig thanks, it's useful for me since I was using a semibold with vim that I'd like to keep using.

jgehrig:
Agreed, it should should be an option in `guifont`.

Once this is squashed, the changes look good to me. :+1:

src/gui/shell.cpp
```diff
@@ -87,6 +87,8 @@ QString Shell::fontDesc()
 		fdesc += ":b";
 	} else if (font().weight() == QFont::Light) {
 		fdesc += ":l";
+	} else if (font().weight() == QFont::DemiBold) {
+	        fdesc += ":sb";
```
jgehrig: mixed tabs and spaces

NeilGirdhar: I noticed that.  Why do you even have tabs?

NeilGirdhar:
Great, thanks for taking a look so quickly, I'll squash these.

src/gui/shell.cpp
jgehrig: Because tabs are better ;)

jgehrig:
Sorry for the delay, I didn't have any fonts supporting `QFont::DemiBold`.

This doesn't work, there is an ordering issue with the font weight checks.

Example of a non-working scenario:
1. Run command `set guifont=*`
2. Select `Source Code Pro` with font style `Semibold`
3. Run command `GuiFont`
4. Notice font description has `:b` modifier, and not `:sb`

src/gui/shell.cpp
```diff
@@ -87,6 +87,8 @@ QString Shell::fontDesc()
 		fdesc += ":b";
 	} else if (font().weight() == QFont::Light) {
 		fdesc += ":l";
+	} else if (font().weight() == QFont::DemiBold) {
+		fdesc += ":sb";
```
jgehrig: This check cannot be reached: `font().weight() == QFont::DemiBold`

The `::bold()` function above returns true for semibold fonts too. The ordering needs to be changed, or the conditionals made more specific.

See: https://doc.qt.io/qt-5/qfont.html#bold

src/gui/shell.cpp
```diff
@@ -83,10 +83,12 @@ void Shell::fontError(const QString& msg)
 QString Shell::fontDesc()
 {
 	QString fdesc = QString("%1:h%2").arg(fontFamily()).arg(fontSize());
-	if (font().bold()) {
+	if (font().weight() == QFont::Bold) {
 		fdesc += ":b";
 	} else if (font().weight() == QFont::Light) {
 		fdesc += ":l";
+	} else if (font().weight() == QFont::DemiBold) {
+		fdesc += ":sb";
```
jgehrig: I think we should end with a check to `font().bold()` or `font().weight() > QFont::Medium`, and add a comment.

```C++
// Must be checked last: fallback to bold if weight is greater than Medium.
else if (font().bold()) { 
    fdesc += :b;
}
```

This is safer because `::weight()` can return values in between the enumerations. Function `::bold()` actually performs the check `::weight() > QFont::Medium`. That would more gracefully handle in-between values.

Sorry for the nitpick... I haven't observed any issues but it would be safer code.

NeilGirdhar: I thought of that, but I think that will cause unintended weight issues.  If the font they chose is not weight=bold, then detecting is a bold is going to produce a different result.  After all, you don't do the reverse of this for light either.

jgehrig:
@NeilGirdhar

I tested the PR, and everything looks good. Thanks!

I have one minor comment about a corner case of in-between `::weight()` values. After a squash, it should be ready for merge.

src/gui/shell.cpp
jgehrig: Sort of...

Let's say the user creates a font of weight 70.

That would fail each check currently in place:
- QFont::Bold == 75
- QFont::DemiBold == 63
- QFont::Light == 25

The existing master behavior would be to mark a font of weight 70 as `:b`, bold. This PR would leave the font unmarked.

With `::bold()` at the end, it checks `70 > QFont::Medium` or `70 > 57`. So the font is at least marked as bold. Not entirely correct, but better than doing nothing and not a regression.

Alternatively, you could properly handle the font ranges. We should do something for in-between values I think

NeilGirdhar: I undertand that.  And then when they select the font with ":b" they wont get the font they created!  How confusing is that?   It's not like you check if weight() < Medium and default to Light.

jgehrig: Sorry for being pedantic, from my testing everything works great...

However, I could see this causing issues for users with obscure configurations in the future :)

jgehrig: I don't follow... Can you elaborate with a specific example?

Here is how I understand the behavior:

**Before:**
weight == Light (25) -> Light
weight > Medium (57) -> Bold
weight <= Medium && weight != Light (25) -> Normal

99 -> Bold (not perfect, but reasonable)

**After:**
weight == Light (25) -> Light
weight == DemiBold (63) -> SemiBold
weight == Bold (75) -> Bold
[None of the above] -> Normal

99 -> Normal (behavior change, regression worse than before)

IMO, the "perfect" ranges are:
```C++
// Light
if (weight <= QFont::Light) {
	fdesc += ":l";
}
// Normal
else if (weight < DemiBold) {
}
// DemiBold
else if (weight < Bold) {
	fdesc += ":sb";
}
// Bold
else {
	fdesc += ":b";
}
```

However, that is rather confusing code. For readability's sake, I you could append `else if (font().bold())` as a reasonable compromise. Yes, that would result in odd behavior for weight range (Medium, DemiBold].

Sorry for being pedantic... Please let me know if I am missing something :)

NeilGirdhar: I understand what you're saying, I just think it should be a bijection.

jgehrig: I suppose we simply disagree :man_shrugging:

Opinion: The behavior is less optimal than what was there before. For example, neovim-qt will now display a regular font for heavy weights such as `Black`. Font weight is graded, so it would not surprise me if this causes issues with `:b` on some systems...

NeilGirdhar: If your ranges interpretation was the right one, then why wouldn't they have coded ranges into Qt?

jgehrig: Font weight is ranged. Why else would QFont::bold() be implemented this way?
``` C++
inline bool QFont::bold() const
{ return weight() > Medium; }
```

Qt just doesn't provide the exact API we need in this application. This is a very specific use case:

1. We are doing odd things: converting QFont to description and back.
2. We don't support every style enumeration.
3. We aren't actually storing the weight directly.

I'm open to solutions... But I think we should handle in-between values. Previously, the behavior made sense given the supported enumerations

Maybe we can add a `:w{num}` for in-between weights? I remember seeing a bug actually requesting this...

Other Issues: #532 #597

NeilGirdhar: Maybe we're spending way too long on something no one is ever going to see?  Would you mind just sending me the code you want me to use, or else checking in the version you like yourself?

jgehrig: Sorry, I'm not trying to be a PITA here :/

You are making a change to the bold logic that is not strictly required. If you want to do that great!

However, you need to prove that your new code will work for all users. I haven't seen this yet...

> Would you mind just sending me the code you want me to use

I've made a few suggestions, see my first comment.

Here is the code in-place:
```C++
QString Shell::fontDesc()
{
	QString fdesc = QString("%1:h%2").arg(fontFamily()).arg(fontSize());
	if (font().weight() == QFont::Bold) {
		fdesc += ":b";
	} else if (font().weight() == QFont::Light) {
		fdesc += ":l";
	} else if (font().weight() == QFont::DemiBold) {
		fdesc += ":sb";
	}
	// Must be checked last: fallback to bold if weight is greater than Medium.
	else if (font().bold()) { 
    		fdesc += :b;
	}
```

If you don't like that, I understand. There is also the range based solution above, or the option to add arbitrary weight `:w` (requested by other users, best solution more work?).

Thank you for your contributions, this is a valuable feature requested by multiple users!

NeilGirdhar: Don't worry, you're not being a pain.  The w solution sounds best.  Will send you a merged changelist this weekend.

NeilGirdhar: Just to confirm, I think the logical thing to do is get rid of "sb" then and keep it simple: b, l, wX for X a number between 0 and 100

jgehrig: Agreed, I like the `:w` solution best too. Thanks!

I don't see any harm in keeping `:sb` around, it seems like a useful option. If you want to remove it, that's fine too.

jgehrig:
Thanks @NeilGirdhar! 👍

Issue #326, opened 2017-08-27
Possibility for scrollbars?


frankie-loves-jesus:
Hi! Any chance nvim-qt (Windows 10) can support a right scrollbar similar to gVim?

Many thanks!

> I know that I can use the mouse wheel, but a scroll bar provides much more feedback about document length and current location of the view related to the complete file.
> IMHO, this should be a priority feature.

Ref. Add GUI scroll bars (Mac OS X): https://github.com/rogual/neovim-dot-app/issues/294

equalsraf:
I assume you were thinking about something similar to the gVim behaviour where two scrollbar layouts are placed at both sides of the window (:h gui-scrollbars).

Yeah this seems possible but it will be some work. We need two things

1. on this repo we need to change the gui shell to have space for the scrollbars. and build a scroll bar layout
2. a vim script (to be part of the gui shim) that keeps track of the windows, their sizes and the scrollbar position and reports this information to the gui. 
3. figure out the correct way to scroll a window using the neovim api

As a side note, I'm not a big fan of the Vim vertical scroll bar setup, because they stop being so convinient when you have more than 2 vertical splits. So if anyone has any other concepts.those would be welcome. This PR https://github.com/neovim/neovim/pull/6373 would likely make that possible.

frankie-loves-jesus:
What I had in mind were vertical scrollbars only, but come to think of it horizontal ones would be friggin' sweet as well.

As for vertical splits I've never felt the need for them when I can just launch another Vim, but I can understand why you terminal-savvy users might want them.

Cheers!

3rdey3:
👍

nash900:
Vertical scroll bar for Neovim is a nice feature. Hope to see this enhancement in future. Thanks.

qwfy:
I think a scrollbar is an essential part of a text editor with a GUI.

Often, you just want to open a file and quickly scroll through it, surely you can do it with keyboard, but sometimes it's just more convenient to use a mouse. Keyboard is good for heavy editing and serious  navigation, but mouse has its use cases.

hasufell:
neovim... 20+ attempts of GUI editors. Half of them abandoned/archived, the others missing basic GUI functionality ;)

facundobatista:
Scroll bars are a "must" in my way of working. 

To explore a file they are mandatory, specially if the text is wider than the window, as moving with the keyboard will make the GUI to reposition according to each line length.

I hope this is implemented soon...

[...]

PR #642, opened 2020-02-06
Issue 326: Add Basic GuiScrollBar Support

jgehrig:
This change adds a primitive GUI `QScrollBar` to interact with the active buffer.

image

The feature can be enabled with:
`:GuiScrollBar 1`

**Design:**
The scrollbar position for the active buffer is computed by attaching to autocmds `CursorMoved`, `CursorMovedI`, and `VimResized`. Whenever one of these events is triggered, we can compute the scrollbar position with the following data:
1. `line('w0')`: Minimum visible line number in active buffer
2. `line('$')`: Current line number of cursor in active buffer
3. `winheight(winnr())`: Size of the active buffer.

These events alone are not enough to keep the scrollbar synchronized. I have also connected the `scroll` and `grid_scroll` events to determine when the buffer scrolls and the cursor does not move.

When a scroll event is triggered from the GUI (user adjusts QScrollBarPosition), `nvim-qt` computes a scroll delta between the last-stored position and the current position. This delta is then used to send an input event `X<C-E>` or `X<C-Y>`, where X is the number of lines to scroll.

I am not particularly fond of the `<C-E>`/`<C-Y>` approach... But I haven't come up with anything better yet. The use of `scroll` and `grid_scroll` also seems a little hack-y.

**Related Issues**:
Issue #326

Scrolling seems like a basic feature that has been missing for too long :)

### DIFF
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim
@@ -210,3 +210,20 @@ anoremenu <script> Gui.Treeview.Toggle :call <SID>TreeViewShowToggle()
 function GuiShowContextMenu() range
 	call rpcnotify(0, 'Gui', 'ShowContextMenu')
 endfunction
+
+" Notify Gui of any cursor movement, used by GuiScrollBar
+function s:GuiCursorMovedUpdateScrollBar()
+	let l:minLineVisible = line('w0')
+	let l:bufferSize = line('$')
+	let l:windowHeight = winheight(winnr())
+	call rpcnotify(0, 'Gui', 'CursorMovedUpdateScrollBar', l:minLineVisible, l:bufferSize, l:windowHeight)
+endfunction
+
+autocmd CursorMoved,CursorMovedI,VimResized * call <SID>GuiCursorMovedUpdateScrollBar()
+
+" Show/Hide the Gui ScrollBar
+function! s:GuiScrollBar(enable) abort
+	call rpcnotify(0, 'Gui', 'SetScrollBarVisible', a:enable)
+endfunction
+
+command! -nargs=1 GuiScrollBar call s:GuiScrollBar(<args>)
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -3,6 +3,7 @@
 #include <QCloseEvent>
 #include <QLayout>
 #include <QSettings>
+#include <QSignalBlocker>
 #include <QToolBar>
 
 namespace NeovimQt {
@@ -68,10 +69,25 @@ void MainWindow::init(NeovimConnector *c)
 	m_tree = new TreeView(c);
 	m_shell = new Shell(c, m_shell_options);
 
+	// GuiScrollBar
+	m_rightScrollBar = new QScrollBar{};
+	m_rightScrollBar->setVisible(false);
+	m_rightScrollBar->setMinimum(1);
+
+	// ShellWidget + GuiScrollBar Layout
+	// QSplitter does not allow layouts directly: QWidget { HLayout { ShellWidget, QScrollBar } }
+	QWidget* shellScrollable{ new QWidget() };
+	QHBoxLayout* layout{ new QHBoxLayout() };
+	layout->setSpacing(0);
+	layout->setMargin(0);
+	layout->addWidget(m_shell);
+	layout->addWidget(m_rightScrollBar);
+	shellScrollable->setLayout(layout);
+
 	m_window = new QSplitter();
 	m_window->addWidget(m_tree);
 	m_tree->hide();
-	m_window->addWidget(m_shell);
+	m_window->addWidget(shellScrollable);
 
 	m_stack.insertWidget(1, m_window);
 	m_stack.setCurrentIndex(1);
@@ -112,6 +128,17 @@ void MainWindow::init(NeovimConnector *c)
 			this, &MainWindow::neovimSendPaste);
 	connect(m_actSelectAll, &QAction::triggered,
 			this, &MainWindow::neovimSendSelectAll);
+
+	// GuiScrollBar Signal/Slot Connections
+	connect(m_shell, &Shell::setGuiScrollBarVisible,
+			this, &MainWindow::setGuiScrollBarVisible);
+	connect(m_shell, &Shell::neovimCursorMovedUpdateScrollBar,
+			this, &MainWindow::neovimCursorMovedUpdateScrollBar);
+	connect(m_shell, &Shell::neovimScrollEvent,
+			this, &MainWindow::neovimScrollEvent);
+	connect(m_rightScrollBar, &QScrollBar::valueChanged,
+			m_shell, &Shell::handleScrollBarChanged);
+
 	m_shell->setFocus(Qt::OtherFocusReason);
 
 	if (m_nvim->errorCause()) {
@@ -172,23 +199,7 @@ void MainWindow::neovimSetTitle(const QString &title)
 
 void MainWindow::neovimWidgetResized()
 {
-	// Neovim finished resizing, resize it back to the actual
-	// widget size - this avoids situations when neovim wants a size that
-	// exceeds the available widget size i.e. the GUI tells neovim its
-	// size, not the other way around.
-	if (isMaximized() || isFullScreen()) {
-		QSize size = geometry().size();
-        if (m_tabline_bar->isVisible()) {
-            size.setHeight(size.height() - m_tabline_bar->geometry().size().height());
-        }
-		if (m_tree->isVisible()) {
-			size.scale(size.width() - m_tree->geometry().size().width(),
-				size.height(), Qt::IgnoreAspectRatio);
-		}
-		m_shell->resizeNeovim(size);
-	} else {
-		m_shell->resizeNeovim(m_shell->size());
-	}
+	m_shell->resizeNeovim(m_shell->size());
 }
 
 void MainWindow::neovimMaximized(bool set)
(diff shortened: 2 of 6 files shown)


src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -213,3 +213,17 @@ anoremenu <script> Gui.Treeview.Toggle :call <SID>TreeViewShowToggle()
...
+" Notify Gui of any cursor movement, used by GuiScrollBar
+function GuiCursorMovedUpdateScrollBar()
+	call rpcnotify(0, 'Gui', 'CursorMovedUpdateScrollBar', line('w0'), line('$'), winheight(winnr()))
+endfunction
+
+autocmd CursorMoved,CursorMovedI,VimResized * call GuiCursorMovedUpdateScrollBar()
+
+" Show/Hide the Gui ScrollBar
+function! s:GuiScrollBar(enable) abort
+	call rpcnotify(0, 'Gui', 'SetScrollBarVisible', a:enable)
```
equalsraf: :+1: the shim should enable this. If anything because the shim might not be there.

Still not sure if there is a point to check GUI support with something like the existing `ui_has_clipboard()`. Might be worth it for user experience.

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -213,3 +213,17 @@ anoremenu <script> Gui.Treeview.Toggle :call <SID>TreeViewShowToggle()
...
+endfunction
+
+command! -nargs=1 GuiScrollBar call s:GuiScrollBar(<args>)
```
equalsraf: This would be a call in ginit.vim right?

Update the docs with GuiScrollBar

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -213,3 +213,17 @@ anoremenu <script> Gui.Treeview.Toggle :call <SID>TreeViewShowToggle()
 function GuiShowContextMenu() range
 	call rpcnotify(0, 'Gui', 'ShowContextMenu')
 endfunction
+
+" Notify Gui of any cursor movement, used by GuiScrollBar
+function GuiCursorMovedUpdateScrollBar()
```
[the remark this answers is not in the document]
jgehrig: Yes, no need to expose this beyond  the script.

src/gui/runtime/plugin/nvim_gui_shim.vim
jgehrig: It shouldn't be possible to enable this without the shim, right?

The scrollbar is off by default, and this function is required to enable. I suppose the user could manually call rpcnotify.

src/gui/runtime/plugin/nvim_gui_shim.vim
jgehrig: Yes, this would be off-by-default and enabled from ginit.vim

facundobatista:
Help for testing this? I merged this branch, compiled, and then running it from `build/bin`, like:

```

**Comment by jgehrig** [comment]
It looks like you are missing the `NVIM_QT_RUNTIME_PATH`, which loads the vimscript for commands like `GuiScrollBar`.

I use the following script to run debug builds of `nvim-qt`:
```bash
#!/bin/bash
pushd ~/Development/neovim-qt/build/bin
NVIM_QT_RUNTIME_PATH=~/Development/neovim-qt/src/gui/runtime ./nvim-qt TestFile.txt
popd
```

I'm aware of some synchronization issues when the scrollbar is adjusted (as opposed to moving the scroll position via neovim). However, I have not been able to track down a root cause, or specific repro conditions.

facundobatista:
I set up that env var, is it pointing to the proper dir, right?

```
10:06:17|facundo@blackfx:~/sistema/neovim-qt/build (review_642)$ echo $NVIM_QT_RUNTIME_PATH
/home/facundo/sistema/neovim-qt/src/gui/runtime
10:06:24|facundo@blackfx:~/sistema/neovim-qt/build (review_642)$ ll $NVIM_QT_RUNTIME_PATH
total 12
drwxrwxr-x 2 facundo facundo 4096 feb 21 09:02 doc
drwxrwxr-x 2 facundo facundo 4096 feb 21 09:02 plugin
-rw-rw-r-- 1 facundo facundo  200 ene  9 20:25 README.md
10:07:34|facundo@blackfx:~/sistema/neovim-qt/build (review_642)$ bin/nvim-qt ~/refresh2.txt
```

It still tells me that it's not an editor command...

facundobatista:
Is there a way to tell neovim-qt to display debug information about what is loaded/available somehow? (just to be able to debug this better) thanks!

jgehrig:
If you are building from source, `-DCMAKE_BUILD_TYPE=Debug` should provide extra information.

You can also add your own `qDebug()` statements. The issues you describe are certainly lack of the shim vimscript.

It looks like you are on Linux/Mac. Have you run the command exactly as follows?
`NVIM_QT_RUNTIME_PATH=~/Development/neovim-qt/src/gui/runtime ./nvim-qt TestFile.txt`

nash900:
Great feature! Will it be included in Neovim v0.5.0 release?

jgehrig:
Thanks @nash900!

I'm not sure yet. There are still some scrollbar synchronization issues that need to be addressed. A read-only scrollbar is currently working, all of the issues I have observed are around adjusting the scrollbar via the Qt Widget.

We might be able to release a read-only scrollbar if there is a time crunch on the 0.5.0 release.

If you are able to build this, testing observations are welcomed! That should make this move faster. :)

facundobatista:
> If you are building from source, `-DCMAKE_BUILD_TYPE=Debug` should provide extra information.

I build it like way, and now when I execute the process I get some more messages. However, when I try to enable scrollbars, nothing extra is shown.


> It looks like you are on Linux/Mac. Have you run the command exactly as follows?
> `NVIM_QT_RUNTIME_PATH=~/Development/neovim-qt/src/gui/runtime ./nvim-qt TestFile.txt`

Of course not. I have the project cloned in other directory. However it's close enough:

    $ ll ~/sistema/neovim-qt/src/gui/runtime
    total 12
    drwxrwxr-x 2 facundo facundo 4096 feb 21 09:02 doc
    drwxrwxr-x 2 facundo facundo 4096 feb 21 09:02 plugin
    -rw-rw-r-- 1 facundo facundo  200 ene  9 20:25 README.md
    $ NVIM_QT_RUNTIME_PATH=~/sistema/neovim-qt/src/gui/runtime bin/nvim-qt ~/refresh2.txt

jgehrig:
Interesting, I'm not sure what is going wrong in your environment with the runtime path.

You can now confirm if the runtime is properly loaded with `--version`:
```
NVIM-QT v0.2.14.0
Build type: Release
Compilation: -Wall -Wextra -Wno-unused-parameter -std=c++11 -Wfatal-errors
Qt Version: 5.13.2
Environment: 
  nvim: nvim
  args: --cmd set termguicolors
  runtime: {RUNTIME PATH SHOULD BE HERE}
...
```

Since you've tried the obvious fixes, I would suggest setting a breakpoint or just print debug with `qDebug()` in `App::getRuntimePath()`. There are two implementations, one generic and the other for MacOS.

jgehrig:
The jumpiness has been fixed for large moves of the scrollbar. The `vim_command` repeat count was ignored due to some invalid `"`s.

There is still one minor issue: a discontinuity when the cursor bounces off the screen edge.

**Steps:**
1. Scroll without causing scrolloff/screen-edge-hit.
2. Click somewhere in the buffer window. Notice the scrollbar does not move.
3. Scroll up or down until the cursor hits an edge. Keep scrolling.
4. Click somewhere in the buffer window.
5. Notice the buffer window does not change, but the scrollbar jumps.

Scroll_Screen_Edge_Jump

jgehrig:
The symptoms above are caused by the order of scroll events. With the cursor at the edge, the scrollbar is moved by two scroll increments.

Neovim sends the cursor move event first, causing the scrollbar to move to the true position. The GUI scroll event is then processed afterwards. The second scroll event causes the incorrect offset.

There needs to be a lockout for when a scroll causes both events to fire.

jgehrig:
This feature does not work with `incsearch`.

Incremental Search

**Upstream Scroll Event Issues:**
neovim/neovim#8854
neovim/neovim#11584
neovim/neovim#11748

jgehrig:
**Progress Update**

Pull Description updated/edited.

I am not aware of any outstanding issues other than `incsearch`. To support incsearch, we will need better upstream API support from neovim. For now, Won't Fix.

This feature should be ready to merge after some testing.

jgehrig:
@equalsraf

Ready for merge?

I am confident in the basic scenarios for this feature, and not aware of any outstanding issues. Since this feature is opt-in and well isolated, let's check-in and get this to users.

Maybe we can turn this on-by-default after some soak time with real users?

equalsraf:
Ack. Thanks.

Shatur95:
@jgehrig, thanks for working on it!
I found a regression. Cursor jumps on wrong position when jumping between splits with different lines (even with GuiScrollBar is disabled)

**Steps to reproduce:**
`nvim-qt -- -u NONE`

```vim
:put=range(0,100)
50gg
:vert h $HOME
<C-w>l
```

**Expected behavior:**
Cursor on line 50 (previous position before jumping to help).

**Current behavior:**
Cursor on line 101.

Issue exists after merging this PR: 333dc5556408d037fe793defe799bd636691a903

jgehrig:
Thanks for the thorough bug report! :+1: 

I created Issue #661 to track the regression. There are likely two parts:
1. Stop GuiScrollBar code from running when the feature is off.
2. Fix the underlying split change issues.

PR #643, opened 2020-02-16
Draw undercurl properly under whitespace ` `

statiolake:
Fixes issue that undercurl is not rendered if it is over whitespace characters.

Before:

broken_undercurl

After:

fixed_undercurl


jgehrig:
Looks good to me!

This will cause conflicts with #623... I would be happy to check this in first and handle the rebase.

equalsraf:
Thanks @statiolake, merged.

PR #644, opened 2020-02-16
Support exit status to differenciate `:q` and `:cq`

statiolake:
In terminal Neovim, we can use `:cq` to indicate that the editor failed to do something.  For example, we can use `:cq` to abort `git commit -v --amend`.  However Neovim-qt always returns 0 whether `:q` or `:cq` is used.  This PR makes the exit status of Neovim-qt to be the same with the inner Neovim's exit status.

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -105,15 +105,15 @@ void App::showUi() noexcept
 	}
 #else
 	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(m_connector.get());
+
+	// delete the main window when closed to emit `destroyed()` signal to
+	// support `:cq` return codes (Pull#644).
+	win->setAttribute(Qt::WA_DeleteOnClose);
 	setQuitOnLastWindowClosed(false);
 
-	// delete the main window when closed to emit `destroyed()` signal.
-	win->setAttribute(Qt::WA_DeleteOnClose);
-
-	QObject::connect(instance(), SIGNAL(openFilesTriggered(QList<QUrl>)),
-		win->shell(), SLOT(openFiles(QList<QUrl>)));
-	QObject::connect(win, SIGNAL(closing(int)), instance(), SLOT(mainWindowClosing(int)));
-	QObject::connect(win, SIGNAL(destroyed()), instance(), SLOT(exitWithStatus()));
+	connect(this, &App::openFilesTriggered, win->shell(), &Shell::openFiles);
+	connect(win, &MainWindow::closing, this, &App::mainWindowClosing);
+	connect(win, &MainWindow::destroyed, this, &App::exitWithStatus);
 
 	// Window geometry should be restored only when the user does not specify
 	// one of the following command line arguments. Argument "maximized" can
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -105,9 +105,15 @@ void App::showUi() noexcept
 	}
 #else
 	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(m_connector.get());
+	setQuitOnLastWindowClosed(false);
+
+	// delete the main window when closed to emit `destroyed()` signal.
+	win->setAttribute(Qt::WA_DeleteOnClose);
 
 	QObject::connect(instance(), SIGNAL(openFilesTriggered(QList<QUrl>)),
 		win->shell(), SLOT(openFiles(QList<QUrl>)));
+	QObject::connect(win, SIGNAL(closing(int)), instance(), SLOT(mainWindowClosing(int)));
+	QObject::connect(win, SIGNAL(destroyed()), instance(), SLOT(exitWithStatus()));
 
 	// Window geometry should be restored only when the user does not specify
 	// one of the following command line arguments. Argument "maximized" can
@@ -340,6 +346,14 @@ void App::showVersionInfo(QCommandLineParser& parser) noexcept
 	out << GetNeovimVersionInfo(nvimExecutable) << QT_ENDL;
 
 	PrintInfo(versionInfo);
+}
+
+void App::mainWindowClosing(int status) {
+	m_exitStatus = status;
+}
+
+void App::exitWithStatus() {
+	exit(m_exitStatus);
 }
 
 } // namespace NeovimQt
--- a/src/gui/mainwindow.h
+++ b/src/gui/mainwindow.h
@@ -90,7 +90,7 @@ private:
 	QAction* m_actPaste{ nullptr };
 	QAction* m_actSelectAll{ nullptr };
 	ScrollBar* m_scrollbar{ nullptr };
-	int m_exitStatus = 0;
+	int m_exitStatus{ 0 };
 
 	// GuiAdaptive Color/Font/Style
 	bool m_isAdaptiveColorEnabled{ false };
--- a/src/gui/mainwindow.h
+++ b/src/gui/mainwindow.h
@@ -4,6 +4,7 @@
 #include <QMainWindow>
 #include <QPalette>
 #include <QSplitter>
+#include <QCloseEvent>
 #include <QStackedWidget>
 #include <QTabBar>
 
@@ -34,6 +35,7 @@ public slots:
 	void delayedShow(DelayedShow type=DelayedShow::Normal);
 signals:
 	void neovimAttached(bool);
+	void closing(int);
 protected:
 	virtual void closeEvent(QCloseEvent *ev) Q_DECL_OVERRIDE;
 	virtual void changeEvent(QEvent *ev) Q_DECL_OVERRIDE;
@@ -44,7 +46,7 @@ private slots:
 	void neovimForeground();
 	void neovimSuspend();
 	void neovimFullScreen(bool);
-	void neovimGuiCloseRequest();
+	void neovimGuiCloseRequest(int);
 	void neovimExited(int status);
 	void neovimError(NeovimConnector::NeovimError);
 	void reconnectNeovim();
@@ -88,6 +90,7 @@ private:
 	QAction* m_actPaste{ nullptr };
 	QAction* m_actSelectAll{ nullptr };
 	ScrollBar* m_scrollbar{ nullptr };
+	int m_exitStatus = 0;
 
 	// GuiAdaptive Color/Font/Style
 	bool m_isAdaptiveColorEnabled{ false };
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -970,15 +970,12 @@ void Shell::handleLineSpace(const QVariant& value) noexcept
 
 void Shell::handleCloseEvent(const QVariantList &args) noexcept
 {
-	if (args.size() == 2 && !args.at(1).canConvert<int>()) {
+	qDebug() << "Neovim requested a GUI close";
+
+	if (args.size() >= 2 && !args.at(1).canConvert<int>()) {
 		qWarning() << "Unexpected exit status for close:" << args.at(1);
 		return;
-	} else if (args.size() >= 3) {
-		qWarning() << "Unexpected arguments for close:" << args;
-		return;
 	}
-
-	qDebug() << "Neovim requested a GUI close";
 
 	// Decide exit status.  If there is an argument, the exit status is
 	// specified by the argument.
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -808,23 +808,8 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			m_mouseHide = variant_not_zero(args.at(1));
 			int val = m_mouseHide ? 1 : 0;
 			m_nvim->api0()->vim_set_var("GuiMousehide", val);
-		} else if (guiEvName == "Close" && (args.size() == 1 || args.size() == 2)) {
-			qDebug() << "Neovim requested a GUI close";
-
-			// Decide which exit status Neovim-qt should exit with.
-			int status;
-			if (args.size() == 1) {
-				// If there are no arguments, then the exit status is zero.
-				status = 0;
-			} if (args.size() == 2) {
-				// Or, if there is an argument, then the exit status is
-				// specified in the argument.
-				bool ok;
-				status = args.at(1).toInt(&ok);
-				if (!ok) status = 1;
-			}
-
-			emit neovimGuiCloseRequest(status);
+		} else if (guiEvName == "Close") {
+			handleCloseEvent(args);
 		} else if (guiEvName == "Option" && args.size() >= 3) {
 			QString option = m_nvim->decode(args.at(1).toByteArray());
 			handleExtGuiOption(option, args.at(2));
@@ -981,6 +966,28 @@ void Shell::handleLineSpace(const QVariant& value) noexcept
 	setLineSpace(linespace);
 	m_nvim->api0()->vim_set_var("GuiLinespace", linespace);
 	resizeNeovim(size());
+}
+
+void Shell::handleCloseEvent(const QVariantList &args) noexcept
+{
+	if (args.size() == 2 && !args.at(1).canConvert<int>()) {
+		qWarning() << "Unexpected exit status for close:" << args.at(1);
+		return;
+	} else if (args.size() >= 3) {
+		qWarning() << "Unexpected arguments for close:" << args;
+		return;
+	}
+
+	qDebug() << "Neovim requested a GUI close";
+
+	// Decide exit status.  If there is an argument, the exit status is
+	// specified by the argument.
+	int status = 0;
+	if (args.size() == 2) {
+		status = args.at(1).toInt();
+	}
+
+	emit neovimGuiCloseRequest(status);
 }
 
 void Shell::handleGridResize(const QVariantList& opargs)
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -968,7 +968,7 @@ void Shell::handleLineSpace(const QVariant& value) noexcept
 	resizeNeovim(size());
 }
 
-void Shell::handleCloseEvent(const QVariantList &args) noexcept
+void Shell::handleCloseEvent(const QVariantList& args) noexcept
 {
 	qDebug() << "Neovim requested a GUI close";
 
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -808,12 +808,22 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			m_mouseHide = variant_not_zero(args.at(1));
 			int val = m_mouseHide ? 1 : 0;
 			m_nvim->api0()->vim_set_var("GuiMousehide", val);
-		} else if (guiEvName == "Close" && args.size() == 2) {
+		} else if (guiEvName == "Close" && (args.size() == 1 || args.size() == 2)) {
 			qDebug() << "Neovim requested a GUI close";
-			// retrieve current exit status
-			bool ok;
-			int status = args.at(1).toInt(&ok);
-			if (!ok) status = 1;
+
+			// Decide which exit status Neovim-qt should exit with.
+			int status;
+			if (args.size() == 1) {
+				// If there are no arguments, then the exit status is zero.
+				status = 0;
+			} if (args.size() == 2) {
+				// Or, if there is an argument, then the exit status is
+				// specified in the argument.
+				bool ok;
+				status = args.at(1).toInt(&ok);
+				if (!ok) status = 1;
+			}
+
 			emit neovimGuiCloseRequest(status);
 		} else if (guiEvName == "Option" && args.size() >= 3) {
 			QString option = m_nvim->decode(args.at(1).toByteArray());
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -808,9 +808,13 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			m_mouseHide = variant_not_zero(args.at(1));
 			int val = m_mouseHide ? 1 : 0;
 			m_nvim->api0()->vim_set_var("GuiMousehide", val);
-		} else if (guiEvName == "Close" && args.size() == 1) {
+		} else if (guiEvName == "Close" && args.size() == 2) {
 			qDebug() << "Neovim requested a GUI close";
-			emit neovimGuiCloseRequest();
+			// retrieve current exit status
+			bool ok;
+			int status = args.at(1).toInt(&ok);
+			if (!ok) status = 1;
+			emit neovimGuiCloseRequest(status);
 		} else if (guiEvName == "Option" && args.size() >= 3) {
 			QString option = m_nvim->decode(args.at(1).toByteArray());
 			handleExtGuiOption(option, args.at(2));
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -979,7 +979,7 @@ void Shell::handleCloseEvent(const QVariantList& args) noexcept
 
 	// Decide exit status.  If there is an argument, the exit status is
 	// specified by the argument.
-	int status = 0;
+	int status{ 0 };
 	if (args.size() == 2) {
 		status = args.at(1).toInt();
 	}
--- a/src/gui/shell.h
+++ b/src/gui/shell.h
@@ -167,6 +167,7 @@ protected:
 	virtual void handleGuiFontFunction(const QVariantList& args);
 	virtual void handleGuiFontWide(const QVariant& value) noexcept;
 	virtual void handleLineSpace(const QVariant& value) noexcept;
+	virtual void handleCloseEvent(const QVariantList &args) noexcept;
 
 	// Modern 'ext_linegrid' Grid UI Events
 	virtual void handleGridResize(const QVariantList& opargs);
--- a/src/gui/shell.h
+++ b/src/gui/shell.h
@@ -90,7 +90,7 @@ signals:
 	void neovimOpacity(double);
 	void neovimSuspend();
 	void neovimFullScreen(bool);
-	void neovimGuiCloseRequest(int);
+	void neovimGuiCloseRequest(int status = 0);
 	/// This signal is emmited if the running neovim version is unsupported by the GUI
 	void neovimIsUnsupported();
 	void neovimExtTablineSet(bool);
--- a/src/gui/shell.h
+++ b/src/gui/shell.h
@@ -90,7 +90,7 @@ signals:
 	void neovimOpacity(double);
 	void neovimSuspend();
 	void neovimFullScreen(bool);
-	void neovimGuiCloseRequest();
+	void neovimGuiCloseRequest(int);
 	/// This signal is emmited if the running neovim version is unsupported by the GUI
 	void neovimIsUnsupported();
 	void neovimExtTablineSet(bool);
--- a/test/tst_shell.cpp
+++ b/test/tst_shell.cpp
@@ -252,11 +252,21 @@ private slots:
 		QCOMPARE(onWindowClosing.takeFirst().at(0).toInt(), event_status);
 
 		// and finally a call to nvim-qt
+		QProcess p;
+		p.setProgram(NVIM_QT_BINARY);
+		QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
+		env.insert("NVIM_QT_RUNTIME_PATH", path_to_src_runtime);
+		p.setProcessEnvironment(env);
 #if defined(Q_OS_WIN32)
-		int actual_exit_status = QProcess::execute(NVIM_QT_BINARY, {NVIM_QT_BINARY, "--",  "-c", command});
+		p.setArguments({NVIM_QT_BINARY, "--",  "-c", command});
 #else
-		int actual_exit_status = QProcess::execute(NVIM_QT_BINARY, {NVIM_QT_BINARY, "--nofork", "--",  "-c", command});
+		p.setArguments({NVIM_QT_BINARY, "--nofork", "--",  "-c", command});
 #endif
+		p.start();
+		p.waitForFinished(-1);
+		QCOMPARE(p.exitStatus(), QProcess::NormalExit);
+		int actual_exit_status = p.exitCode();
+
 		QCOMPARE(actual_exit_status, exit_status);
 	}
 
--- a/test/tst_shell.cpp
+++ b/test/tst_shell.cpp
@@ -211,7 +211,8 @@ private slots:
 		QTest::newRow("Exit with Invalid Code: -1cq")
 			<< -1 << 0 << QByteArray("-1cq");
 
-		// Some exit-status scenarios are platform dependent. Ex) Overflow on Windows
+		// Some exit-status scenarios are platform dependent.
+		// Ex) Overflow on UNIX-like operating systems.
 		AddPlatformSpecificExitCodeCases();
 	}
 
--- a/test/tst_shell.cpp
+++ b/test/tst_shell.cpp
@@ -5,7 +5,9 @@
 #include <QClipboard>
 #include <gui/mainwindow.h>
 #include <msgpackrequest.h>
+
 #include "common.h"
+#include "tst_shell.h"
 
 #if defined(Q_OS_WIN) && defined(USE_STATIC_QT)
 #include <QtPlugin>
@@ -14,7 +16,7 @@ Q_IMPORT_PLUGIN (QWindowsIntegrationPlugin);
 
 namespace NeovimQt {
 
-class Test: public QObject
+class Test : public QObject
 {
 	Q_OBJECT
 private slots:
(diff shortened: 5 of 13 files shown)


src/gui/shell.cpp
```diff
@@ -725,9 +725,13 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			m_mouseHide = variant_not_zero(args.at(1));
 			int val = m_mouseHide ? 1 : 0;
 			m_nvim->api0()->vim_set_var("GuiMousehide", val);
-		} else if (guiEvName == "Close" && args.size() == 1) {
+		} else if (guiEvName == "Close" && args.size() == 2) {
```
equalsraf: Allow for size 1 or 2 to remain compatible with older versions i.e. return 0 if the argument is not present.

src/gui/shell.h
```diff
@@ -61,7 +61,7 @@ class Shell: public ShellWidget
 	void neovimMaximized(bool);
 	void neovimSuspend();
 	void neovimFullScreen(bool);
-	void neovimGuiCloseRequest();
```
equalsraf: like wise on the C++ side make the exit code an optional argument with default 0.

src/gui/app.cpp
```diff
@@ -116,10 +116,16 @@ void App::showUi() noexcept
 		win->show();
 	}
 #else
+	setQuitOnLastWindowClosed(false);
```
equalsraf: any particular reason for this change?

src/gui/shell.cpp
```diff
@@ -725,9 +725,23 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			m_mouseHide = variant_not_zero(args.at(1));
 			int val = m_mouseHide ? 1 : 0;
 			m_nvim->api0()->vim_set_var("GuiMousehide", val);
-		} else if (guiEvName == "Close" && args.size() == 1) {
+		} else if (guiEvName == "Close" && (args.size() == 1 || args.size() == 2)) {
```
jgehrig: This has grown in size enough that it should probably be its own function. Keeping the size of these large switch statements down helps read-ability :)

You can take a look at something like `void Shell::handleGridLine(const QVariantList& opargs)` as an example.

Ex)
```C++
void Shell::handleGuiCloseEvent(const QVariantList& args) noexcept
```

src/gui/shell.cpp
```diff
@@ -725,9 +725,23 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
...
+			// Decide which exit status Neovim-qt should exit with.
+			int status;
+			if (args.size() == 1) {
+				// If there are no arguments, then the exit status is zero.
+				status = 0;
+			} if (args.size() == 2) {
+				// Or, if there is an argument, then the exit status is
+				// specified in the argument.
+				bool ok;
+				status = args.at(1).toInt(&ok);
```
jgehrig: It might be a good idea to perform some basic command format checking before parsing or `toInt` conversion?

Ex)
```C++
if (opargs.size() < 2
		// || !opargs.at(0).canConvert<???>() <- May or may not be necessary?
		|| !opargs.at(1).canConvert<int>() {
		qWarning() << "Unexpected arguments {????}" << opargs;
		return;
	}
```

Not strictly necessary, since you check `ok`...

src/gui/shell.cpp
```diff
@@ -725,9 +725,23 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			m_mouseHide = variant_not_zero(args.at(1));
 			int val = m_mouseHide ? 1 : 0;
 			m_nvim->api0()->vim_set_var("GuiMousehide", val);
-		} else if (guiEvName == "Close" && args.size() == 1) {
+		} else if (guiEvName == "Close" && (args.size() == 1 || args.size() == 2)) {
 			qDebug() << "Neovim requested a GUI close";
-			emit neovimGuiCloseRequest();
+
+			// Decide which exit status Neovim-qt should exit with.
+			int status;
```
jgehrig: Provide a default value, even something like `{}` is fine.

This helps prevent undefined behavior bugs.

src/gui/shell.h
statiolake: I fixed the mentioned issue by passing 0 to `neovimGuiCloseRequest` if there are no argument with the Close event.  Is this still needed?  (I can easily add this but will it be used from somewhere...?)

src/gui/shell.cpp
statiolake: OK, It seems better for me too.

src/gui/shell.h
jgehrig: I read this comment as changing the function signature to have a default parameter with value 0:
```C++
void neovimGuiCloseRequest(int status = 0);
```

No strong opinion, just my interpretation :)

src/gui/shell.cpp
statiolake: Now it emits warning message if `ok` is `false`.  If I should use `canConvert()` or something, I'll fix it.

src/gui/shell.h
statiolake: I also interpreted as you said, but I thought there would be no place calling `neovimGuiCloseRequest` with no argument.

Just curious.  I'm OK if it's about readability --- it's good to indicate that an argument of Close event is optional (I fixed it to have a default value 0).

statiolake: I noticed I've added the default argument to the wrong place (`mainwindow.h` which defines a slot connected to this signal).  Should I add it to both or signal only?

jgehrig: It's just a convention thing... There will be different opinions depending on who you ask. I don't have a strong opinion here. Either way is fine with me.

In this case, a default parameter allows you to make this change and keep the old calling pattern valid. This might be useful if you want to exit Neovim, but don't care about status code (default to OK).

jgehrig: Just for clarification... Sorry if you already know this!

Technically "optional" isn't the correct term for this language feature.

C++ supports "default arguments", which are different than an optional/keyword argument in some languages like Python. Default arguments simply have a default value when one is not provided.

For example, a call to `neovimGuiCloseRequest()` is converted by the compiler to `neovimGuiCloseRequest(0)` at compile-time.

statiolake: Thank you for your clarification!  Now it's much clearer for me.  I removed a default value from the slot's argument, and added one to the signal's argument.

You are true, "optional" was my bad wording.  I know it's just a compile time thing.   Sorry if I misunderstood as I'm new to Qt, but in my understanding basically signals and slots are used by connecting them each other (`connect(source, SIGNAL(neovimGuiCloseRequest(int)), receiver, SLOT(neovimGuiCloseRequest(int)))`).  From that point, the signature `neovimGuiCloseRequest(int)` is the same regardless of the default value's presence (as it's not a optional argument but just a default value), so at first I got confused.  However if there's default value in the signal's definition, it's certainly useful in the situation you described. I understand it's better to add default value to the signal's argument.

I'm sorry to take up your time.  Thank you for your detailed explanation!

equalsraf:
Looking good. I'm being pedantic, but maybe add a test to prevent regressions?
I'm wandering what is the best way to test this:

```
# this should return 0, :3cq would return 3
nvim-qt -- -c ':q'
```

There is no natural way to test it right now, maybe the only way we can test is check the signal emitted after cq

jgehrig:
```bash
# this should return 0, :3cq would return 3
nvim-qt -- -c ':q'
```

Would this work? You need a `--nofork` right?

I would expect Neovim to spawn a background process and immediately exit with `0` regardless of command. I'm curious so I can test...

Maybe the test could be written as a bash/batch script:
https://stackoverflow.com/questions/25627336/integrate-bash-test-scripts-in-cmake

src/gui/shell.h
jgehrig: I use the word "optional" for this C++ feature too sometimes... No worries.

I think having the default parameter on the signal is the correct solution.

No problem, I'm glad to hear you're learning something new! :)

src/gui/shell.cpp
```diff
@@ -867,6 +866,38 @@ void Shell::handleGuiFontFunction(const QVariantList& args)
 	setGuiFont(fdesc, force, true /*setOption*/);
 }
 
+void Shell::handleCloseEvent(const QVariantList &args)
+{
+	qDebug() << "Neovim requested a GUI close";
+
+	// Decide which exit status Neovim-qt should exit with.
+	int status{};
+	if (args.size() == 1) {
```
jgehrig: You could remove this block entirely.

Just initialize status to 0: `int status{ 0 };`

src/gui/shell.cpp
```diff
@@ -867,6 +866,38 @@ void Shell::handleGuiFontFunction(const QVariantList& args)
...
+		// specified in the argument.
+		bool ok{};
+		status = arg.toInt(&ok);
+		if (!ok) {
+			// If there is some conversion error, emit warning and ignore this
+			// event.  (Though it cannot prevent Neovim from closing.)
+			qWarning() << "Unexpected argument for a exit status: " << arg;
+			return;
+		}
+	} else {
```
jgehrig: nitpick/optional: Generally having failures like this early on streamlines the code.

Ex) check for bad command format at the very beginning and exit on failure. This also prevents the need to check `ok`.

src/gui/app.cpp
```diff
@@ -116,10 +116,16 @@ void App::showUi() noexcept
...
 	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(m_connector.get(), opts);
 
-	QObject::connect(instance(), SIGNAL(openFilesTriggered(const QList<QUrl>)),
+	// delete the main window when closed to emit `destroyed()` signal.
```
jgehrig: nitpick/optional: You can clean up the diff by keeping the line below on its own at the top. Preserves better VCS history.

`QObject::connect(instance(), SIGNAL(openFilesTriggered(const QList<QUrl>)),`

src/gui/shell.cpp
jgehrig: Slight preference for the format in the comment above. That pattern is used throughout `shell.cpp`, and keeping consistency makes reading the code easier.

Somewhat pedantic/nitpicky... Feel free to ignore!

jgehrig:
Looks good, Thanks!

Can you provide a quick summary of how you are using/testing this (a command or two is fine). I only ask so I can verify the behavior on my machine. :)

statiolake:
Actually I'm using only Windows and I don't know if it works on Linux (or other platforms).  I don't know how forking mechanism (to run in background) works on Linux, either.  So I have no idea how to test this feature except on Windows.

In Windows, it's a bit strange but `neovim-qt` runs in background if it is called directly from `cmd.exe`, and runs in foreground (waits for a process exit) if it is called from a batch file (`.bat`) or another process (like `git`).  Therefore you can test this feature by using a simple batch file like below:

```bat
nvim-qt.exe %*
echo %ERRORLEVEL%
``` 

Assume this batch file is saved as `vf.bat` in some directory contained in `%PATH%` environment variable.  If you run this batch file from `cmd.exe`, like `vf -- -c ":q"`, it launches neovim-qt and waits until the process exits, and the exit status (in this case `0`) will be printed after a process exit.  If you run this batch file using `vf -- -c ":3cq"`, then the printed exit status will be `3`.

jgehrig:
Thanks for the description!

It is good to know this hasn't been tested on Linux/Mac yet. Not a problem, I can handle that...

src/gui/app.cpp
```diff
@@ -116,10 +116,16 @@ void App::showUi() noexcept
...
 #else
+	setQuitOnLastWindowClosed(false);
 	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(m_connector.get(), opts);
 
+	// delete the main window when closed to emit `destroyed()` signal.
+	win->setAttribute(Qt::WA_DeleteOnClose);
+
 	QObject::connect(instance(), SIGNAL(openFilesTriggered(const QList<QUrl>)),
 		win->shell(), SLOT(openFiles(const QList<QUrl>)));
+	QObject::connect(win, SIGNAL(closing(int)), instance(), SLOT(mainWindowClosing(int)));
```
jgehrig: Sorry I didn't notice this earlier; Qt deprecated this syntax.

We should use the following instead:
``` C++
	// Pull#644: Add support for `:cq` return codes. 
	win->setAttribute(Qt::WA_DeleteOnClose);
	setQuitOnLastWindowClosed(false);

	connect(win, &MainWindow::closing, this, &App::mainWindowClosing);
	connect(win, &MainWindow::destroyed, this, &App::exitWithStatus);
```

src/gui/shell.cpp
```diff
@@ -867,6 +866,28 @@ void Shell::handleGuiFontFunction(const QVariantList& args)
...
+{
+	if (args.size() == 2 && !args.at(1).canConvert<int>()) {
+		qWarning() << "Unexpected exit status for close:" << args.at(1);
+		return;
+	} else if (args.size() >= 3) {
+		qWarning() << "Unexpected arguments for close:" << args;
+		return;
+	}
+
+	qDebug() << "Neovim requested a GUI close";
```
jgehrig: Let's get rid of the `>=3` case, and change to `>=2`. There could be additional parameters in the future...

``` C++
qDebug() << "Neovim requested a GUI close";

if (args.size() >= 2 && !args.at(1).canConvert<int>()) {
	qWarning() << "Unexpected exit status for close:" << args.at(1);
	return;
}
```

Might as well add the qDebug at the very top. The request was made even if malformed.

jgehrig:
@statiolake 

Test on MacOS/Linux... Everything looks good!

For Linux:
1. The default is fork, this always returns exit code 0
2. When arg --nofork is added, the exit code from `:cq` is returned.

For MacOS:
1. The default is no fork, and the exit code from`:cq` is returned.

I added two minor comments. After a squash, this looks ready to me! :+1:

equalsraf:
> I would expect Neovim to spawn a background process and immediately exit with 0 regardless of command. I'm curious so I can test...

Got some tests going in https://github.com/equalsraf/neovim-qt/commit/453b783af9de3a5c65425c84b2de419a557344d4 can you try cherry picking that commit into this?

Like you said --nofork was needed.

I only came across two corner cases

- `256cq` causes a return code 0, but this is not unexpected.since exit codes in linux are limited to 8 bits so we can just pass the value along. I put in a test case for 256cq but might have to remove it in windows.
- -`1cq`  runs for me but causes a **0** exit status, however when calling this using `-c` it fails with an invalid range error. `-2cq` always fails.

statiolake:
I cherry-picked 65add2c as-is but it caused an error in CI...

jgehrig:
@statiolake 

I think it is complaining that you cannot compare a `QVariant` and `int`.

Try adding a `.toInt()` to resolve the compiler error?

```C++
QVERIFY(SPYWAIT(onClose));
QCOMPARE(onClose.takeFirst().at(0).toInt(), event_status);

QVERIFY(SPYWAIT(onWindowClosing));
QCOMPARE(onWindowClosing.takeFirst().at(0).toInt(), event_status);
```

statiolake:
I can't figure out what is the main cause of CI failure... I don't even have Linux machine so I can't run tests on my side. What does "Error writing to process" mean?

jgehrig:
@statiolake

It looks like the process hung. The test is failing due to timeout.

On my Linux machine, ShellWidget shows a dialog box with the failure code. I think `tst_shell` hangs until the dialog box is closed.

image

statiolake:
@jgehrig Thank you for testing. Is the dialog box shown when you execute `:cq` in the opened Neovim-qt? Or is it shown only when command is executed by the command line arguments, e.g. `Neovim-qt --nofork -- -c cq`? If former is true, some fix is needed as the dialog should not be displayed in such a case. If latter is true, there might be two options:
1. Close the dialog somehow in the test
2. Fix the code so that the dialog won't be shown when the status code comes from `cq` command

jgehrig:
@statiolake

No problem!

Correction, it looks like that is `nvim-qt` itself and not a dialog box. When run manually, the issue triggers from `nvim-qt --nofork -- -c cq3`. Interestingly, I don't see the error code propagate in that case. The value is always 1; even with the command above I get `Neovim exited with status code (1)`.

Here is the output of a run with some additional `qDebug()` output:
**Modifications:**
```C++
// and finally a call to nvim-qt
qDebug() << QStringList{NVIM_QT_BINARY, "--nofork", "--",  "-c", command}; // ADDED LINE
#if defined(Q_OS_WIN32)
int actual_exit_status = QProcess::execute(NVIM_QT_BINARY, {NVIM_QT_BINARY, "--",  "-c", command});
#else
int actual_exit_status = QProcess::execute(NVIM_QT_BINARY, {NVIM_QT_BINARY, "--nofork", "--",  "-c", command});
#endif
qDebug() << "actual_exit_status returned!!!"; // ADDED LINE
QCOMPARE(actual_exit_status, exit_status);
```
**Output:**
```
6: QDEBUG : NeovimQt::Test::CloseEvent(cq) ("/home/jgehrig/Development/neovim-qt/build/bin/nvim-qt", "--nofork", "--", "-c", "cq")

// The test hangs here, and I had to manually close nvim-qt for it to continue.

6: QDEBUG : NeovimQt::Test::CloseEvent(cq) actual_exit_status returned!!!
6: FAIL!  : NeovimQt::Test::CloseEvent(cq) Compared values are not the same
6:    Actual   (actual_exit_status): 0
6:    Expected (exit_status)       : 1
6:    Loc: [../test/tst_shell.cpp(262)]
6: QWARN  : NeovimQt::Test::CloseEvent(2cq) qt.svg: Cannot open file ':/neovim.svg', because: No such file or directory
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim process error  "Error writing to process"
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim API version compatible with 0 supported 6
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim requested a GUI close
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim process error  "Error writing to process"
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) ("/home/jgehrig/Development/neovim-qt/build/bin/nvim-qt", "--nofork", "--", "-c", "2cq")
6: Neovim API version compatible with 0 supported 6

// There was another hang here
```

equalsraf:
Does this mean we need something like this?

```diff
diff --git a/src/gui/mainwindow.cpp b/src/gui/mainwindow.cpp
index 189292c..6439141 100644
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -160,6 +160,10 @@ void MainWindow::neovimExited(int status)
                m_errorWidget->setText(m_nvim->errorString());
                m_errorWidget->showReconnect(m_nvim->canReconnect());
                m_stack.setCurrentIndex(0);
+       } else if (status == m_exitStatus) {
+               // Neovim exited with an error code, but a matching exit code
+               // was set via Gui Close request, treat this as a normal close
+               close();
        } else if (status != 0) {
                m_errorWidget->setText(QString("Neovim exited with status code (%1)").arg(status));
                m_errorWidget->showReconnect(m_nvim->canReconnect());


```

Based on that popup it seems like we ran into a timing issue where nvim exits before nvim-qt.

I have pushed a branch to run the tests (still running)
https://github.com/equalsraf/neovim-qt/tree/exitstatus

jgehrig:
I was about to write that I thought this was a timing-related failure of the shim to load `autocmd VimLeave`, and trigger `GuiClose()`... Actually, the issue is simpler.

This is runtime path failure! :smile: 

I added `qDebug()` statements to all of the exit point signals/slots.

**Here is a trace of a failing `tst_shell`:**
```
6: "App::mainWindowClosing(0)"
6: "App::exitWithStatus(0)"
6: FAIL!  : NeovimQt::Test::CloseEvent(cq) Compared values are not the same
6:    Actual   (actual_exit_status): 0
6:    Expected (exit_status)       : 1
6:    Loc: [../test/tst_shell.cpp(260)]
6: QWARN  : NeovimQt::Test::CloseEvent(2cq) qt.svg: Cannot open file ':/neovim.svg', because: No such file or directory
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim process error  "Error writing to process"
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) "Shell::neovimExited(1)"
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) "MainWindow::neovimExited(1)"
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim API version compatible with 0 supported 6
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Shell::handleCloseEvent (QVariant(QByteArray, "Close"), QVariant(qulonglong, 2))
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim requested a GUI close
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) "MainWindow::neovimGuiCloseRequest(2)"
6: QDEBUG : NeovimQt::Test::CloseEvent(2cq) Neovim process error  "Error writing to process"
6: Neovim API version compatible with 0 supported 6
6: "Shell::neovimExited(2)"
6: "MainWindow::neovimExited(2)"
```

**Here is a trace of manually running `nvim-qt --nofork` `:cq2` on Linux:**
```
Neovim API version compatible with 0 supported 7
Shell::handleCloseEvent (QVariant(QByteArray, "Close"), QVariant(qulonglong, 2))
Neovim requested a GUI close
"MainWindow::neovimGuiCloseRequest(2)"
"App::mainWindowClosing(2)"
"App::exitWithStatus(2)"
Exit Code: 2
```

**Trace with the following added to `Test::CloseEvent()`:**
```C++
qDebug() << "RUNTIME PATH:" << QProcess::execute(NVIM_QT_BINARY, {"--version"});
```

```
6: NVIM-QT v0.2.16.9999
6: Build type: Debug
6: Compilation: -Wall -Wextra -Wno-unused-parameter -std=c++11 -Wfatal-errors
6: Qt Version: 5.13.2
6: Environment: 
6:   nvim: nvim
6:   args: --cmd set termguicolors
6:   runtime: 
6: 
6: NVIM v0.4.3
6: Build type: Release
6: LuaJIT 2.0.5
6: Compilation: /usr/bin/x86_64-pc-linux-gnu-gcc -march=native -O2 -pipe -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/tmp/portage/app-editors/neovim-0.4.3/work/neovim-0.4.3_build/config -I/tmp/portage/app-editors/neovim-0.4.3/work/neovim-0.4.3/src -I/usr/include -I/tmp/portage/app-editors/neovim-0.4.3/work/neovim-0.4.3_build/src/nvim/auto -I/tmp/portage/app-editors/neovim-0.4.3/work/neovim-0.4.3_build/include
6: Compiled by portage@jgehrig-dev
6: 
6: Features: +acl +iconv +tui
6: See ":help feature-compile"
6: 
6:    system vimrc file: "/etc/vim/sysinit.vim"
6:   fall-back for $VIM: "/usr/share/nvim"
6: 
6: Run :checkhealth for more info
6: 
6: QDEBUG : NeovimQt::Test::CloseEvent(cq) RUNTIME PATH: 0
```
Notice `runtime:` is empty.

I can confirm the behavior by running `nvim-qt` without setting `NVIM_QT_RUNTIME_PATH`.

Maybe we can set `NVIM_QT_RUNTIME_PATH` in the `QProcess` call?

equalsraf:
> This is runtime path failure! smile

Uuuuf, I was starting to get worried here. But there are definitely some timing issues lurking.

> Maybe we can set NVIM_QT_RUNTIME_PATH in the QProcess call?

One more try then (https://github.com/equalsraf/neovim-qt/tree/tb-exitstatus)

```diff
diff --git a/test/tst_shell.cpp b/test/tst_shell.cpp
index 168d60c..295f1ba 100644
--- a/test/tst_shell.cpp
+++ b/test/tst_shell.cpp
@@ -252,11 +252,21 @@ private slots:
                QCOMPARE(onWindowClosing.takeFirst().at(0).toInt(), event_status);
 
                // and finally a call to nvim-qt
+               QProcess p;
+               p.setProgram(NVIM_QT_BINARY);
+               QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
+               env.insert("NVIM_QT_RUNTIME_PATH", path_to_src_runtime);
+               p.setProcessEnvironment(env);
 #if defined(Q_OS_WIN32)
-               int actual_exit_status = QProcess::execute(NVIM_QT_BINARY, {NVIM_QT_BINARY, "--",  "-c", command});
+               p.setArguments({NVIM_QT_BINARY, "--",  "-c", command});
 #else
-               int actual_exit_status = QProcess::execute(NVIM_QT_BINARY, {NVIM_QT_BINARY, "--nofork", "--",  "-c", command});
+               p.setArguments({NVIM_QT_BINARY, "--nofork", "--",  "-c", command});
 #endif
+               p.start();
+               p.waitForFinished(-1);
+               QCOMPARE(p.exitStatus(), QProcess::NormalExit);
+               int actual_exit_status = p.exitCode();
+
                QCOMPARE(actual_exit_status, exit_status);
        }

```

equalsraf:
Ha https://circleci.com/gh/equalsraf/neovim-qt/246 you were right, nice going @jgehrig.

jgehrig:
Maybe this is crazy idea...

What if we get rid of `GuiClose()` completely? Can we handle all close paths through `Shell::neovimExited` instead?

This would remove the shim dependence and simplify the code... It would also fix a scenario I personally find annoying where the file already open error dialog "Abort" option prompts for "Retry".

```C++
/** The Neovim process has exited */
void Shell::neovimExited(int status)
{
	setAttached(false);
	if (m_nvim->errorCause() == NeovimConnector::NoError) {
		emit neovimGuiCloseRequest(status);
	}
}
```

equalsraf:
> Maybe this is a crazy idea...

Not crazy, you are just missing a couple of use cases.

It is is possible for nvim-qt to connect to a neovim instance without launching its process (via tcp or named sockets). In that case neovimExited never gets called, and that is the reason GuiClosed exists.

You do raise a good question though. Should we ignore GuiClosed when nvim is a local process spawned by the connector? There is one case that should still trigger the error popup - when nvim crashes - and i think QProcess::exitStatus signals that.

jgehrig:
I figured there was a corner case I wasn't considering... We should just set `NVIM_QT_RUNTIME_PATH` for this Pull Request.

Here is an example without `GuiClose` b8005034c44a7c13cef5dd34a39290701a01d340 if anyone wants to experiment.

> You do raise a good question though. Should we ignore GuiClosed when nvim is a local process spawned by the connector? There is one case that should still trigger the error popup - when nvim crashes - and i think QProcess::exitStatus signals that.

Yeah, it would be nice to have the close handler work regardless of the runtime shim.

I suspect `QProcess::exitStatus` and NeovimConnector's `m_error` would be enough to handle any process spawn failures.

statiolake:
Due to a lack of Neovim's official windows build, CI is still failing....

statiolake:
@jgehrig @equalsraf Sorry for pinging you...
How do you feel about this PR? If I should take a different approach (such as removing GuiClose), I'll try it when I have time.

src/gui/mainwindow.h
```diff
@@ -88,6 +90,7 @@ private slots:
 	QAction* m_actPaste{ nullptr };
 	QAction* m_actSelectAll{ nullptr };
 	ScrollBar* m_scrollbar{ nullptr };
+	int m_exitStatus = 0;
```
jgehrig, line 93: nitpick: Initialize with `{}` when possible. The compiler will do more aggressively type-checking for that syntax.

Elsewhere too.

src/gui/shell.cpp
```diff
@@ -969,6 +968,25 @@ void Shell::handleLineSpace(const QVariant& value) noexcept
 	resizeNeovim(size());
 }
 
+void Shell::handleCloseEvent(const QVariantList &args) noexcept
```
jgehrig, line 971: nitpick: `QVariantList& args`

Sorry for the nitpicks. Only mentioned since we're changing other code anyways :)

test/tst_shell.cpp
```diff
@@ -188,6 +188,88 @@ private slots:
...
+		// Overflow except Windows, this generates an exit code 0, but the
+		// original count is passed to us
+		// Commented out as this test doesn't pass on Windows
+//		QTest::newRow("256cq")
+//			<< 256 << 0 << QByteArray("256cq");
+		// a bit of corner case, but nvim exits with
+		// status 0 and provides us count 0 - calling it with nvim -c
+		// actually causes an error with an invalid range, so it can not run
+//		QTest::newRow("-1cq")
+//			<< 0 << QByteArray("-1cq");
```
jgehrig, line 216: This commented-out segment should be cleaned up before check-in.

Two thoughts:
1) Let's avoid checking in commented-out blocks of code.
2) If this isn't too hard: let's explicitly test the behavior differences between Windows/Linux/MacOS. You can use tst_input* as a pattern to write platform specific code. If this is to much trouble, let's write a clear comment about the behavior differences. Please avoid `#ifdef` ;)

We should have a single clear comment about the differences between Windows vs Linux/Mac.

test/tst_shell.cpp
```diff
@@ -188,6 +188,88 @@ private slots:
...
+		QVERIFY(SPYWAIT(onWindowClosing));
+		QCOMPARE(onWindowClosing.takeFirst().at(0).toInt(), event_status);
+
+		// and finally a call to nvim-qt
+		QProcess p;
+		p.setProgram(NVIM_QT_BINARY);
+		QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
+		env.insert("NVIM_QT_RUNTIME_PATH", path_to_src_runtime);
+		p.setProcessEnvironment(env);
+#if defined(Q_OS_WIN32)
```
jgehrig, line 260: Let's avoid `#ifdef` (if possible) in new code.

statiolake:
@jgehrig I'm not familiar with Qt and I don't know how to add platform-specific tests.

According to your instructions I referenced `tst_input_*`. They seem to be declared at `test/CMakeLists.txt`:

```cmake
# Platform Specific Input Tests
add_xtest(tst_input_mac
	${CMAKE_SOURCE_DIR}/src/gui/input.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input_mac.cpp)
add_xtest(tst_input_unix
	${CMAKE_SOURCE_DIR}/src/gui/input.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input_unix.cpp)
add_xtest(tst_input_win32
	${CMAKE_SOURCE_DIR}/src/gui/input.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input_win32.cpp)

# Platform Agnostic Input Tests
add_shared_test(tst_input_common_mac
	tst_input_common.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input_mac.cpp)
add_shared_test(tst_input_common_unix
	tst_input_common.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input_unix.cpp)
add_shared_test(tst_input_common_win32
	tst_input_common.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input.cpp
	${CMAKE_SOURCE_DIR}/src/gui/input_win32.cpp)
```

But I can't figure out what these lines mean. Especially, where can I state that a test should be run only on a specific platform? Is it automatically estimated by the postfix `_mac`, `_win32` and `_unix`?

I would appreciate it if you could give me a bit more detailed explanation.

jgehrig:
@statiolake 

Do you mind if I quickly write the cross-platform handling code and push it to your PR? That might be easiest...

---
**Detailed explanation:**

The code for `tst_input_*` uses the build system to handle cross-platform code. All platforms share a single header file, and all platform-dependent behavior is moved into functions. Each platform provides it's own implementation of the function, and we select the correct implementation to compile and link with CMake.

In the case of `tst_input`, we test the logic for all platforms regardless of the host platform. This level of coverage is probably overkill (and likely not possible) for `tst_shell`; `Shell` depends on platform-specific Qt code.

Instead, we can create  a cross-platform `testhelper.h`.

Here is a quick example:
`testhelper.h`:
```
int GetExitCodeForPlatform() noexcept;
```

`testhelper_{platfrom}.cpp`
```
int GetExitCodeForPlatform() noexcept
{
    // Platform-dependant logic goes here
}
```

`CMakelists.txt`:
```
if(WIN32)
	set(SRCS_PLATFORM
		testhelper_win32.cpp
		)
elseif(APPLE)
	set(SRCS_PLATFORM
		testhelper_mac.cpp
		)
else()
	set(SRCS_PLATFORM
		testhelper_unix.cpp
		)
endif()

function(add_xtest SOURCE_NAME)
	add_executable(${SOURCE_NAME} ${SOURCE_NAME}.cpp ${SRCS_PLATFORM} ${ARGV1} ${ARGV2})
	target_link_libraries(${SOURCE_NAME} ${QTLIBS} ${MSGPACK_LIBRARIES} neovim-qt)
	add_test(NAME ${SOURCE_NAME} COMMAND ${SOURCE_NAME})
	add_dependencies(check ${SOURCE_NAME})
endfunction()
```

Same test, different computation for expected value.

statiolake:
> Do you mind if I quickly write the cross-platform handling code and push it to your PR? That might be easiest...

No, not at all, in fact that's very helpful. Thank you.

And thank you for your detailed explanation. Now I understand a little.

test/tst_shell.cpp
statiolake: OK, I think It's possible with your `testhelper.h` approach, by writing a function to get appropriate command line arguments (or a properly configured `QProcess` instance) in each platform-specific source file.

jgehrig:
@statiolake 

I just pushed the build logic changes required for platform-specific test code.

The scaffolding is there for you to add platform-specific test cases. Offhand, I'm not sure what the behavior differences for `:cq` are between platforms. You should probably fix-up and expand up the platform-specific test cases.

statiolake:
@jgehrig Thanks! I'll finish it based on your work.

By the way, `:-1cq` case gives me an invalid range error in **nightly version**. Both on Windows (v0.5.0-dev+1285 and +1062) and Linux (neovim-ppa/unstable on WSL 1).
According to the original description of this behavior (https://github.com/equalsraf/neovim-qt/pull/644#issuecomment-594247405), this was a bit strange from the beginning: `:-1cq` within Neovim gives us no error and returns 0, whereas `nvim -c -1cq` gives us an invalid range error, so it may have been fixed at some point on Neovim side. Although the current stable version (0.4.4) *still* has that strange behavior, but I think this behavior will soon be *fixed* in the released version too (maybe Neovim 0.5?). I think we don't have to test against `:-1cq` case.

jgehrig:
Awesome! :+1: 

> I think we don't have to test against :-1cq case.

Based on what you've described, I agree.

I don't think there is much value to this test if the behavior for nvim is not good. Even if the issue is fixed in 0.5.0, we would be coupling our tests to work with only the latest nvim release. I think that is best avoided.

jgehrig (approved):
The code looks good to me!

Let me know when you feel it is ready (and adequately tested), and I will merge these changes.

statiolake:
@jgehrig Finally I fixed the initialization to use `{}`, and rebased to clean up `fixup!`. It's be ready to be merged when CI completes. I don't think it will change the CI result since I didn't change the logic.

Thanks for your help and reviewing!

statiolake:
Fixed one more part of initialization to `{}`... This time for sure.

jgehrig:
@statiolake

Thanks for the PR! :+1:

somini:
It seems this regressed, I can still reproduce on the latest version. Should I open a new issue instead?

---

I'm using Arch Linux:

```
$ nvim-qt --version
NVIM-QT v0.2.16.1
Build type: Release
```

When running `nvim-qt --nofork`, `:cq` doesn't exit the editor process with error.

zdm:
It still not work for Windows. `
:cq` always throws `Neovim exited with status code` message and no possibility to just exit with error status.

PR #651, opened 2020-03-03
Support arbitrary font weight

NeilGirdhar:


src/gui/shell.cpp
```diff
@@ -137,6 +141,10 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
 				weight = QFont::Bold;
 			} else if (attr == "l") {
 				weight = QFont::Light;
+			} else if (attr == "sb") {
+				weight = QFont::DemiBold;
+			} else if (attr.length() > 0 && attr.at(0) == 'w') {
+				weight = (attr.right(attr.length()-1)).toInt();
```
jgehrig: It might be good to add some user input sanitization like `:h` does above?

```C++
bool ok{ false };
const int fontWeight{ attr.right(attr.length()-1).toInt(&ok) };
if (!ok || fontWeight < 0 || fontWeight > 99) {
	m_nvim->api0()->vim_report_error("Invalid font weight: 0-99");
	return false;
}
weight = fontWeight;
```

Feel free to ignore, not strictly necessary :)

jgehrig:
@NeilGirdhar

This looks great, thanks for working on this! :+1:

I will try to add some unit tests for these changes. The tests are tricky (platform specific ifdef's), so I can handle them along with some code cleanup.

One small comment, feel free to ignore. I can add it in too along with the unit tests. 

We can close #638.

jgehrig:
@equalsraf 

Tested and everything looks good. This should be ready for merge.

NeilGirdhar:
Awesome.  Sorry that I don't have more time to work on this.  Good luck with the project :)

Issue #648, opened 2020-02-22
Incremental search misbehaving


facundobatista:
Incremental search is working sometimes, sometimes not. 

For example, I open a simple file with this content:

```
mailsocios.csv
mails_seminario.txt
Male2019
```

Then I hit `/`, which goes to search. Then I hit `s`. At this point 4 "s" are highlighted, the three in the first line, and the one after the underscore in the second line, but NOT the one previous to the underscore (it should!). Then I hit `e`. Nothing is highlighted anymore (it should highlight the "se" after the underscore in the second line!).

Note that if at that point I hit "enter", the cursor goes to the "se" after the underscore, which is properly highlighted. The *incremental* part of the search is broken, not the search...

This is a short file, but if I have a longer one, the incremental search should "move my view" to what is matching, and that also happens sometimes and sometimes not.

I think this is a neovim-qt issue, because doing the same in nvim (in this file and in all others that I tested this) works just fine.

jgehrig:
I would guess this is a neovim issue, `nvim-qt` is essentially a dumb terminal. All of the logic for text editing is handled by `nvim`, and we simply display a grid of characters.

It would be good to have some example files (you can attach *.txt files), along with a `minimal.vim`. You can load the file with `nvim-qt -- -u minimal.vim example.txt`. You should also test this with a few different versions of `nvim` too.

Once you have a sparse example, and some information on which versions of `nvim` this reproduces for, we can do some quick debugging. If we can verify this is not a nvim-qt issue, we should file a bug to `neovim` instead.

facundobatista:
It's not nvim, in nvim through the terminal works just fine.

I attached [an example file](https://github.com/equalsraf/neovim-qt/files/4258720/search.txt), with the same content I described above.

If I run it "without any start file", it also reproduces: `nvim-qt -- -u None search.txt`

Versions:

```
NVIM-QT v0.2.14.0
Build type: Release
Compilation: -Wall -Wextra -Wno-unused-parameter -std=c++11 -Wfatal-errors
Qt Version: 5.13.2
Environment: 
  nvim: nvim
  args: --cmd set termguicolors
  runtime: 

NVIM v0.4.3
Build type: RelWithDebInfo
LuaJIT 2.0.5
[...]

PR #655, opened 2020-03-13
Issue 648: IncSearch not rendered properly with ext_linegrid.

jgehrig:
**Fixes Issue #648: Incremental search misbehaving**

With `ext_linegrid` some highlights do not render properly:
QMap(("reverse", QVariant(bool, true)))

Any highlight that uses the default colors and "reverse" does not render in
reverse.

This occurs because the reverse parameter is not actually passed to the
rendering layer in ShellWidget.

Adds the "reverse" flag, so default colors can be determined at render time.
This is necessary to support `set background`.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -1053,7 +1053,8 @@ void Shell::handleGridLine(const QVariantList& opargs)
 				hl_attr.IsBold(),
 				hl_attr.IsItalic(),
 				hl_attr.IsUnderline(),
-				hl_attr.IsUndercurl());
+				hl_attr.IsUndercurl(),
+				hl_attr.IsReverse());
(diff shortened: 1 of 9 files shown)


jgehrig:
@facundobatista

This works on my machine for `nvim-qt -- -u NONE` where `ext_linegrid` is on.

I tried for `:set background=light` and `:set background=dark`.

If this doesn't work for you let me know.

facundobatista:
Hello @jgehrig! I can confirm that with this branch, having `ext_linegrid` on, and even with my full config, incremental search behaves just fine! Thanks!

jgehrig:
Thanks for testing @facundobatista!

@equalsraf This one is ready for merge.

Issue #294, opened 2017-06-19
Set GUI title to the current working directory


qwfy:
I often open several nvim-qt instances and often confused which is which, have to click every one of them to find out.

Will nvim-qt add a function/command that allows one to modify the window title?

equalsraf:
isnt this somethign that can be done in vimscript already? A combination of `set titlestring` and dirchanged perhaps?

This https://github.com/neovim/neovim/issues/6260 seems similar to what you want but it uses the status line instead.

I did a quick test with this

```vim
augroup dirchange
    autocmd!
    autocmd DirChanged * let &titlestring=v:event['cwd']
augroup END
```

This needs further tweaking to check the event scope, and to set the correct title string on startup. Perhaps there is already a plugin that does all this?

qwfy:
My bad, tried `titlestring` once before, didn't work, turns out it needs `title` to be on. Now it works.

KevinJohnMulligan:
```vim
set title

augroup dirchange
    autocmd!
    autocmd DirChanged * let &titlestring=v:event['cwd']
augroup END
```

I can confirm that this sets the GUI window title to the file name and adds the the current working directory - Which is very useful! 

Thank you.

rerrabolu:
Any ideas on getting this to work on nvim-qt.

rerrabolu:
Hi Gehrig thanks for taking time to respond:

Just to add results of my experiment:

1. Make sure nvim config file "init.vim" **does not define** the parameter "title"
2. Define the parameter "title" for nvim-qt in its config file "ginit.vim"
3. e.g. set title

Once I did I was able to see nvim-qt window show filename in its title bar. Don't know how all of these things are inter-related. Happy to have it working.

PR #657, opened 2020-03-14
Make the window title match the current tab text

MadCatX:
Presently the Neovim-qt window title does not contain the name of the file open in the active tab. If there are multiple Neovim windows open at once it is very difficult to tell them apart because of this. This tiny patch sets with window title to match the tab text.


jgehrig:
This functionality already exists :)

Take a look at `:help title` or Issue #294 for more details.

Example `vimrc` settings:
```
" Set buffer name as window title
set title

"Optional: set custom window title string
let &titlestring="Neovim Qt Window Title"
```

Issue #649, opened 2020-02-22
Cursor disappear when lines are moved when getting to top/bottom borders


facundobatista:
In any long enough file (more than the window height), when I get 3 lines before the bottom of the window (because I have `set scrolloff=3`), and press the down key, the lines of the buffer go one line up, which is the perfect behaviour for the scrolloff setting.

The problem is that at that moment the cursor disappears.

The same happens when the scroll goes up.

I created [a small video](https://github.com/equalsraf/neovim-qt/files/4239440/disapp-cursor.mp4.gz) to show this better.

jgehrig:
I have noticed similar behavior too.

Does this reproduce with `nvim` terminal? What version of `nvim`/`nvim-qt`?

You should also try this scenario with `:set guicursor=`. There have been some cursor changes recently (if you built from master).

facundobatista:
With nvim in the terminal, it behaves correctly (cursor is always there).

Doing `:set guicursor=` does not affect these behaviours.

Versions:

```
NVIM-QT v0.2.14.0
Build type: Release
Compilation: -Wall -Wextra -Wno-unused-parameter -std=c++11 -Wfatal-errors
Qt Version: 5.13.2
Environment: 
  nvim: nvim
  args: --cmd set termguicolors
  runtime: 

NVIM v0.4.3
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/gcc-5 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -DMIN_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
[...]

PR #662, opened 2020-03-28
Issue 649: Improper cursor redraw for grid_scroll

jgehrig:
**Issue #649**

The cursor should be updated both before and after scroll events. One will
remove the old cursor and the other will draw the new cursor.

Without both draw events, the `scrolloff` setting can cause artifacts.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -1114,13 +1114,16 @@ void Shell::handleGridScroll(const QVariantList& opargs)
 
 	m_scroll_region = QRect(QPoint(left, top), QPoint(right, bot));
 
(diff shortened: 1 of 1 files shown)


jgehrig:
@facundobatista

Let me know if this works for you.

facundobatista:
Hello @jgehrig ! Sorry for the delay! 

I can confirm that with this branch the problem is fixed! Thanks!

jgehrig:
@facundobatista
Not a problem. Thank you for all of your help finding and fixing these bugs!

@equalsraf
Ready for merge.

equalsraf:
Thanks @jgehrig.

PR #666, opened 2020-03-29
Fix build with Qt 5.15

heirecka:
QPainterPath is no longer included via qtransform.h (since
5.15.0-beta2, 50d2acdc93b4de2ba56eb67787e2bdcb21dd4bea in qtbase.git).


jgehrig:
Looks good to me, thanks!

Issue #659, opened 2020-03-18
Alter the Drag and Drop command


j718:
I am trying to insert images into my markdown by drag and drop. So far I have the following script to save them to my default image location and then copy them to the register. Is there any way to edit the drag and drop to handle jpg files without opening a buffer for them?

```
fu! Imgsave()
  let filename=expand("%:t")
  let img_name="images/" . localtime() . filename
  cd ~/repos/notes
  execute "silent w" img_name
  call setreg("", ["", "![" . filename . "]" . "(../" .img_name . ")"])
  History
endfunction

au BufReadPost,BufNewFile *.png,*.jpeg,*.jpg :call Imgsave()
```

The main issue with this approach is that I have to close the new buffer an navigate back to the most recent buffer. It would nice to just not open a new buffer to begin with.

jgehrig:
@j718

Yes, I think the behavior you are looking for is possible. The Drag and Drop feature is simply a call to `function GuiDrop(...)` in `src/gui/runtime/plugin/nvim_gui_shim.vim`.

If you can override that function, you have an entry point to handle Drag and Drop events however you'd like with vimscript.

Maybe we can add something to make overriding the function easier:
```patch
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim  Sat Mar 28 02:54:55 2020 -0400
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim  Sun Mar 29 14:01:00 2020 -0400
@@ -98,7 +98,11 @@
 
 " GuiDrop('file1', 'file2', ...) is similar to :drop file1 file2 ...
 " but it calls fnameescape() over all arguments
-function GuiDrop(...)
+function! GuiDrop(...)
+       if exists('*GuiDropCustom')
+               return call("GuiDropCustom", a:000)
+       endif
+
        let l:fnames = deepcopy(a:000)
        let l:args = map(l:fnames, 'fnameescape(v:val)')
        exec 'drop '.join(l:args, ' ')
```

vimrc example:
```
" This is just an example, it appends all dropped filenames into the current buffer.
function! GuiDropCustom(...)
	let l:fnames = deepcopy(a:000)
	let l:args = map(l:fnames, 'fnameescape(v:val)')

	for l:file in l:args
		echo l:file
		"let line = getline('.')
		"call setline('.', strpart(line, 0, col('.') - 1) . text . strpart(line, col('.') - 1))
		call append(line('.'), l:file)
	endfor
endfunction
```

[...]

PR #669, opened 2020-04-04
Issue 659: Add custom GuiDrop handler

jgehrig:
**Issue #659**

Adds an easy entry point `:GuiDropCustom` for the user to override the default `:GuiDrop` behavior. This is useful for altering the open behavior of nvim-qt.

Example:
```
" Echos the full path of any file opened in TreeView or DragDrop,
function! GuiDropCustom(...)
	let l:fnames = deepcopy(a:000)
	let l:args = map(l:fnames, 'fnameescape(v:val)')

	for l:file in l:args
		echo "You just opened: " . l:file
		exec 'drop '.join(l:file, ' ')
	endfor
endfunction
```

### DIFF
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim
@@ -99,6 +99,11 @@ command! -nargs=1 GuiPopupmenu call s:GuiPopupmenu(<args>)
 " GuiDrop('file1', 'file2', ...) is similar to :drop file1 file2 ...
 " but it calls fnameescape() over all arguments
 function GuiDrop(...)
+	" Check for user defined handler function
+	if exists('*GuiDropCustomHandler')
+		return call("GuiDropCustomHandler", a:000)
+	endif
+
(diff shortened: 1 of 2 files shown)


src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -99,6 +99,11 @@ command! -nargs=1 GuiPopupmenu call s:GuiPopupmenu(<args>)
 " GuiDrop('file1', 'file2', ...) is similar to :drop file1 file2 ...
 " but it calls fnameescape() over all arguments
 function GuiDrop(...)
+	" Check for user defined handler function
+	if exists('*GuiDropCustom')
+		return call("GuiDropCustom", a:000)
```
equalsraf, line 104: Looks good. My only only nitpick would be the name of `GuiDropCustom` vs something like `GuiDropCustomHandler`

jgehrig: Sure, I was struggling with a good name, I like `GuiDropCustomHandler` more.

Issue #671, opened 2020-04-09
`q`/`p`/`r` key is sent by Mute/Volume DOWN/Volume UP


statiolake:
My keyboard has Mute, Volume DOWN and Volume UP keys. If I press Mute key on the Neovim-qt in insert mode, Neovim-qt seems to receive `q` key. That means, pressing Mute key inserts `q` to a document in insert mode or recording is triggered in normal mode macro. The same goes for Volume DOWN and Volume UP key, but the input key is `p` for Volume DOWN key and `r` for Volume UP key. (Note: the volume control function still works as expected, the issue is just the unexpected key stroke Neovim-qt receives.)

According to `git bisect`, this issue is introduced by the commit e13251a.

---

Environment:

- Windows 10 19041.153
- Neovim-qt version: e13251a

PR #672, opened 2020-04-09
Issue 671: Ignore volume keys

statiolake:
**Issue #671**
Fix the issue that pressing volume keys produces `p`, `q` or `r` key event. For now just ignore volume keys. I checked there are no other keys causing this kind of problem in my keyboard (mute mic, brightness down/up) but there might be other keys causing this problem.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -195,6 +195,14 @@ QString convertKey(const QKeyEvent& ev) noexcept
 			return {};
 		}
 
+		// Ignore special keys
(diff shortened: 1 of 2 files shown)


jgehrig:
Sorry for the break, the issue doesn't reproduce for me. I think that shows expanding the input-handling test coverage is essential. There are many quirks and system specific behaviors...

Thank you for finding this and writing unit tests. Let's get this merged! :+1:

equalsraf:
Merged, thanks for this.

Issue #674, opened 2020-04-10
Highlights with `gui=reverse` are not working


statiolake:
In GVim:
![GVim](https://user-images.githubusercontent.com/20490597/78970669-cfa65b00-7b44-11ea-94cb-221589f88936.png)
In Neovim-qt:
![Neovim-qt](https://user-images.githubusercontent.com/20490597/78970618-b3a2b980-7b44-11ea-9be1-5a79f2a428cc.png)

It looks like that `gui=reverse` does not working at all:
![Neovim-qt-gui-NONE](https://user-images.githubusercontent.com/20490597/78970710-ef3d8380-7b44-11ea-8d73-f88b185f314d.png)

By `git bisect`, the problem first occurs in the commit 333dc5556408d037fe793defe799bd636691a903. I think the main cause of this problem is to reverse the color twice.

In `handleGridLine()` in shell.cpp, there is a function call to `ShellWidget::put()`:

```cpp
put(text, row, col_next,
	hl_attr.GetForegroundColor(),
	hl_attr.GetBackgroundColor(),
	hl_attr.GetSpecialColor(),
	hl_attr.IsBold(),
	hl_attr.IsItalic(),
	hl_attr.IsUnderline(),
	hl_attr.IsUndercurl(),
	hl_attr.IsReverse());
```

This is passed to the `Cell` function in `ShellContent::put()` called by `ShellWidget::put()`:

```cpp
cell = { chr, fg, bg, sp, bold, italic, underline, undercurl, reverse };
```

And `Cell`'s constructor creates new `HighlightAttribute` like this ...

```cpp
Cell::Cell(
	uint character,
	QColor fgColor,
	QColor bgColor,
	QColor spColor,
	bool bold,
	bool italic,
	bool underline,
	bool undercurl,
	bool reverse)
	: m_highlight{ fgColor, bgColor, spColor, reverse, italic, bold, underline, undercurl }
```

... which is used in the `ShellWidget::paintEvent()` in shellwidget.cpp.

```
// ... snip ...
	QColor bgColor{ cell.GetBackgroundColor() };
// ... snip ...
	QColor fgColor{ cell.GetForegroundColor() };
// ... snip ...
```

Note that the `GetForegroundColor()` returns reverse-aware foreground color:

```cpp
if (IsReverse()) {
	return m_background;
}
return m_foreground;
```

The same goes for the `GetBackgroundColor()`.

[...]

PR #675, opened 2020-04-10
Issue 674: Render `gui=reverse` color correctly

statiolake:
**Issue #674**
To fix the issue, I added other variants of the related functions which takes a HighlightAttribute and pass the highlight attribute directly instead of reconstructing one. I'd be happy to change the implementation if I should solve this issue in a different way.

### DIFF
--- a/src/gui/shellwidget/cell.h
+++ b/src/gui/shellwidget/cell.h
@@ -7,6 +7,13 @@
 
 class Cell {
 public:
+	/// Create a cell having a specified HighlightAttribute
+	Cell(uint character, const HighlightAttribute& attribute) noexcept :
+		m_highlight(attribute)
+	{
+		SetCharacter(character);
+	}
+
 	Cell(
 		uint character,
 		QColor fgColor,
@@ -16,12 +23,18 @@ public:
 		bool italic,
 		bool underline,
 		bool undercurl,
-		bool reverse);
+		bool reverse) noexcept :
+		Cell{
+			character,
+			{ fgColor, bgColor, spColor, reverse, italic, bold, underline, undercurl } }
+	{
+	}
 
 	/// Create an empty Cell with a background color
-	Cell(QColor bgColor)
-		: m_highlight{ QColor::Invalid, bgColor, QColor::Invalid,
-			false, false, false, false, false }
+	Cell(QColor bgColor) noexcept :
+		Cell{
+			' ',
+			{ QColor::Invalid, bgColor, QColor::Invalid, false, false, false, false, false } }
 	{
 	}
 
--- a/src/gui/shellwidget/shellcontents.h
+++ b/src/gui/shellwidget/shellcontents.h
@@ -30,14 +30,7 @@ public:
 		const QString& str,
 		int row,
 		int column,
-		QColor fg = {},
-		QColor bg = {},
-		QColor sp = {},
-		bool bold = false,
-		bool italic = false,
-		bool underline = false,
-		bool undercurl = false,
-		bool reverse = false) noexcept;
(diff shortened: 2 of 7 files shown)


src/gui/shellwidget/cell.h
```diff
@@ -7,6 +7,13 @@
 
 class Cell {
 public:
+	/// Create a cell having a specified HighlightAttribute
+	Cell(uint character, HighlightAttribute const &attribute) noexcept :
```
jgehrig, line 11: nit: `const HighlightAttribute& attribute`

src/gui/shellwidget/shellcontents.h
```diff
@@ -30,14 +30,7 @@ class ShellContents
...
 		int column,
-		QColor fg = {},
-		QColor bg = {},
-		QColor sp = {},
-		bool bold = false,
-		bool italic = false,
-		bool underline = false,
-		bool undercurl = false,
-		bool reverse = false) noexcept;
+		HighlightAttribute const &hl_attr = HighlightAttribute()) noexcept;
```
jgehrig, line 33: nit: `const HighlightAttribute& hl_attr = {}`

src/gui/shellwidget/shellwidget.h
```diff
@@ -69,6 +69,12 @@ public slots:
 		bool undercurl = false,
 		bool reverse = false);
 
+	int put(
+		const QString& text,
+		int row,
+		int column,
+		HighlightAttribute const &hl_attr);
```
jgehrig, line 76: Here also, `const HighlightAttribute& hl_attr`

src/gui/shellwidget/highlight.h
```diff
@@ -25,6 +25,20 @@ class HighlightAttribute {
 	{
 	}
 
+	/// Copies a existing HighlightAttribute.
+	HighlightAttribute(const HighlightAttribute &other) noexcept :
```
jgehrig, line 29: We should use the default copy constructor instead. The compiler can automatically generate code like this. http://www.cplusplus.com/articles/y8hv0pDG/

Try simply deleting the code first. If that doesn't work, we may be able to use `HighlightAttribute(const HighlightAttribute&) noexcept = default`? I can never remember the exact rules for implicit CTORS...

jgehrig:
@statiolake

Thanks for the detective work! I wasn't aware of `gui=reverse`, that would have been useful while testing...

The FG/BG flip is a vestige of the old protocol rendering code, and was intentional at the time of writing. I think what you have is a good solution. We should pass raw data to the lower layers and let the renderer handle the flipping. I also think `GetBackground()/GetForeground()` should return the reverse-aware colors.

If I understand correctly, both of these statements are true for your solution.

I ran some tests with the old rendering protocol, `ext_linegrid=false`. Everything looks good with these changes for that protocol too.

src/gui/shellwidget/highlight.h
statiolake: Actually simply deleting those lines worked. I've misunderstood when implicit constructors are defined, and also I've completely forgot about `= default`... They're removed in 7a1434d8a186f16bdf9cf906ab099e584d380430.

jgehrig:
Thanks! The changes look good to me now :)

If you can, it'd be nice to clean up the history with a `git rebase -i`. You should be able to delete `03c8252`/`7a1434d` and squash `c893e25`.

jgehrig:
@equalsraf This looks good to me.

Let's get this merged to resolve Issue 674 and Issue 687.

equalsraf:
ACK, thanks @statiolake

PR #676, opened 2020-04-11
Add GuiFontWide

yehuohan:
Add GuiFontWide

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -157,66 +157,50 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
 bool Shell::setGuiFontWide(const QString& fdesc, bool force, bool updateOption)
 {
 	bool setShellFontSuccess{ false };
-	const bool isGuiDialogRequest{ fdesc == "*" };
 
-	if (isGuiDialogRequest) {
-		bool fontDialogSuccess = false;
-		QFont font{ QFontDialog::getFont(&fontDialogSuccess, QWidget::font(), this, {},
-			QFontDialog::MonospacedFonts) };
-
-		// The font selection dialog was canceled by the user.
-		if (!fontDialogSuccess) {
+	const QStringList wideFontList{ fdesc.split(',') };
+	for (const auto& wfdesc : wideFontList)
+	{
+		// Exit early if the font description does not change, prevent loops.
+		if (wfdesc.compare(fontDesc(true), Qt::CaseInsensitive) == 0) {
 			return false;
 		}
 
-		setShellFontSuccess = setShellFont(font.family(), font.pointSize(), font.weight(),
-			font.italic(), force, true);
-	}
-	else {
-		const QStringList wideFontList{ fdesc.split(',') };
-		for (const auto& wfdesc : wideFontList)
-		{
-			// Exit early if the font description does not change, prevent loops.
-			if (wfdesc.compare(fontDesc(true), Qt::CaseInsensitive) == 0) {
-				return false;
-			}
-
-			QStringList attrs = fdesc.split(':');
-			if (attrs.size() < 1) {
-				m_nvim->api0()->vim_report_error("Invalid wide font");
-				return false;
-			}
-
-			// Use pointSizeF of non-wide font by default
-			qreal pointSize = font().pointSizeF();
-			int weight = -1;
-			bool italic = false;
-			for (const auto& attr : attrs) {
-				if (attr.size() >= 2 && attr[0] == 'h') {
-					bool ok{ false };
-					qreal height = attr.mid(1).toFloat(&ok);
-					if (!ok) {
-						m_nvim->api0()->vim_report_error("Invalid wide font height");
-						return false;
-					}
-					pointSize = height;
-				} else if (attr == "b") {
-					weight = QFont::Bold;
-				} else if (attr == "l") {
-					weight = QFont::Light;
-				} else if (attr == "sb") {
-					weight = QFont::DemiBold;
-				} else if (attr.length() > 0 && attr.at(0) == 'w') {
-					weight = (attr.right(attr.length()-1)).toInt();
-				} else if (attr == "i") {
-					italic = true;
-				}
-			}
-
-			setShellFontSuccess = setShellFont(attrs.at(0), pointSize, weight, italic, force, true);
-			if (setShellFontSuccess)
-				break;
+		QStringList attrs = fdesc.split(':');
+		if (attrs.size() < 1) {
+			m_nvim->api0()->vim_report_error("Invalid wide font");
+			return false;
 		}
+
+		// Use pointSizeF of non-wide font by default
+		qreal pointSize = font().pointSizeF();
+		int weight = -1;
+		bool italic = false;
+		for (const auto& attr : attrs) {
+			if (attr.size() >= 2 && attr[0] == 'h') {
+				bool ok{ false };
+				qreal height = attr.mid(1).toFloat(&ok);
+				if (!ok) {
+					m_nvim->api0()->vim_report_error("Invalid wide font height");
+					return false;
+				}
+				pointSize = height;
+			} else if (attr == "b") {
+				weight = QFont::Bold;
+			} else if (attr == "l") {
+				weight = QFont::Light;
+			} else if (attr == "sb") {
+				weight = QFont::DemiBold;
+			} else if (attr.length() > 0 && attr.at(0) == 'w') {
+				weight = (attr.right(attr.length()-1)).toInt();
+			} else if (attr == "i") {
+				italic = true;
+			}
+		}
+
+		setShellFontSuccess = setShellFont(attrs.at(0), pointSize, weight, italic, force, true);
+		if (setShellFontSuccess)
+			break;
 	}
 
 	// Only update the ShellWidget when font changes.
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -85,36 +85,10 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
 	}
 
 	bool setShellFontSuccess{ false };
-	const bool isGuiDialogRequest{ fdesc == "*" };
-
-	if (isGuiDialogRequest) {
-		bool fontDialogSuccess = false;
-		QFont font{ QFontDialog::getFont(&fontDialogSuccess, QWidget::font(), this, {},
-			QFontDialog::MonospacedFonts) };
-
-		// The font selection dialog was canceled by the user.
-		if (!fontDialogSuccess) {
-			return false;
-		}
-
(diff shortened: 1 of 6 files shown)


jgehrig:
Very cool!

However, can we use `guifontwide` instead? We should use the `neovim` configuration APIs whenever possible. We should add `neovim-qt` specific configuration options only as absolutely necessary.

You can handle the `guifontwide` event here:
```C++
void Shell::handleSetOption(const QString& name, const QVariant& value)
{
	if (name == "guifont") {
		setGuiFont(value.toString(), false /*force*/, false /*setOption*/);
	} else if (name == "guifontset") {
	} else if (name == "guifontwide") {
		// { HANDLE HERE }
		// You will need to parse "value" for the relevant arguments
		// bool setGuiFont(... , true /*wide*/)
	} else if (name == "linespace") {
```

This would allow users to modify the wide font with `set guifontwide=...`.

Such a change also simplifies this Pull. You shouldn't need to add anything to the shim/runtime.

src/gui/shell.cpp
```diff
@@ -882,9 +884,10 @@ void Shell::handleExtGuiOption(const QString& name, const QVariant& value)
 void Shell::handleSetOption(const QString& name, const QVariant& value)
 {
 	if (name == "guifont") {
-		setGuiFont(value.toString(), false /*force*/, false /*setOption*/);
+		setGuiFont(value.toString(), false /*force*/, false /*setOption*/, false);
 	} else if (name == "guifontset") {
 	} else if (name == "guifontwide") {
+		setGuiFont(value.toString(), false /*force*/, false /*setOption*/, true);
```
jgehrig, line 890: I think the logic here needs some modification. The `guifontwide` option is supposed to be a list of fonts.

From `:help guifontwide`:
> When not empty, specifies a comma-separated list of fonts to be used for double-width characters.  The first font that can be loaded is used.

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -32,15 +32,19 @@ void ShellWidget::setDefaultFont()
 {
 #if defined(Q_OS_MAC)
 #  define DEFAULT_FONT "Courier New"
+#  define DEFAULT_FONT_WIDE "Courier New"
```
jgehrig, line 35: Let's not expand preprocessor macros; the values are the same anyways.

You can use `DEFAULT_FONT`.

src/gui/shell.cpp
```diff
@@ -95,7 +95,7 @@ QString Shell::fontDesc()
 ///
 /// @param updateOption controls update of the guifont option, which should not
 /// be updated when the user calls from 'set guifont=...'.
-bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
+bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption, bool wide)
```
jgehrig, line 98: I think `guifontwide` should have a separate function `Shell::setGuiFontWide`. This function contains `guifont/GuiFont` specific behavior that you likely don't need to run for `guifontwide`.

I think the only part you need is the `else` block that parses description. We should probably move its own function. This would be a little tricky... Let me know if I can help with this refactor.

src/gui/shellwidget/shellwidget.cpp
yehuohan: I removed DEFAULT_FONT_WIDE, and use same font for wide by default.

src/gui/shell.cpp
yehuohan: Of course. I'm glad you can help with this refactor. Actually. I'm not familiar with the behavior of list fonts to fontwide option.

jgehrig: Excellent, thanks!

I will put something together that makes it easy for you to write this without `setGuiFont`. I can get you a patch file, or I can commit the changes and you can re-base on top. Let me know if you have a preference.

The code that you need in `setGuiFont` handles the options parsing (`:b`, `:hNN`, etc). We can add a function `GetQFontFromDescription(const QString& description)` and modify the ShellWidget to accept a QFont for `setShellFont(...)` instead of the existing API. You can then write a `Shell::setGuiFontWide` function that handles wide font specific scenarios only.

For the `guifontwide` list behavior (rough code):
```C++
const QStringList wideFontList{ fdesc.split(',') };

for (const auto& fontDescription : wideFontList)
{
    QFont wideFont{ GetQFontFromDescription(fontDescription) };
    // There should probably be some validation here?
    // Also, we might want to update the list only after validating all provided fonts are good?
    m_wideFontList.emplace_back(std::move(wideFont));
}
```

The user should be able to provide multiple wide fonts. Rendering should then pick the first font from that list that contains the double-wide character.

`set guifont=Some\ Font:h11,Another\ Font:h12:b`

We should check `gVim` to make sure our behavior is the same... Ideally, this feature should behave identically.

yehuohan: Thank you and I'd like get the helps with your commit.
With your explanation of guifontwide list behavior, I try to add the funtion `Shell::setGuiFontWide`. There's is one point confused: use which pointsize of font on ShellWidget (font's or widefont's)? Currently, `ShellWidget`'s font set to `m_font` by default, and m_fontWide is used only for double-width char as the code below, which may cause the issue that the command `GuiFontWide` won't update the rendering of chars immediately when it's executed.

```cpp
void ShellWidget::setFont(const QFont& f, bool wide)
{
	if (wide)
		this->m_fontWide = f;
	else {
		this->m_font = f;
		QWidget::setFont(f);
	}
}
```

yehuohan: What's more, I simply use the result of `ShellWidget::setShellFont` as the validation of wide font. And I think some commit more is required to check wide font containing the double-wide character or not.

jgehrig:
Moving the conversation...

> There's is one point confused: use which pointsize of font on ShellWidget (font's or widefont's)? Currently, ShellWidget's font set to m_font by default, and m_fontWide is used only for double-width char as the code below, which may cause the issue that the command GuiFontWide won't update the rendering of chars immediately when it's executed.

I need to check some behavior with `gVim`... Pending.

I think the behavior would be `set guifontwide=Some_Font:h11,Another_Font:h12:b,Yet_Another_Font`. So each font should be able to parse the usual font options like `:h`. I'm not sure what happens when no modifiers are provided?

We will need something like: https://github.com/jgehrig/neovim-qt/commit/d80e0e872d88539340597889e0e01802becbb8db

This moves the description parsing code to `ShellWidget::GetQFrontFromDescription(...)`. I may change the return type, this seems like an abuse of QVariant. However, `std::optional<T>` is not available in this project with only C++11.

> I moved the Shell::fontDesc to ShellWidget

Great! I think that is a logical place for the function to live.

> And I think some commit more is required to check wide font containing the double-wide character or not.

Yes, I think there needs to be a `std::vector<QFont> ShellWidget::m_fontWide;`. Or maybe `m_fontWideList`?

We then need some additional logic in the rendering event (pseudocode):
```C++
if (/*isDoubleWideChar*/) {
    QFont renderFont = m_font;
    for (auto fontWide : m_fontWideList) {
        if (/*current char is a valid char in fontWide*/) {
            break;
        }
    }
    // Paint double wide character with renderFont
```

src/gui/shell.cpp
```diff
@@ -161,12 +143,96 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption, bool
...
+
+		setShellFontSuccess = setShellFont(font.family(), font.pointSize(), font.weight(),
+			font.italic(), force, true);
+	}
+	else {
+		const QStringList wideFontList{ fdesc.split(',') };
+		for (const auto& wfdesc : wideFontList)
+		{
+			// Exit early if the font description does not change, prevent loops.
+			if (wfdesc.compare(fontDesc(true), Qt::CaseInsensitive) == 0) {
```
jgehrig, line 180: This should be outside of the for loop.

This checks if the user has changed the value of the font register. IE there is no need to do anything when the register remains unchanged.

src/gui/shell.cpp
```diff
@@ -161,12 +143,96 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption, bool
...
 	// unwanted recursion. Only update this option for ':GuiFont', and dialog calls.
 	if (isGuiDialogRequest || updateOption) {
-		m_nvim->api0()->vim_set_option("guifont", fontDesc());
+		m_nvim->api0()->vim_set_option("guifont", fontDesc(false));
+	}
+
+	return true;
+}
+
+bool Shell::setGuiFontWide(const QString& fdesc, bool force, bool updateOption)
```
jgehrig, line 157: A lot of this logic is duplicated from `setGuiFont`, but no longer necessary. 

We should get rid of `GuiFontWide`, so the update logic is not necessary.
We should consider getting rid of `force`.
I'm not sure how the font dialog works with a list of fonts, it should also be removed.

This feature should mimic gvim.

jgehrig: Also... Fine for now to get things working.

However, we shouldn't copy-paste code for the final iteration. Instead we should move the re-used parts to functions. Just a friendly reminder :)

yehuohan: > Yes, I think there needs to be a std::vector<QFont> ShellWidget::m_fontWide;. Or maybe m_fontWideList?

I use `QVector<QFont>` to store the wide font list. Is it neccessary use `std::vector`?
With using `m_fontWideList`, I add `ShellWidget::createShellFont` and I think the `ShellWidget::setShellFontWide(const QVector<QFont>& font)` is required. 

I'm agree that re-using code, and there `Shell::getFontFromDesc` was add to handle all font description with help of `ShellWidget::GetQFrontFromDescription(...)` .

yehuohan: > We then need some additional logic in the rendering event (pseudocode):

`QFontMetrics` is used to check whether wide font contains double-character. But it's seemed that `QFontMetrics(non-wide font).inFontUcs4` return `true` too. Is it possible that `QFontMetrics` find a default font from system for double-character? 

```
if (cell.IsDoubleWidth()) {
	for (const auto& fw : m_fontWideList) {
		if (QFontMetrics(fw).inFontUcs4(cell.GetCharacter())) {
			f = QFont(fw);
			break;
		}
	}
} else {
        ...
}
```

jgehrig: > I use QVector<QFont> to store the wide font list. Is it neccessary use std::vector?

No, Qt containers work fine too. I think the optimal Qt container for this case is `QList`, we don't need the memory continuity guarantee of `QVector`.

> But it's seemed that QFontMetrics(non-wide font).inFontUcs4 return true too.

Interesting... I'm not familiar enough with the inner workings of QFontMetrics. We'll have to investigate this.

yehuohan: `QList` was used to store wide fonts. :)

jgehrig:
Merged in #689.

Thanks for your contribution @yehuohan!

Issue #170, opened 2016-08-12
<C-6> mappings dont work


MaikoID:
Hey guys very nice work so far.

I'm having a strange behaviour where I can't switch between current and previous windows buffer from nvim-qt, nvim in terminal works just fine.
There is no CTRL printed when I do the cmd bellow:
`:map <C-V><C-6>`
it gets replaced by just  `:map 6`

Any map using CTRL+NUMBER simple doesn't work; This is a huge problem for me because I'm used to change my buffer by numbers =|

equalsraf:
Ok we need to narrow this down a bit, Ctrl-number mappings should work, for example the following works for me in normal mode

```
nmap <C-1> iC-1 is ok<CR><ESC>
```

But other numbers e.g. **6** dont work

```
nmap <C-6> iC-6 is ok<CR><ESC>
```

I could use some more info about your setup

- what system are you running on Linux/Windows
- What keyboard layout do you use?

Here is a code snippet if anyone wants to debug it

```diff
diff --git a/src/gui/shell.cpp b/src/gui/shell.cpp
index 05ca325..3f6996a 100644
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -563,12 +563,14 @@ void Shell::keyPressEvent(QKeyEvent *ev)
 
        // FIXME mousehide - conceal mouse pointer when typing
 
+       qDebug() << "Received keypress" << ev->text() << ev->key();
        QString inp = Input.convertKey(ev->text(), ev->key(), ev->modifiers());
        if (inp.isEmpty()) {
                QWidget::keyPressEvent(ev);
                return;
        }
 
+       qDebug() << "Sending input" << inp;
        m_nvim->neovimObject()->vim_input(m_nvim->encode(inp));
        // FIXME: bytes might not be written, and need to be buffered
 }
```

For C-6 it gets me ([Qt::key reference](http://doc.qt.io/qt-5/qt.html#Key-enum))

```
Received keypress "" 16777249
Received keypress "\u001E" 54
Sending input "\u001E"
Received keypress "" 16777250
```

MaikoID:
Hi, thanks for your answer, see the solicited info bellow: 

- Ubuntu 14.04
- pt-br

BR

MaikoID:
Update:
I had to install Ubuntu 16.04 because a fault intel video driver; 
I'm glad I can't reproduce this issue for now, because it's working as expected in on 16.04.

BR

[...]

PR #682, opened 2020-04-16
Issue 170: Send <C-6> as <C-^>

jgehrig:
**Issue #170:**

The key event <C-6> should be sent as <C-^>. When other modifiers are present (Shift, Meta) they should be ignored. For example, <C-S-^> should be <C-^> instead.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -175,6 +175,14 @@ QString convertKey(const QKeyEvent& ev) noexcept
 		return ToKeyString(GetModifierPrefix(modNoShift), "lt");
 	}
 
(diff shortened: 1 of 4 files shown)


jgehrig:
@equalsraf

This one is ready. The behavior isn't *exactly* the same as nvim, but it is a step in the right direction, as we no longer send invalid input such as `<C-S-^>`.

On Window/Mac, there are some small input differences with nvim. We are slightly more aggressive in sending `<C-^>`, for example on Mac Ctrl + 6 in `nvim` may not trigger `<C-^>`; Ctrl + Shift + 6 is required.

The test failures are not related... I just ran `tst_msgpackiodevice` on Mac successfully.

Issue #665, opened 2020-03-28
Busy mouse cursor overridden by mouse hide


jgehrig:
The mouse hide feature overrides the busy mouse cursor.

See screenshot:
![busy-mouse-cursor](https://user-images.githubusercontent.com/11207308/77817311-5ae11480-70a0-11ea-9759-82c752f64850.gif)

**Repro Steps:**
1. `nvim-qt -- -u NONE`
2. Place the mouse over the nvim-qt window.
3. `:terminal`
4. Enter the terminal with `i`
5. Notice the busy cursor
6. Type something, cursor disappears.
7. Move the mouse.
8. Notice the busy cursor is gone, bug.

PR #688, opened 2020-05-08
Issue 665: Busy mouse cursor overridden by mouse hide

jgehrig:
**Issue #665:**

After the mouse is hidden, movements can trigger the cursor to reappear. We
need to check the busy state to determine if a wait cursor or regular cursor
should be displayed.

@equalsraf
Simple fix, ready for merge.


equalsraf:
Merged, thanks.

Issue #544, opened 2019-05-31
How can I set an alternative font？


mzr1996:
- `nvim --version`:
```
NVIM v0.4.0-819-g66d127957
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe /DWIN32 /D_WINDOWS /W3 /MD /Zi /O2 /Ob1 /DNDEBUG -DMIN_LOG_LEVEL=3 /W3 -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -DWIN32 -D_WIN32_WINNT=0x0600 -DINCLUDE_GENERATED_DECLARATIONS -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -IC:/projects/neovim/build/config -IC:/projects/neovim/src -IC:/projects/nvim-deps/usr/include -IC:/projects/neovim/build/src/nvim/auto -IC:/projects/neovim/build/include
Compiled by appveyor@APPVYR-WIN

Features: -acl +iconv +tui
```
- Operating system/version:
Windows Home Basic 1903

I want to set Latin characters' font and Chinese characters' font separately. And in qvim，I can do it by the option `guifontwide`, while that option does not work in nvim-qt. I have also tried Guifont,  but it can't seem to set more than one font. 

I'm wondering if I can set an alternative fonts, just like URxvt's configuration:
```
URxvt.font:xft:Hack Nerd Font Mono:pixelsize=23, xft:Sarasa Mono SC:pixelsize=26
```
It will apply the first font primarily, and apply the second font when the character can't be found in the first font.

Or just like Gvim's configuration:
```
set guifont=Hack\ NF:h14
set guifontwide=等宽更纱黑体\ SC:h16
```
It will apply `guifontwide` to the double-width characters.

jgehrig:
This feature is available in `master` now. It should be included in the `v0.2.17` release.

Marking as closed.

PR #689, opened 2020-05-09
Add option guifontwide

jgehrig:
**Issue #544:** Set alternative wide-character font.
**Pull #676:** Used as a starting point, refactored to keep a clean diff.

guifontwide_screen_recording

@yehuohan

Sorry for the silence on this feature... This is how I envision `guifontwide` operating in `neovim-qt`:

1. `:GuiFontWide` is no longer necessary, since `guifontwide` exists.

2. The option `guifontwide` can be tolerant of mono-space issues. Disregard invalid fonts at runtime.

3. We should use this as an opportunity to add test coverage for font conversion.

The refactor is done before any implementation work on `gufontwide`, commits:
fb22912 83fcd17 e9411ab

With these changes you could cleanly implement `guifontwide`.

**Option A:** I check in the refactor, and you finish up `guifontwide`. Your changes would be re-based on top.

**Option B:** You're tired of working on this... I can finish the work for the entire feature.

Let me know... I don't want to steal your feature, my intent is only to help :)


yehuohan:
I think your commit is perfect and I'd perfer that you can finish the work.
I build the branch ` jgehrig:jg-guifontwide`, which worked well for me.

> The option guifontwide can be tolerant of mono-space issues. Disregard invalid fonts at runtime.

This point is cool. With this I can set font and wide-font just in init.vim but not using `GuiFont!` command at `VimEnter` to avoid reporting `bad fixed pitch metrics` .

yehuohan:
I think it's a good idea that performing the cell's bold and italic attributes to the widefont in the function `ShellWidget::GetCellFont`. A simple code like below :)

```
	if (fontHasCharacter) {
		if (cell.IsBold() || cell.IsItalic()) {
			QFont cellFont{ wideFont };
			cellFont.setBold(cell.IsBold());
			cellFont.setItalic(cell.IsItalic());
			return cellFont;
		}
		return wideFont;
	}
```

jgehrig:
Sounds good. I will pick up where you left off and get this checked-in.

> I think it's a good idea that performing the cell's bold and italic attributes to the widefont in the function ShellWidget::GetCellFont

Good catch! That is definitely a bug...

jgehrig:
Interesting bug... It looks like the fix for Issue #575 does not propagate through `QVariant`. Fixed.

The call to `setStyleName("")` has been moved to before render-time in `ShellWidget::paintEvent`. In this location, we only need to clear this value once. Some of the other properties are required for fixed-pitch validation to work.

I did validate that all other properties translate from `QVariant` to `QFont`.

jgehrig:
Testing help would be appreciated on this issue.

I only speak English, so I am not particularly familiar with multi-byte characters or corner cases in the usage of `guifontwide`.

PR #692, opened 2020-05-12
add makefile for quick lunch up

charlie-wong:


equalsraf:
Hi @charlie-wong

While i understand the appeal, I'd rather not have to maintain yet another layer of indirection just for the purpose of aliasing the default build commands. This leads to duplicating cmake targets in make and we have to work we two build systems, even if cmake is generating makefiles.

To the best of my knowledge this is how terse you can get with cmake

```
$ cmake -S . -B build
(... snip)
$ cmake --build build
(... snip)
```
or if you dont care about building in the source folder

```
$ cmake .
(... snip)
$ cmake --build .
(... snip)
```

Having said all this, I dont mind having build/packaging recipes for specific targets in contrib/. And the change to update .gitignore with user cmakelists is welcome.

PR #697, opened 2020-05-20
Add support for Float API values

jamessan:
nvim 0.5 (API 7) has methods that use `Float` (`MSGPACK_OBJECT_FLOAT`) as their type.  Update bindings generation and msgpack handling so neovim-qt can access these APIs.

### DIFF
--- a/bindings/generate_bindings.py
+++ b/bindings/generate_bindings.py
@@ -50,6 +50,7 @@ class NeovimTypeVal:
     SIMPLETYPES = {
             'void': 'void',
             'Integer': 'int64_t',
+            'Float': 'double',
             'Boolean': 'bool',
             'String': 'QByteArray',
             'Object': 'QVariant',
--- a/src/msgpackiodevice.cpp
+++ b/src/msgpackiodevice.cpp
@@ -515,6 +515,21 @@ bool MsgpackIODevice::decodeMsgpack(const msgpack_object& in, int64_t& out)
 	return false;
 }
 
+void MsgpackIODevice::send(double d)
+{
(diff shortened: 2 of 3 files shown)


jgehrig:
Is there a plan to add functionality to `neovim-qt` that requires msgpack `Float` parameters?

If there isn't a caller, I think we should wait to merge this code. Less code means less maintenance, smaller binaries, and faster compile times.

I think this code is useful as a reference and would be worth merging once required by a caller.

jamessan:
> If there isn't a caller, I think we should wait to merge this code. Less code means less maintenance, smaller binaries, and faster compile times.

The amount of potentially unused code here is nothing compared to the vast swaths of nvim API that have generated code and aren't used.  The neovimapiX.cpp modules are some of the longest compile times in the code base.

This change ensures that the nvim API is fully reflected in neovim-qt.  In particular, the new API that uses `Float` is `nvim_ui_pum_set_bounds` which would likely be helpful to ensure the popup menus and nvim floats interact nicely.

jgehrig:
Understood. I wish there was less unused auto-generated code too...

> This change ensures that the nvim API is fully reflected in neovim-qt. In particular, the new API that uses Float is nvim_ui_pum_set_bounds which would likely be helpful to ensure the popup menus and nvim floats interact nicely.

Thanks for the additional context. This sounds like a good reason to merge the changes, I'm convinced.

I filed an Issue to make sure I generate `neovimapi7.cpp` once API 7 is finalized.

equalsraf:
Thanks @jamessan

Issue #707, opened 2020-06-18
"nvim-qt --help" crashes if it can't connect to $DISPLAY


jamessan:
Since 847f0271f2af61a35e8676e069a2e8b31a527760, running `env -u DISPLAY nvim-qt --help` crashes.

I noticed this because the Debian packaging uses `nvim-qt --help` (via help2man) to create a man page.  In the build, there's no `$DISPLAY` unless one is specifically setup so this started crashing with v0.2.16.

jgehrig:
@jamessan Thanks for finding the root cause!

@equalsraf Since I introduced this, I am happy to write a fix.

Interesting bug... The `QApplication` constructor causes the crash. We will need to use `QCoreApplication`, unfortunately Qt makes sharing code between the two difficult.

jgehrig:
Also worth noting...

The behavior is common to most Qt applications. All of my KDE applications crash in the same manner:
```
$ env -u DISPLAY dolphin --help

qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: wayland-org.kde.kwin.qpa, minimal, offscreen, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, xcb.

Aborted (core dumped)
```

This seems like an upstream bug in Qt. However, this is also a regression and bad behavior...

jgehrig:
Confirmed `env -u DISPLAY nvim-qt --help` is working in the latest release. Closing.

PR #709, opened 2020-06-19
--help should work when $DISPLAY is not set

equalsraf:
see #707

No code yet, just a test case.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -30,7 +30,7 @@ int ui_main(int argc, char **argv)
 
 	NeovimQt::App app(argc, argv);
 
-	app.checkArgumentsMayTerminate();
+	app.checkArgumentsMayTerminate(app.commandLineParser());
 
 	app.connectToRemoteNeovim();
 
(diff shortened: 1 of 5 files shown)


src/gui/main.cpp
```diff
@@ -51,9 +51,10 @@ int cli_main(int argc, char **argv)
 		argsNoFork << argv[i];
 	}
 
-	NeovimQt::App app(argc, argv);
-
-	app.checkArgumentsMayTerminate();
+	QCoreApplication app(argc, argv);
+	QCommandLineParser p;
+	NeovimQt::App::processCommandlineOptions(p, app.arguments());
+	NeovimQt::App::checkArgumentsMayTerminate(p);
```
equalsraf, line 57: I''m not entirely happy with this just yet. But i cant fully split cli argument handling from App.

equalsraf:
I had to remove --help-all from the tests,  it was not behaving consistently between CI and my box.

I'm not too happy about the way the code went, but i don't really see an obvious way out. @jgehrig  any bright ideas for that QCoreApplication bit?

jgehrig:
I think this is a bug in Qt.

Any solutions I can think of to work around the issue are ugly. We can't have build a `QApplication` that extends `QCoreApplication` due to the way `Q_OBJECT` works. We also can't use the deprecated `QApplication` CTOR with `GUIenabled = false`.

So far, I can't think of anything better that the code you've written here.

equalsraf:
The only thing that came to mind was to pass the QCommandLineParser as an argument to App (after it is already built and processed arguments)

But I was under the impression that QCommandLineParser was dependent on an existing QCoreApplication instance, but maybe i was wrong?

PR #710, opened 2020-06-20
Install SVG icon to scalable directory

jamessan:
N.B., neovim.png is actually 200x200, not 192x192.  Not sure how strict the consumers of the icon are, but thought I'd mention it.


jgehrig:
@jamessan

Looks good to me, thanks!

> N.B., neovim.png is actually 200x200, not 192x192

Good catch, I think this is worth fixing. We should re-generate or re-scale the PNG to 192x192. 

Let me know if you'd like to handle this. If not, I can fix the issue in a separate PR.

jamessan:
> We should re-generate or re-scale the PNG to 192x192.
> 
> Let me know if you'd like to handle this. If not, I can fix the issue in a separate PR.

I'd rather let you handle it.  Thanks.

jgehrig:
Sounds good, I'll work on the PNG size issue. Thanks.

PR #711, opened 2020-06-20
Do not install src/gui/runtime/README.md

jamessan:


jgehrig:
Thanks for cleaning up the install scripts!

PR #715, opened 2020-06-20
Resize icon neovim.png to 192x192

jgehrig:
This icon installs to a directory for icons of size 192x192, but the size o this PNG is 200x200. The icon should be re-sized to correct size of 192x192.

This image was generated from `neovim.svg` using GIMP and the default PNG encoding settings. Do we need any special settings?.

This issue was brought up in Pull Request #710


equalsraf:
Nope. This is good to go. Pressing the button.

PR #716, opened 2020-06-21
Fix coverage measurements

equalsraf:
These have been broken for a while, lets see if we can bring it back to life

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,9 @@ jobs:
           command: sudo apt-get update
       - run:
           name: Install dependencies
-          command: sudo apt-get install -y cmake libqt5svg5 libqt5svg5-dev neovim python3-dev python3-jinja2 python3-msgpack python3-pip qt5-default
+          command: |
+            sudo apt-get install -y cmake libqt5svg5 libqt5svg5-dev neovim python3-dev python3-jinja2 python3-msgpack python3-pip qt5-default
+            pip install --user cpp-coveralls PyYAML
       - run:
           name: build
           command: |
@@ -29,4 +31,9 @@ jobs:
           command: |
             cd build
             ctest -VV
+      - run:
+          name: Upload coverage report
+          # trick cpp-coveralls into doing the right thing
+          # https://github.com/eddyxu/cpp-coveralls/issues/143
+          command: TRAVIS_JOB_ID="#${CIRCLE_BUILD_NUM}" ~/.local/bin/cpp-coveralls -y contrib/coveralls-circle.yml -E '.*/build/.*' -E '.*/third-party/.*' -E '.*/test/.*' -E '.*/bindings/.*'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,16 @@ before_install:
   - brew ls | grep -wq cmake || brew install cmake
   - brew ls | grep -wq qt5 || brew install qt5
   - brew ls | grep -wq msgpack || brew install msgpack
+  # cpp-coveralls https://github.com/eddyxu/cpp-coveralls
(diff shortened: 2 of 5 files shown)


jgehrig:
Purely curiosity... What will this test coverage do?

It's great to see so much development activity lately. Good stuff! 😁

equalsraf:
> Purely curiosity... What will this test coverage do?

It just tells you what portions of the code were in fact executed by your tests. Which is nice to highlight stuff that you forgot to write tests for.

For example here is one attempt, but it is reporting on some code that it should not (tests, autogenerated moc, third-party msgpac)

- [tree](https://coveralls.io/builds/31588202)
- [specific file](https://coveralls.io/builds/31588202/source?filename=src/neovimconnector.cpp)

equalsraf:
I cant get Travis to work, so i think i'll leave it with Circle CI for now.

The coverage is currently at 10% which is significantly lower than previously (>40%). This is a bit hard to  see in the coveralls [UI](https://coveralls.io/jobs/64305621), but if we order the list view by missed lines itshows we fail to cover the auto generated code in `src/auto`. I assume fixing `TestCallAllMethods::callAll` to cover other api versions will help a bit (but not by much).

If we ignore `src/auto` we are probably closer to the 50% mark. `src/gui` clocks 52% and other large files are similar.

equalsraf:
Mmmm, what would be the best way to call tests for all api versions. This is as far as i got here. The template method allows me to write one single test function, but I'm still missing a way to avoid the repetition when calling the test.

```diff
diff --git a/test/tst_callallmethods.cpp b/test/tst_callallmethods.cpp
index e810363..eb8f1ac 100644
--- a/test/tst_callallmethods.cpp
+++ b/test/tst_callallmethods.cpp
@@ -24,7 +24,11 @@ private slots:
        void vim_call_function();
        void callAll();
 
+       void buffer_line_count();
 private:
+       template <typename API>
+       void buffer_line_count(API*);
+
        void call_metaobject_slots(QObject *obj);
        NeovimQt::NeovimConnector *m_c;
 };
@@ -120,5 +124,29 @@ void TestCallAllMethods::vim_call_function()
        QCOMPARE(result.at(0).at(0), QVariant(2));
 }
 
+void TestCallAllMethods::buffer_line_count()
+{
+       buffer_line_count(m_c->neovimObject());
+       buffer_line_count(m_c->api0());
+       buffer_line_count(m_c->api1());
+       buffer_line_count(m_c->api2());
+       buffer_line_count(m_c->api3());
+       buffer_line_count(m_c->api4());
+       buffer_line_count(m_c->api5());
+       buffer_line_count(m_c->api6());
+}
+
+template <typename API>
+void TestCallAllMethods::buffer_line_count(API* obj)
+{
+       QVERIFY(obj);
+
+       QSignalSpy result(obj, SIGNAL(on_buffer_line_count(int64_t)));
+       QVERIFY(result.isValid());
+
+       obj->buffer_line_count(0);
+       QVERIFY(SPYWAIT(result));
+}
+
 QTEST_MAIN(TestCallAllMethods)
 #include "tst_callallmethods.moc"
```

I can always "macro" my way out of it. I wonder if there is something in c++17, but we probably dont want to depend on that.

jgehrig:
I don't have any elegant/clever solutions to remove the macro or hard-coding.

I don't think there is a type-safe way to accomplish this. Each API is an unrelated class, and each function call is a different overload. Also, nothing guarantees that a specific function is available in all APIs. 

I think the macro, or even explicit hard-coding of the calls is fine.

> I wonder if there is something in c++17, but we probably dont want to depend on that.

Unrelated, we should probably bump the C++ version at some point.

I have occasionally wanted to use `std::optional` and other new features. C++14 or C++17 should be available almost everywhere now.

PR #719, opened 2020-07-02
Add GuiAdaptiveFont GuiAdaptiveColor

jgehrig:
Inspired by Pull Request #517

This adds the commands `:GuiAdaptiveFont` and `:GuiAdaptiveColor` which each make the Qt GUI respond to the current Neovim theme.

Specific logic was added to `:GuiPopupmenu` uses the `Pmenu` highlight group.

**Screenshot:**
GuiAdaptiveFontColor

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -7,10 +7,9 @@
 
 namespace NeovimQt {
 
-MainWindow::MainWindow(NeovimConnector *c, ShellOptions opts, QWidget *parent)
-:QMainWindow(parent), m_nvim(0), m_errorWidget(0), m_shell(0),
-	m_delayedShow(DelayedShow::Disabled), m_tabline(0), m_tabline_bar(0),
-	m_shell_options(opts), m_neovim_requested_close(false)
+MainWindow::MainWindow(NeovimConnector* c, ShellOptions opts, QWidget* parent)
+	: QMainWindow(parent)
+	, m_shell_options(opts)
 {
 	m_errorWidget = new ErrorWidget();
 	m_stack.addWidget(m_errorWidget);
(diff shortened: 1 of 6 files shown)


jgehrig:
There is still some small work left to do...

- Windows Testing
- MacOS Testing
- General code review/cleanup
- Documentation in nvim_gui_shim

Also, should this feature be ON by default?

I think many users dislike `GuiPopupmenu` and `GuiTabline` because the Qt theming does not line up with their Neovim theme. On Windows this is especially noticeable because the default font is very small.

jgehrig:
@equalsraf 

The CircleCI failure looks related to Pull #716 and the test coverage report. Any ideas what is happening here?

jgehrig:
**MacOS Screenshot:**
image

It looks like the `QPalette` theming for certain UI elements is ignored on MacOS/Windows. It looks like this is determined by the implementation details of the active Qt style.

A compatible style could be forced `QStyleFactory::create("Fusion")` across all platforms when `:GuiAdaptiveColor 1`. Alternatively, we could allow the user to specify a Qt Theme.

Unfortunately, each platform will required settings for an optimal 'out-of-the-box' experience... I need to think about this more.

equalsraf:
I have no idea what the CI error is about, still searching.

> Unfortunately, each platform will required settings for an optimal 'out-of-the-box' experience... I need to think about this more.

I suspect runtime Qt stylesheets might be the only way to work around the platform issues. 
Maybe it is easier to generate some qss when color font changes and apply it to the main window.

jgehrig:
I haven't found a way to resolve the QPalette or QStyleSheet native style themeing on MacOS/Windows.

Instead, I have added to option to override the default styling, `:GuiAdaptiveStyle Fusion`. A non-default theme will accept QPallette themeing. Even with the native widgets, a `colorscheme` could be carefully crafted to work with the native UI.

We won't be able to turn this on-by-default unfortunately, but I think that is fine.

Issue #723, opened 2020-07-14
Ligatures not worked?


RealXIHNIK:
Ligatures don't work

My Font: Fira Code
Program: neovim-qt
Linux Mint

---

Path:
~/.config/nvim/ginit.vim

Contains:
GuiFont FiraCode NF:h12

[Fira Code Regular Nerd Font Complete Mono Windows Compatible.otf.zip](https://github.com/equalsraf/neovim-qt/files/4921956/Fira.Code.Regular.Nerd.Font.Complete.Mono.Windows.Compatible.otf.zip)

Quetion: What i need do to make ligatures work in neovim-qt?

jgehrig:
This is a known issue, see Pull Request #470

It is on my radar, hopefully we can figure this out by the next release.

jgehrig:
@RealXIHNIK Ligatures do not currently work.

That Pull Request is the code required to make Ligatures work with neovim-qt. You'd need to build that code....

However, there are some outstanding concerns with font spacing and the code is not currently mergable due to conflicts. Generally it works, but is not officially supported.

jgehrig:
Ligatures are now available in `master`, run `:GuiRenderLigatures 1`.

The feature should be available for release `0.2.17`.

Issue #705, opened 2020-06-16
Underline + Undercurl Rendering Incorrect


jgehrig:
**minimal.vim**
```
filetype on
set spell
set nocompatible
```

**Repro Steps**
1. Run: `nvim-qt -- -u minimal.vim`
3. Set markdown file type: `:set filetype=markdown`
4. Type: `[asdfasdfasdf](asdfasdf)`

Notice the undercurl is not rendered, and the underline is the wrong color.

![image](https://user-images.githubusercontent.com/11207308/84717149-05036680-af43-11ea-870a-b615ec1ff205.png)

jgehrig:
Closed by d0200bcde51d1df6efc8b53e404d23edc38f1b8f.

The fix should be available in release 0.2.17.

Issue #166, opened 2016-08-06
ligatures font support


idealist1508:
It would be fantastic that NeoVim could support some ligatures programming font like [FiraCode](https://github.com/tonsky/FiraCode).

There exists a [patch to VIM](https://groups.google.com/forum/#!topic/vim_dev/0sETSAwe5Wo) to enable this.

equalsraf:
That is something that would have to be done upstream see https://github.com/neovim/neovim/issues/1408

idealist1508:
I thought it is something that could be implemeted only in client gui. 
See https://github.com/neovim/neovim/issues/1408#issuecomment-223696580

equalsraf:
Possible. But would require some large changes in drawing. Need to check further.

- http://doc.qt.io/qt-5/qglyphrun.html

zovt:
This weekend I'll try to take a look at this and see if I can figure something out

equalsraf:
@zovt you might want to start from https://github.com/equalsraf/neovim-qt/pull/141 instead of master

One aspect that is not clear to me is how to determine if two chars form a ligature, or if the fonts support it.

razor-x:
I'd love is this could be a thing. I switch to neovim-gtk for a couple months simply for this feature, but I'm about to abandon that project due to UI lag, flickering, slowness, and general breakage of mappings. Never had any of those issues with neovim-qt.

Is there anyway you can use whatever ligature logic they have from that project and adapt it here?

unclechu:
@razor-x It's probably **offtopic**, about flickering in *neovim-gtk* (I'm using this GUI for about half a year), it have been fixed in master branch, see [this issue](https://github.com/daa84/neovim-gtk/issues/57) by enabling double-buffering with environment variable `NVIM_GTK_DOUBLE_BUFFER` set to `1`.

razor-x:
Thanks for the update. I may have been a bit harsh with my feedback on neovim-gtk. Some issues were plugin related I think. I also noticed a lag when writing with one of my keybinds, but now I suspect this is a lower-level issue unrelated to the GUIs (`nnoremap <silent> <Leader>s :<C-U>update<CR>`).

I did end up switching back to qt out of eventual frustration. The key factor was speed for me: neovim-qt just feels faster for my day to day work.

kierun:
Any update on this one?…

equalsraf:
None on my end.

concatime:
[...]

PR #727, opened 2020-07-29
Add Font Ligature Support :GuiRenderLigatures

jgehrig:
**Issue #166:** Font ligatures support request.
**Issue #723:** Font ligatures support request.
**Issue #705:** Underline + Undercurl rendering incorrect.

**Pull #470:** Change printing to add ligature support

First, let me say... Thanks @Nuclearo!
This work would not have happened without your code as reference. The underlying rendering scheme is still being used.

The general approach here is very similar to Pull #470. We update the buffer on a line-by-line basis, and we draw blocks of similarly styled cells together. However, we now use `QTextLayout` and `QGlyphRun` instead of `drawText`. The new API should provide the extra control needed to adjust the spacing of non-monospace fonts and perform ligature detection.

I have tried to structure this code so there is a known-good "old codepath" for `:GuiRenderLigatures 0`. Keeping the code exactly as-is would require large segments to be copy-pasted, so I have tried to decompose rendering tasks into functions. Each function should have code that is almost identical to segments of the old codepath.

**Algorithm:**
1. ShellWidget modifications now `update()` an entire line.
2. On a given line, we look at the first `Cell`'s style.
3. Cell text is appended until a `Cell` style change is detected.
4. Text layout is calculated via `QTextLayout`, resulting in a `QGlyphRun`.
5. The `QGlyphRun` is a list of glyph indices and positions.
6. Ligature are detected from differences in the glyph index of individual characters vs the glyph run.
7. The glyph run values are modified where necessary (cursor on glyph, position, etc)
8. Render the QGlyphRun
9. Back to Step 3 until the line is fully rendered.

**Qt Documentation:**
[QTextLayout](https://doc.qt.io/qt-5/qtextlayout.html)
[QGlyphRun](https://doc.qt.io/qt-5/qglyphrun.html)


jgehrig:
**Not ready for merge yet.**
I am Posting this early because I assume users will be interested.

There are code cleanup, `update()` performance improvements, and benchmarking still left to do. It would also be nice to have some `ShellWidget` test coverage given such extensive modification.

I may work on integrating this with #690...

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -233,6 +233,7 @@ endfunction
 
 command! -nargs=1 GuiScrollBar call s:GuiScrollBar(<args>)
 
+<<<<<<< HEAD
```
equalsraf, line 236: Watch out, seems like a merge leftover.

jgehrig:
@equalsraf

I think this one is ready for merge. Let me know if you have any final comments.

The `GuiRenderLigatures 0` state should be very safe, it uses the same logic (refactored) as today.

For `GuiRenderLigatures 1` I am confident in the basic functionality and not aware of any issues. Time for an initial release.

As a side note, the [Clazy Pipeline](https://dev.azure.com/equalsraf/neovim-qt/_build/results?buildId=52&view=logs&jobId=494e7640-935e-5c68-eff6-35d5e173f87c&j=494e7640-935e-5c68-eff6-35d5e173f87c&t=9dd5c925-984f-56cf-8a94-43d4987e307a) already caught a subtle Qt issue :)

equalsraf:
Looking glorious, go forth and press the merge button.

> As a side note, the Clazy Pipeline already caught a subtle Qt issue :)

ooops :D

Issue #733, opened 2020-08-23
Color emoji


nu8:
With regular Vim:

https://github.com/vim/vim-win32-installer/releases

I can `set renderoptions=type:directx`, and it will produce color emoji:

>![Untitled](https://user-images.githubusercontent.com/61808913/90966346-1e6ee480-e497-11ea-8358-fc8391622869.png)

is this possible with Neovim?

jgehrig:
Yes, this should be supported, but you may find rendering issues.

Your font must support color emojis. My default font `Liberation Mono` does not render emojis in color.
![image](https://user-images.githubusercontent.com/11207308/91664329-753e7480-eabc-11ea-9ce9-1bbd76130d23.png)

With the latest `master` you can provide a `guifontwide` list. This will change the wide-character font.

Here is the same text rendered with `set guifontwide=Noto\ Color\ Emoji:h14`:
![image](https://user-images.githubusercontent.com/11207308/91664370-dbc39280-eabc-11ea-9cc3-55e3066e0983.png)

I don't think either feature is widely used. Please let me know if you see any rendering issues :)

nu8:
With this file:

~~~
C:\Users\Steven\AppData\Local\nvim\init.vim
~~~

As:

~~~
set guifontwide=Segoe\ UI\ Emoji:h14:
~~~

doesnt seem to make any difference. Note I am using this file:

https://github.com/neovim/neovim/releases/download/nightly/nvim-win64.zip

which has date of 2020-08-30.

jgehrig:
> https://github.com/neovim/neovim/releases/download/nightly/nvim-win64.zip

That is a nightly neovim release, but it does not contain all of the changes in `master` for neovim-qt.

You should be able to confirm this with `nvim-qt --version`, which I'm guessing will show `v0.2.16.1`.

Here is a download link for the latest neovim-qt commit (available a few months):
https://ci.appveyor.com/project/equalsraf/neovim-qt/build/job/g4lbd7a2kw3jd0td/artifacts

You can find these Windows builds clicking the AppVeyor details for any recent commit.

nu8:
Ok tried again with NVIM-QT v0.2.17.9999 (and previous `init.vim`), still seems
to be black and white.

jgehrig:
It works for me on Windows 10 with or without `guifontwide`.

My default font is `Consolas:h11`, and I can set a secondary font to `Segoe\ UI\ Emoji:h14`.

[...]

PR #736, opened 2020-08-30
Bump AppVeyor to Qt 5.15.0

jgehrig:
The AppVeyor Qt version used to generate Windows builds is old.

Update to the latest available version of Qt, 5.15.0.


jgehrig:
Issue #733 Color Emojis are only rendered in recent versions of Qt.

Not working in 5.11.3. Tested working in 5.13.0.

jgehrig:
@equalsraf

It looks like the Qt version bump causes an interesting bug with `windeployqt`. I reproduced on AppVeyor and one of my Windows machines. Error message `Unable to find the platform plugin.` during build.

It looks like an instance of this:
https://bugreports.qt.io/browse/QTBUG-62103

I think the fix (cc7eb13) is to remove `--release`, `--debug`, etc. However, I am not very familiar with Qt + Windows packaging.

Perhaps we want to scope this fix to only the impacted Qt releases? Any thoughts? Do you see any issues with the workaround?

equalsraf:
The story for debug/release libraries in msvc really is very different from mingw, I was not even aware the options would have an effect for mingw builds.

Instead of outright removing this, we can just disable this for the MinGW generator e.g. https://github.com/equalsraf/neovim-qt/commit/b769ecc944886f85bb9cab2af825dce10e71c88f, I expect them to still be needed for MSVC debug builds.

I took the mingw appveyor build and they worked without problems, so i think we are good on that front.

jgehrig:
Sounds good. Your change b769ecc looks like a better approach. I will cherry pick it into the PR and handle the merge.

I will also file a bug for tracking. The option may require a re-enable or re-address if Qt fixes this in the future.

Issue #411, opened 2018-05-16
[future request] I need to make neovim-qt window transparent, just like wimproved.vim.


augustblizzard:
I need a plugin or config to make neovim-qt window transparent just like wimproved.vim(just need WSetAlpha)

adrianrocamora:
It would be nice to have

sentavian:
It is possible to do this with third party apps / utilities / settings depending on your OS etc(although still nice to have built in). A missing killer feature, however, would be setting only the background as transparent or setting a background image.

matheusfillipe:
> It is possible to do this with third party apps / utilities / settings depending on your OS etc(although still nice to have built in). A missing killer feature, however, would be setting only the background as transparent or setting a background image.

Do you mean it is posible to set true transparency and if so I would like to
know how, or do you mean managing the compositor and having both the background
and the text with opacity?

augustblizzard:
![image](https://user-images.githubusercontent.com/11937856/90994780-71a67b80-e5ec-11ea-9acd-80c434cc7571.png)

just like image that I uploaded.

equalsraf:
Opened a CR for this https://github.com/equalsraf/neovim-qt/pull/737, a GuiWindowOpacity command sets the opacity in the range 0.0 - 1.0. Can you test if it works for you?

augustblizzard:
sure, I'll test when I'm free of work.

Atharva-k:
does this work now ?

jgehrig:
@Atharva-k 

Yes, this should work. However, it is only available in the `master` branch.

`:GuiWindowOpacity 0.7`

It will be added to the next release, `0.2.17`.

matheusfillipe:
Btw this is not real transparency. Real transparency would be background only transparency. Having the text transparent is not ideal. I hope this can be addressed in the future, would be cool.

jgehrig:
The request here was full window transparency, like `wimproved.vim` in the screenshot.

I agree, that would be a cool improvement. I'm not sure if it is (easily) possible with Qt. If anyone wants to take a look at implementing such a feature, `QGraphicsEffect` is probably a good starting point.

thePHTest:
What is the cycle for new releases? I would love to have this available.

PR #737, opened 2020-09-05
Add GuiWindowOpacity command

equalsraf:
Directly maps to QWidget::setWindowOpacity().
Takes a double as an argument, the valid range is 0.0 to 1.0.

I have tested this in windows and it works, I have not checked X11 but I assume it depends on a working compositor.

ref https://github.com/equalsraf/neovim-qt/issues/411

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -107,6 +107,8 @@ void MainWindow::init(NeovimConnector *c)
 			this, &MainWindow::neovimFullScreen);
 	connect(m_shell, &Shell::neovimGuiCloseRequest,
 			this, &MainWindow::neovimGuiCloseRequest);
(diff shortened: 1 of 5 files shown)


jgehrig:
This also looks somewhat related to #249.

It works for me on Linux + X11 (KWin):
image

It might take me a while, but I could test Windows/Mac if needed...

I wonder if we should we write the opacity value to one of the `QSettings` to prevent startup flicker?
- `QSettings settings{ "nvim-qt", "window-geometry" }`
- `QSettings settings("nvim-qt", "nvim-qt")`

I should do this for a few other options like `GuiTabline`, `GuiRenderLigatures`, `GuiFont`, etc... It looks much cleaner when we can set these values before rendering the `MainWindow`.

Issue #684, opened 2020-04-22
Buffers are not reflected as separate tabs in GUI


f-squirrel:
**How to reproduce:**
1. Open `neovim-qt`
2. Open two files in separate buffers: `e: file1.txt<CR>`, `e: file2.txt`<CR>

**Actual:**
There is only one tab despite two buffers were open.
**Expected:**
Every buffer to be in a separate tab
**Note:**
There is a possibility to switch between the buffers with regular commands as `:bprev` and `:bnext`.

**Additional information:**
```
bin/nvim-qt.app/Contents/MacOS/nvim-qt --version
NVIM-QT v0.2.16.9999
Build type: Release
Compilation: -Wall -Wextra -Wno-unused-parameter -std=c++11 -Wfatal-errors
Qt Version: 5.14.1
Environment:
  nvim: nvim
  args: --cmd let &rtp.=',/Users/user/neovim-qt/build/bin/nvim-qt.app/Contents/MacOS/../Resources/runtime' --cmd set termguicolors
  runtime: /Users/user/neovim-qt/build/bin/nvim-qt.app/Contents/MacOS/../Resources/runtime

NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5
[...]

PR #739, opened 2020-09-06
Improved GuiTabline show buffers

jgehrig:
**Issue #684:** Buffers as GUI tabs

The `GuiTabline` feature does not behave as many users expect it to. Many users expect buffers to display as gui-tabs.

These changes attempt to mimic the default behavior of plugins like `vim-airline`. When only 1 tab view is available, each buffer is displayed as a gui-tab. Once multiple tabs are available, gui-tabs are drawn to represent both available buffers and tabs.

image
image
image

API updates required: neovim/neovim#12481

### DIFF
--- a/contrib/clang-format.txt
+++ b/contrib/clang-format.txt
@@ -1,6 +1,6 @@
 ﻿---
 AccessModifierOffset: '-4'
-AlignAfterOpenBracket: DontAlign
+AlignAfterOpenBracket: AlwaysBreak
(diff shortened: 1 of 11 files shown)


jgehrig:
Not ready yet... The API hasn't been merged yet.

I also need to add some `neovim-qt` side checks to ensure everything works when the new API is unavailable.

I would also like to make the display mode configurable:
- No Tabs
- Buffers + Tabs (see above)
- Tabs Only (existing behavior)

PR #740, opened 2020-09-09
Enable Link Time Optimization in Release

jgehrig:
We have lots of generated but unused code, and `nvim-qt` would benefit from Link Time Optimization.

On Linux, my binary is reduced 10-15% (~300KB). I haven't done any performance profiling...

This requires a CMake minimum version bumped to `3.9` (I think). See:
https://cmake.org/cmake/help/latest/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html
https://cmake.org/cmake/help/latest/policy/CMP0069.htm

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -78,6 +78,19 @@ if(APPLE)
 #		COMMENT "Bundling Qt libraries")
 endif()
 
+# Link Time Optimization, especially useful for unused auto-generated code.
+# On Windows, MinGW may not support LTO.
+# See https://github.com/neovim/neovim/pull/8654
+if(POLICY CMP0069 AND NOT MINGW)
+	cmake_policy(SET CMP0069 NEW)
+
+	include(CheckIPOSupported)
+	check_ipo_supported(RESULT IPO_SUPPORTED)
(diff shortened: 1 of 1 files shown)


jgehrig:
Whoops, I still need to add or remove this comment:
`# FIXME Comment!`

justinmk:
I think you can check `if (POLICY CMP0069)` and that will skip on cmake older than 3.9?  https://github.com/neovim/neovim/commit/c8e7a447c5694e7e2a58f2b112530d896efa30ba#diff-7abbb8315fea3e74ba20ce24ac297519R402

jgehrig:
@justinmk

Thanks, that is a much better solution! I will incorporate the snippet...

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 3.9)
```
justinmk, line 1: don't need to require newer cmake , with the policy check now in place

jgehrig: Ack, my plan was to revert this during the rebase.

I still need to validate the Windows MinGW build. The PR you linked mentions issues with LTO + MinGW.

Issue #639, opened 2020-02-03
Bump Minimum Qt Version from 5.2


jgehrig:
Any thoughts on bumping the minimum Qt version from 5.2?

It would be nice to get rid of `#ifdef`s like the one from Pull #630  for old Qt versions.

There are several other improvements we can make:
- Get rid of `Q_ENUM` vs `Q_ENUMS`
- Fix compiler warnings for `QRegion::rects()` and other deprecations.

Qt 5.8 seems like a good minimum version candidate. It is old enough to be widely available, contains useful performance improvements, and is higher than many of the `#ifdef` checks.

equalsraf:
Preliminaries, trying to track down used versions

- Ubuntu xenial seems to run 5.5 https://packages.ubuntu.com/search?keywords=qt&searchon=names&suite=xenial&section=all
- OpenBSD 6.3 seems to be 5.9
- Appveyor goes from Qt 5.3 to 5.14 https://www.appveyor.com/docs/windows-images-software/#qt
- Void is 5.13 https://voidlinux.org/packages/
- anyone else wants to add more here?

Q_ENUMS was introduced in 5.5 right. I'm tempted to stick with that. What version is required for `QRegion::rects`?

jgehrig:
`QRegion::rects()` has been deprecated in favor of `::begin()/::end()`, which occurred in Qt 5.8.

- https://doc.qt.io/qt-5/qregion-obsolete.html#rects
- https://doc.qt.io/qt-5/qregion.html#begin

Yes, `Q_ENUM` was introduced in Qt5.5 from what I can tell.

#630 has one preprocessor check for >=Qt5.5, and another for >=Qt5.7

There are also some warnings around `QFontMetrics::width()` vs `QFontMetricsF::horizontalAdvance()`, which require Qt 5.11. However, I think Qt5.11 is too aggressive as a minimum version.

equalsraf:
Lets stick with 5.5. We can live with one ifdef, and the enums are the most significant one.

The deprecation warnings are annoying but we will have to deal with them a bit longer.

PR #742, opened 2020-09-14
Add Clazy (Clang Qt linter) to CI builds

jgehrig:
**Work in Progress**

The KDE project has built a set of `clang-tidy` style rules for Qt: https://github.com/KDE/clazy

We can take advantage of this to improve the compile-time safety and performance of neovim-qt.

Eventually, I would like to add support for clang-tidy too, but adding it here would make this already large Pull even bigger.

**Azure Pipelines:**
I did this work in Azure Pipelines. It is my favorite CI since it supports all platforms in one place (MacOS, Linux, Window), and is free for OpenSource projects. It was mostly to make these changes easier, I was running the CI loop directly on my repo.

Since the work is done, perhaps it is worth checking the YAML into `contrib/azure-pipelines.yml`?

I'm not sure if it switching over, since or current CI stack is working fine...

PR #744, opened 2020-09-15
Bump Minimum Qt Version 5.8

jgehrig:
**#Issue #639:** Bump minimum Qt Version

We should bump the minimum Qt version, several APIs in key areas are being deprecated in recent releases. For now 5.8 seems like a good balance. It is available almost everywhere and contains `QRegion` performance along with `QScrollEvent` modifications that are necessary to support most high-resolution touch pads.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     machine:
-      image: ubuntu-1604:201903-01
+      image: ubuntu-2004:202008-01
     working_directory: ~/project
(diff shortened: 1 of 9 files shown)


jgehrig:
The `endl` to `Qt::endl` change is currently invalid. We need a wrapper, or to continue using `endl`.

jgehrig:
Required for Pull #742.

As written, we need Qt >= 5.7 for `qAsConst` for the `clazy-range-loop` warning fix.

PR #754, opened 2020-09-22
Enable codecov coverage reports

equalsraf:
We are seeing some issues with coveralls, trying out codecov as an alternative.


equalsraf:
I'm not seeing an output in the PR checks but this might be normal since there was no previous coverage in earlier commits. We can see the coverage report for the branch though https://codecov.io/gh/equalsraf/neovim-qt/branch/tb-codecov

Issue #760, opened 2020-09-25
Need way to set custom WM_CLASS on startup


cvincent:
Currently there doesn't appear to be a way to assign a custom WM_CLASS on startup. This is useful for doing interesting things with window managers, such as setting up specific layouts automatically. With terminal NeoVim, I could do this using my terminal emulator, Kitty:

```
kitty --name dotfiles --title dotfiles bash -c "cd ~/dotfiles && nvim -S Session.vim"
```

Unfortunately there does not appear to be a way to do this with `nvim-qt`.

equalsraf:
Does `nvim-qt --name thewmclassyouwant` work?

Based on [this](https://bugreports.qt.io/browse/QTBUG-29396?focusedCommentId=208874&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel) this should be available in qt 5.1.

jgehrig:
For me, `--name` is an unrecognized argument:
``` bash
$ nvim-qt --name FooBar
nvim-qt: Unknown option 'name'.
```

From latest master and  built with Qt 5.15.0.

equalsraf:
```
$ nvim-qt --name birubirubiru
```

Then used xprop to get the wm class

```
$ xprop
XdndAware(ATOM) = BITMAP
WM_NAME(STRING) = "[No Name] - NVIM"
_NET_WM_NAME(UTF8_STRING) = "[No Name] - NVIM — Neovim"
_MOTIF_WM_HINTS(_MOTIF_WM_HINTS) = 0x3, 0x3e, 0x7e, 0x0, 0x0
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_NORMAL
_XEMBED_INFO(_XEMBED_INFO) = 0x0, 0x1
WM_CLIENT_LEADER(WINDOW): window id # 0x2600008
WM_HINTS(WM_HINTS):
                Client accepts input or input focus: True
                window id # of group leader: 0x2600008

_NET_WM_PID(CARDINAL) = 32017
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 39845895
WM_CLASS(STRING) = "birubirubiru", "nvim-qt"
WM_PROTOCOLS(ATOM): protocols  WM_DELETE_WINDOW, WM_TAKE_FOCUS, _NET_WM_PING, _NET_WM_SYNC_REQUEST
[...]

PR #761, opened 2020-09-25
Add commandline option --class to override default WM_CLASS

cvincent:
I went ahead and came up with a patch. Haven't written C++ in ages!

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -163,9 +163,9 @@ void App::processCommandlineOptions(QCommandLineParser& parser, QStringList argu
 				QCoreApplication::translate("main", "Error if nvim does not responde after count milliseconds"),
 				QCoreApplication::translate("main", "ms"),
 				"20000"));
-	parser.addOption(QCommandLineOption("class",
-				QCoreApplication::translate("main", "set custom WM_CLASS"),
-				QCoreApplication::translate("main", "WM_CLASS"),
+	parser.addOption(QCommandLineOption("name",
+				QCoreApplication::translate("main", "set custom application name"),
+				QCoreApplication::translate("main", "app_name"),
 				"nvim-qt"));
 
 	// Some platforms use --qwindowgeometry, while other platforms use the --geometry.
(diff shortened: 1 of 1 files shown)


cvincent:
Sorry, this is in reference to https://github.com/equalsraf/neovim-qt/issues/760

src/gui/app.cpp
```diff
@@ -159,6 +163,10 @@ void App::processCommandlineOptions(QCommandLineParser& parser, QStringList argu
 				QCoreApplication::translate("main", "Error if nvim does not responde after count milliseconds"),
 				QCoreApplication::translate("main", "ms"),
 				"20000"));
+	parser.addOption(QCommandLineOption("class",
```
jgehrig, line 166: I wouldn't use `WM_CLASS` in the description; this is only correct for Linux + X11. Also capitalization.

Perhaps something more generic: `Set a custom application name`?

jgehrig, line 166: I would use `--name` as the argument, just like your kitty example.

What this argument really does is change the application name. `WM_CLASS` is a Linux/X11 specific implementation detail.

jgehrig:
Thanks for the Pull Request!

Looking at this property in the documentation suggests it may have side-effects:
https://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop

We should test on all platforms and investigate what impacts changing `applicationName` may have.

src/gui/app.cpp
cvincent: Yeah there doesn't appear to be a standard. For example, with `qutebrowser` you set it like `--qt-arg name <NAME>`. But `--name` seems reasonable, and matches QT's terminology.

cvincent: Updated argument name and description.

src/gui/app.cpp
```diff
@@ -97,6 +97,10 @@ void App::showUi() noexcept
 {
 	ShellOptions opts{ GetShellOptionsFromQSettings() };
 
+	if (m_parser.isSet("name")) {
+		setApplicationName(m_parser.value("name"));
```
equalsraf, line 101: Wouldn't this cause QSettings to also be affected? This would mean config files would use the wrong paths.

jgehrig: I think it would only affect the default CTOR. Since we call `QSettings{ "nvim-qt", "nvim-qt")` explicitly it should not have any impact.

Issue #757, opened 2020-09-24
NVR should activate the MainWindow


jgehrig:
From Issue #734.

@Shatur95 Wrote:

```
export NVIM_LISTEN_ADDRESS=127.0.0.1:1234
nvim --headless &
nvim-qt --server $NVIM_LISTEN_ADDRESS
nvr File1.txt
```

But after last step neovim instance will not be active. I mean, if it is minimized it, it will remain minimized. Usually, text editors, when opening a new file with an instance already running, make it active. Is it possible to make Neovim Qt active on the last action?

equalsraf:
Sounds like you want to issue a call to `call GuiForeground()` as part of your command.

I'm not familiar with nvr, so there might be a better way, this looks like a start.

```
nvr --servername /tmp/nvim5ANTQk/0 -c 'call GuiForeground()' /tmp/test.txt
```

Shatur95:
I tried id, no effect :(

equalsraf:
Strange. This kind of behaviour depends a bit on the window manager, so we may have to try a couple of things. Which window manager are you using?

What about `call GuiWindowMaximized(1)`?

Shatur95:
> Strange. This kind of behaviour depends a bit on the window manager, so we may have to try a couple of things. Which window manager are you using?

KWin on KDE Plasma. I have another application in Qt that shows correctly. [Here](https://github.com/crow-translate/crow-translate/blob/b378e5dcc7f12d9208568b2f2e1ae39766d71380/src/mainwindow.cpp#L242) is the code.

> What about call GuiWindowMaximized(1)?

Still no effect :(

equalsraf:
Strange, are we failing to do it properly here? https://github.com/equalsraf/neovim-qt/blob/e5841d267229c9066861f453dafda50ebfaeac19/src/gui/shell.cpp#L753

My system is a bit hard to test this, I cant raise the window in my window manager, but I do get a window alert notification.

equalsraf:
Maybe this change helps, https://github.com/equalsraf/neovim-qt/commit/48c490bafbb4a5db51f625a58219bd6119f71942 but in my case i see no difference. Can you build that change and let us know?

Shatur95:
Thanks, I tried, but still not works. Maybe try the following?

```cpp
    setWindowState(windowState() & ~Qt::WindowMinimized);
    show();
    activateWindow();
```

This will always make window active.

PR #763, opened 2020-09-26
Fix GuiForeground()

equalsraf:
There were two issues with the function to bring to foreground:

1. unlike similar functions, when working with a widget there was not
   signal to the parent window
2. the current code doesn't seem to work with some window managers where
   the window remains minimized

-----

I dont see a change in my system, but see https://github.com/equalsraf/neovim-qt/issues/757

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -750,8 +750,13 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 		if (guiEvName == "Font") {
 			handleGuiFontFunction(args);
 		} else if (guiEvName == "Foreground" && args.size() == 1) {
-			activateWindow();
-			raise();
+			if (isWindow()) {
+				setWindowState(windowState() & ~Qt::WindowMinimized);
+				show();
+				activateWindow();
+			} else {
+				emit neovimForeground();
+			}
 		} else if (guiEvName == "WindowMaximized" && args.size() == 2) {
 			if (isWindow()) {
 				setWindowState(variant_not_zero(args.at(1)) ?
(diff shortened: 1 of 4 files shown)


Shatur95:
Thank you, now works as expected!

src/gui/shell.cpp
```diff
@@ -750,8 +750,12 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 		if (guiEvName == "Font") {
 			handleGuiFontFunction(args);
 		} else if (guiEvName == "Foreground" && args.size() == 1) {
-			activateWindow();
-			raise();
+			if (isWindow()) {
+				activateWindow();
+				raise();
```
Shatur95, line 755: Should we use the same approach as in `MainWindow::neovimForeground()`?

equalsraf: I updated this to use the same approach in both.

Shatur95: Thanks!

Shatur95:
Is it ready to be merged?

jgehrig:
@Shatur95

> Thank you, now works as expected!

I can't get it working... But perhaps I'm missing a step something? What did you do for validation?

If we can confirm this works, I comfortable merging the change.

jgehrig:
@equalsraf

A friendly ping :smile:

Shatur95:
> I use the following script for `nvr`:

```bash
#!/bin/bash

SOCKET=/tmp/nvimsocket

if [ -S $SOCKET ]; then
    nvr -c 'call GuiForeground()' --remote-wait -s "$@"
else
    $NVR_CMD --nofork -- --listen $SOCKET "$@"
fi
```

Where `$NVR_CMD` defined as `neovim-qt`. I use this branch for my daily use.

equalsraf:
> @Shatur95
> 
> > Thank you, now works as expected!
> 
> I can't get it working... But perhaps I'm missing a step something? What did you do for validation?
> 
> If we can confirm this works, I comfortable merging the change.

Sorry was away.

To me this worked before and after the change, it seems to be window manager specific. Merging.

Issue #771, opened 2020-10-01
GuiRenderLigatures Issues


jgehrig:
@kierun

Creating a bug here so as to not spam the large thread.

From #166:
>Ah, I see. The runtime was wrong, installed and I can see ligatures.
>
>screenshot
>
>I mean, they are horribly broken but that could be my font.
>
>screenshot1
>
>Which it is.
>
>However, using the mouse now causes neovim to crash… Should I open a new bug for that?

Yes, please provide more details on your `GuiRenderLigatures` issues here.

Sorry for the break, the changes are still experimental.

Please be specific, I'm not sure what each screenshot above correlates to.

What font are you using? Steps leading up to the issues? What specific actions result in a crash?

Here is the feature on my machine:
![image](https://user-images.githubusercontent.com/11207308/94824112-a4554800-03d2-11eb-821c-aa1aee32e30f.png)

kierun:
Another oddness: the font width appears to be variable!

![Peek 2020-10-01 16-22](https://user-images.githubusercontent.com/3079575/94829345-522c1b00-0402-11eb-849a-6b61c88cb7d7.gif)

This is using a PrgamataPro patched (with Nerd Fonts) font. At the end, I did a left mouse click.

jgehrig:
> Font is Monoid Nerd Font. However, any mouse input (left, middle, and right click) gets me this:

The input issues are unrelated to `GuiRenderLigatures`. A bisect suggests 3a59d6bd6e24a0935272a87e4300f97472e40d13 is to blame.

I will file a new bug (or pull request) shortly...

jgehrig:
> Another oddness: the font width appears to be variable!

Interesting, is the font you are using well-formed? Is there a space before b?

The raw characters like `==` and their equivalent ligature must be the same width.

Here is my machine with `Fira Code`:
![Cursor_Movement_Ligatures](https://user-images.githubusercontent.com/11207308/94831285-8a1f6800-03da-11eb-8832-4857ed279379.gif)

File Content:
```
a == b
a != b
a -> b
a => b
```

Your machine looks okay for `Monoid`. Does the issue disappear when you change fonts?

jgehrig:
What setting causes the highlight lines to appear vertically and horizontally outward from the cursor?

The feature has some interesting implications. `GuiRenderLigatures 1` renders entire lines and `GuiRenderLigatures 0` renders individual cells. With your config, cursor movements trigger an entire screen redraw... Ouch!

[...]

PR #772, opened 2020-10-01
Fix convertMouse QString::arg issue

jgehrig:
Commit 3a59d6bd6e24a0935272a87e4300f97472e40d13 contains an error.

The following does not work as expected:
```C++
.arg(pos.x(), pos.y()
```

Odd.. Seems like this should work fine. However, the implicit conversion from `int` to `QString` does not work properly.

Use an alternative API.

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -116,8 +116,13 @@ QString convertMouse(
 		return {};
 	}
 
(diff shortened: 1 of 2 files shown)


jgehrig:
Found by @kierun

See Issue #771.

kierun:
Thank you!

kierun:
That has fixed it. Thank you so much!

PR #776, opened 2020-10-15
Update README.md

nilsmartel:
It might be personal preference, but I think highlighting this keyword increases readability and newcomers have an easier start with this


jgehrig:
@nilsmartel

Looks reasonable to me. I added a small update to do the same for `NVIM_QT_RUNTIME` too. Thanks for the Pull Request!

PR #779, opened 2020-10-17
Fix Clazy Warning: clazy-overloaded-signal

jgehrig:
Qt signals should not be overloaded, change names.


jgehrig:
This is broken... And also not provided by clazy 1.6 included in ubuntu-20.04.

Abandoning. Maybe we will re-implement this correctly when ubuntu updates clazy.

PR #781, opened 2020-10-22
WIP: Fix GuiRenderLigatures for OperatorMono

jgehrig:
**Issue #771:** Operator Mono Font Not Working

The Operator Mono font uses a different encoding scheme than Fira Code. For Fira Code, each character maps has a glyph in QGlyphRun. For Operator Mono, a ligature maps to a single glyph. I did not expect or observe this, so the logic was not tolerant of such an encoding scheme.

Operator Mono also has some monospace positioning irregularities: each character is not exactly the same width. Ligatures may may be slightly larger or smaller than their atomic parts/characters.


jgehrig:
This still needs some code cleanup.

I am not aware of any outstanding issues, but I have only done limited testing.

Issue #784, opened 2020-10-29
crash when scrolling using mouse


jdhao:
I have experienced a lot of crash lately with nvim-qt on Windows.

nvim-qt version: 0.2.17.9999
nvim version:  v0.5.0-705-g9f704c88a

Steps to reproduce:

+ Open nvim-qt: `nvim-qt -- -u NORC some_file.py`
+ Set it to use mouse: `:set mouse=a`
+ Scroll the screen using middle mouse button right after the above step (do not click using the mouse or anything).

Then nvim-qt crashed with the message:

> The Neovim process has crashed.

If we click the window, then start scroll using mouse. It may not crash. I don't know why.

jgehrig:
This is very likely related to bad mouse handling in `shell.cpp`.

You can see an example of the bad input in #696:
```
QWheelEvent
   QWheelEvent(Qt::ScrollMomentum, pixelDelta=QPoint(58,0), angleDelta=QPoint(116,0))
   "<ScrollWheelRight><86.25,28.8667>"
```

I think I broke this with a combination of Qt Version Bump, 3a59d6bd6e24a0935272a87e4300f97472e40d13, and 6e309981c3fdedd2a09dcfd8c203c3667f54bf1f.

PR #785, opened 2020-11-04
Fix touchpad scrolling

matijaskala:
use trunc() to round floating point numbers towards zero

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -1315,45 +1315,25 @@ void Shell::setCursorFromBusyState() noexcept
 	}
 }
 
-static int normalize(float x)
-{
-	if (x == 0.0f) {
-		return 0;
-	}
-	if (x < 0) {
-		return -1;
-	}
-	return 1;
-}
-
 void Shell::wheelEvent(QWheelEvent *ev)
 {
 	if (!m_attached || !m_mouseEnabled) {
 		return;
 	}
 
-	QPointF scroll_delta_f{ ev->angleDelta() / QWheelEvent::DefaultDeltasPerStep };
+	// Issue 785: {Touchpad scrolling only works if the user scrolls really fast, scroll_delta_f can be < 1 for slow scroll events}
+	QPointF scroll_delta_f{ QPointF{ ev->angleDelta() } / QWheelEvent::DefaultDeltasPerStep };
 
 	if (ev->inverted()) {
 		scroll_delta_f = -scroll_delta_f;
 	}
 
-	// Reset scroll remainder if we change scroll direction;
-	const int direction_x = normalize(scroll_delta_f.x());
-	if (direction_x != m_scroll_last_direction.x()) {
-		m_scroll_remainder.setX(0.0);
-		m_scroll_last_direction.setX(direction_x);
-	}
-	const int direction_y = normalize(scroll_delta_f.y());
-	if (direction_y != m_scroll_last_direction.y()) {
-		m_scroll_remainder.setY(0.0);
-		m_scroll_last_direction.setY(direction_y);
-	}
+	// Don't reset scroll remainder when we change scroll direction
 
 	// Scroll distance considering previous events
 	const QPointF total_delta = scroll_delta_f + m_scroll_remainder;
-	const QPointF scroll_delta(std::floor(total_delta.x()),
-			std::floor(total_delta.y()));
+	const QPointF scroll_delta(std::trunc(total_delta.x()),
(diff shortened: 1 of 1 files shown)


src/gui/shell.cpp
```diff
@@ -1332,28 +1332,20 @@ void Shell::wheelEvent(QWheelEvent *ev)
...
+	// Convert angleDelta() to QPointF then divide it with DefaultDeltasPerStep
+	QPointF scroll_delta_f{ ev->angleDelta() };
+	scroll_delta_f /= QWheelEvent::DefaultDeltasPerStep;
 
 	if (ev->inverted()) {
 		scroll_delta_f = -scroll_delta_f;
 	}
 
-	// Reset scroll remainder if we change scroll direction;
-	const int direction_x = normalize(scroll_delta_f.x());
```
jgehrig, line 1342: Function `normalize` is now unused in your PR. Can you please remove it?

src/gui/shell.cpp
```diff
@@ -1332,28 +1332,20 @@ void Shell::wheelEvent(QWheelEvent *ev)
 		return;
 	}
 
-	QPointF scroll_delta_f{ ev->angleDelta() / QWheelEvent::DefaultDeltasPerStep };
+	// Convert angleDelta() to QPointF then divide it with DefaultDeltasPerStep
+	QPointF scroll_delta_f{ ev->angleDelta() };
+	scroll_delta_f /= QWheelEvent::DefaultDeltasPerStep;
```
jgehrig, line 1337: No need to change this line, right?

This should be functionally equivalent to what existed before.

matijaskala: Previous code divided angleDelta with that number BEFORE converting it from QPoint to QPointF, so there was an undesired rounding.

jgehrig:
Thanks @matijaskala for the PR! :+1: 

Could you quickly explain your fix? Sorry, my ability to test these scenarios is limited; I don't have a high-resolution TrackPad.

A few questions:
- What is wrong with scroll events? A demonstration with `qDebug` output is usually helpful.
- Why are we getting rid of scroll remainder reset on direction change?
- Why `std::trunc` instead of `std::floor`? Why do we need different rounding for negative values?

Everything looks good. A few small comments above. I am comfortable with a merge once I understand the break and fix.

matijaskala:
floor() round -0.1 to -1, trunc() rounds it to 0.
If I scroll 0.9 mm down then 0.01 mm up and keep repeating than, I expect that the program would slowly scroll downwards. If the program resets scroll remainder then it won't scroll anywhere.
angleDelta() is QPoint with integer precision where both coordinates are usually smaller than 120. If we divide an integer below 120 by 120 we get 0, meaning scroll deltas below certain limits will be lost and scrolling will only work if you scroll really fast (or use a normal mouse I presume). By assigning original scrollDelta to scroll_delta_f (which has floating-point precision) and dividing scroll_delta_f with 120, this problem goes away.

src/gui/shell.cpp
jgehrig: Ahhh I see, thanks. Very subtle, `{}` usually catches such issues.

Let's call this out in a comment or write this in a way that clearly indicates the desired types.
`QPointF scroll_delta_f{ QPointF{ ev->angleDelta() } / QWheelEvent::DefaultDeltasPerStep };`?

Alternatively, move `scroll_delta_f` calculation small function + write some tests to prevent regression in the future. Perhaps overkill here.

jgehrig:
Got it, this makes perfect sense to me. :+1: 

> The problem is that currrently scrolling only works if I scroll really fast. If I modify the code to fix that problem, scrolling upwards is still too slow while scrolling downwards is too fast

This is worth quickly mentioning in a comment somewhere. `\\ Issue 785: {Your description, scroll_delta_f can be < 1 for slow scroll events}`. Anywhere in your PR works.

I think @psagers described this issue in Issue #695:
> If I drag slowly (which is more typical), it never moves.

jgehrig:
Merged, thanks for your contribution! :+1:

jdhao:
@jgehrig  does it fix issue https://github.com/equalsraf/neovim-qt/issues/784. Random crash when scrolling the mouse is annoying...

jgehrig:
@jdhao

No, I don't think so. I haven't observed the crashes, but I would guess it is a `float` vs `int`, which still exists.

Sorry, I've been busy, I'll try to get to this soon. I'm working on some changes to add test coverage and a crash fix.

PR #801, opened 2020-12-21
Fix width of vertical bar of cursor on double-width chars

### LINKED ISSUE #800: Vertical bar's width should be the same regardless of char's width
On the double width char:
image

On the single width char:
image

The width of the cursor on double width chars are definitely wider than that on single width chars. Generally (in other editors) vertical bar's width is the same regardless of the char width (while horizontal or block cursor has the same width with the char's so current behavior is OK for them). The cause is in the calculation of the vertical bar size:

https://github.com/equalsraf/neovim-qt/blob/5be7b9ef36dbab0c459a84f4bf4005eae841c626/src/gui/shellwidget/shellwidget.cpp#L134-L137

Here `cursorRect` is from `cellRect` which is the size of entire character, so the double width char's `cellRect` is twice as wide as single width char's:

https://github.com/equalsraf/neovim-qt/blob/5be7b9ef36dbab0c459a84f4bf4005eae841c626/src/gui/shellwidget/shellwidget.cpp#L887-L890

For one of simplest solution we can divide the bar width by char width for the vertical bar.

statiolake:
Closes #800.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -133,7 +133,12 @@ QRect ShellWidget::getNeovimCursorRect(QRect cellRect) noexcept
 
 		case Cursor::Shape::Vertical:
 		{
-			cursorRect.setWidth(cursorRect.width() * m_cursor.GetPercentage() / 100);
+			// Issue #800: for vertical bar cursor, use single cell width
+			// instead of the provided cell rect. Cell rect may have doubled
+			// width, for example, when the cell contains a double width char.
+			// Always using single cell width prevents vertical bar to be a bit
+			// thicker on a double width char.
+			cursorRect.setWidth(m_cellSize.width() * m_cursor.GetPercentage() / 100);
 		}
 		break;
 	}


jgehrig:
@statiolake Nice work finding this bug!

I also noticed this behavior, but assumed it was how things are supposed to work: X% of cursor width -> X% of double-width cell.

Based on observations from `gVim`, you are correct. The statement above is not true for `ver` (vertical bar) cursors.

I believe a simpler code-fix exists:
```
case Cursor::Shape::Vertical:
{
	// Issue #800: Double width characters change the size of cursorRect.
	// Vertical cursors should be based on the cell size only. A double wide
	// character should not have a double wide vertical (insert mode) cursor.
	const QRect cursorRectVertical{ cursorRect.topLeft(), m_cellSize };
	cursorRect.setWidth(cursorRectVertical.width() * m_cursor.GetPercentage() / 100);
}
```

This should be the same calculation and approach, but this shouldn't require any function signature changes. We should avoid adding new function parameters if possible.

Feel free to tweak the snippet above as you see fit, comment or code. Proof of concept with quick testing only.

Can you please fix this without adding `int charWidth` to function calls? Thanks!

jgehrig (changes requested):
Let's avoid adding the `charWidth` parameter. Keeping the code as simple as possible will be better in the long run.

Also, I think only 2 characters widths are possible: single and double. If we absolutely must add this parameter, lets make it something like `bool isDoubleWidth`.

statiolake:
> I believe a simpler code-fix exists:

That's a nice idea, I'll fix it soon. Thank you for pointing it out!

jgehrig:
Two heads are better than one! It improved with each iteration. 

Merged, thanks for the PR :+1:

PR #803, opened 2020-12-21
Beautify undercurl

### LINKED ISSUE #802: Undercurl is a bit random
image

This is current undercurl. It's OK, but I feel this a bit random.

I fixed it for my personal use (https://github.com/statiolake/neovim-qt/commit/78c1129b854ac0bc982c0fc620083b9e9b4d3921) and now undercurl looks like this:

image

I think it is more clean and I'd be happy to open PR if it's worth.

statiolake:
Closes #802.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -242,8 +242,24 @@ static QPainterPath GetUndercurlPath(QRect cellRect) noexcept
 	const QPoint& start{ underline.p1() };
 	const QPoint& end{ underline.p2() };
 
-	QPainterPath path(start);
 	static constexpr int offset[8]{ 1, 0, 0, 1, 1, 2, 2, 2 };
+
+	// Be careful to set correct offset for the starting point of the undercurl
+	// so that it doesn't make discontinuity at the cell boundary.
+	//
+	// Simply using `start` always set the y offset of starting point to zero,
+	// but it is incorrect in most cases; the ending point of the undercurl on
+	// the previous cell has non-zero offset except
+	// offset[previous_undercurl_ending_x % 8] == 0.
+	//
+	// To make undercurl smoothly connected, we need to set correct offset to
+	// the starting point according to the x coordinate of it.
+	//
+	// See also Pull Request #803 (comment):
+	// https://github.com/equalsraf/neovim-qt/pull/803#issuecomment-751166085
+	QPoint undercurlStart{ start.x(), start.y() - offset[start.x() % 8] };
+
+	QPainterPath path{ undercurlStart };
 	for (int i = start.x() + 1; i <= end.x(); i++) {
 		path.lineTo(QPoint(i, start.y() - offset[i%8]));
 	}


jgehrig:
On my machine, I don't notice any rendering differences.

Looking closely at your screenshots, I notice small impulse discontinuities are causing the "wonky" undercurl lines. I suspect you are experiencing a rounding issue.

I'm curious can we resolve this issue with a simple:
` static constexpr float offset[8]{ 1, 0, 0, 1, 1, 2, 2, 2 };`
...
`path.lineTo({i, start.y() - offset[i%8]});` or if required `path.lineTo(QPointF{ i, start.y() - offset[i%8] });`

Also, this would be a good time for `int foo{ 0 }` instead of `int foo = 0`. Let the compiler catch rounding/conversion errors.

I'm inclined to check in your code even if the above works... It is elegant and easier to read than what exists today.

statiolake:
> I'm curious if we can resolve this issue with a simple:

I changed code to what you suggested, and it doesn't fix, but I noticed that the glitch is only at the char boundary - checking by the larger font:

image

Finally I found the below lines are causing the issue.

https://github.com/equalsraf/neovim-qt/blob/9b19e4b4a4424eb199ad631cd93f359c2ffbe30c/src/gui/shellwidget/shellwidget.cpp#L250

This line enforces the start position's y coordinates to be always the same with the underline regardless of x coordinates, which is sometimes incorrect. The issue was fixed by changing to:

```cpp
QPainterPath path{};
static constexpr qreal offset[8]{ 1, 0, 0, 1, 1, 2, 2, 2 };
for (int i = start.x(); i <= end.x(); i++) {
	path.lineTo(QPointF{static_cast<qreal>(i), start.y() - offset[i%8]});
}
```
image

I noticed my fix creates another issue for larger font (*), I'll fix this issue by the above.

(*) For larger font undercurl wave is unexpectedly wide because the wave width is for one period regardless of the single cell's width.
image

jgehrig:
I think you're on right track, but I'm not convinced the current fix is 100% correct either...

Removing the `start` value may fix the issue, but I am concerned we're just masking the symptoms.

Here's my current hypothesis of what's going wrong:
1. We compute the start/end positions as a straight line in `GetUnderline`. (X1, Y1) to (X2, Y2)
2. We're iterating over the path 1 pixel (really a `qreal` canvas position) adding the y-offset value from `offset`
3. When approaching a cell-boundary, there is no guarantee we're at a position evenly divisible by 8.
4. Once we've arrived at X2, we are at an arbitrary position in offset.
5. There is no guarantee each cell starts/stops at the same value -> discontinuity.

Illustration trying to capture this idea, where `width % 8 != 0`:
image

I think we need to guarantee the y-offset starts and ends at the same value, 0 for example.

Perhaps we can modify the period sine wave so that it is evenly divisible by `m_cellSize.width()` and approximately 8?

For `GuiRenderLigatures 1` we could also resolve this by rendering the line all at once.

statiolake:
Yes, I also thought what's causing the discontinuity is exactly what you explained and I still believe the commit resolves the discontinuity issue. (Is there something wrong?)

The code in the above comment

```cpp
QPainterPath path{};
static constexpr qreal offset[8]{ 1, 0, 0, 1, 1, 2, 2, 2 };
for (int i = start.x(); i <= end.x(); i++) {
	path.lineTo(QPointF{static_cast<qreal>(i), start.y() - offset[i%8]});
}
```

 was not only removing the `start` from the constructor argument but also including `start.x()` in the range of `i` in the for-loop so that the offset of y coordinate of the starting point was equal to that of the previous cell's ending point. In other words, this is equivalent to "fix" the y offset for each x coordinate (relative to window, not each cell) to value `offset[x % 8]`.

This strategy is also used in GVim (or rather the Neovim-qt's implementation is based on this?):

https://github.com/vim/vim/blob/032f40afb88b24043f96655aa248267a4f54faa8/src/gui_w32.c#L6343-L6356

The only issue I can think of is that now the y offset of the starting point is not always 0. I don't think it creates a new problem (at least as far as appearance concerned) though.

jgehrig:
Ahhh, I think I understand...

The problem is `QPainterPath path(start)` contains no offset. We're indexing into `offset` based on the absolute y-pixel count modulo 8. The line above always has a zero y-offset.

That line is incorrect for all cases except `offset[i % 8] == 0`, most cases. The issue is most noticeable when `offset[i % 8] == 2`.

Correct?

If so, let's add a quick comment to the code. I don't think this insight is immediately obvious.

Also, preference for the old CTOR style:
```
QPainterPath { /* start + offset[start.x() % 8]*/ }
for (int i = start.x() + 1;...
```

Yes, your code is cleaner... But the Qt documentation is vague: the behavior of `toLine` without a `currentPosition` is not clearly defined. I think it is better to be explicit and safe.

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -239,13 +239,13 @@ void ShellWidget::paintUnderline(
 static QPainterPath GetUndercurlPath(QRect cellRect) noexcept
 {
 	const QLine underline{ GetUnderline(cellRect) };
-	const QPoint& start{ underline.p1() };
-	const QPoint& end{ underline.p2() };
+	const QPointF& start{ underline.p1() };
```
jgehrig, line 242: The `&` is no longer necessary.

`QPoint` to `QPointF` conversion, so a new object is created either way.

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -239,13 +239,13 @@ void ShellWidget::paintUnderline(
...
+	const QPointF& end{ underline.p2() };
 
-	QPainterPath path(start);
-	static constexpr int offset[8]{ 1, 0, 0, 1, 1, 2, 2, 2 };
-	for (int i = start.x() + 1; i <= end.x(); i++) {
-		path.lineTo(QPoint(i, start.y() - offset[i%8]));
+	QPainterPath path{};
```
jgehrig, line 245: Let's use the old CTOR. We can still fix this with `for (int i = start.x() + 1`.

Qt is vague on how `toLine` behaves without an explicit `currentPoint`, which was previously provided in the CTOR.

src/gui/shellwidget/shellwidget.cpp
statiolake: Actually there is no floating-point operations and the cause was not related to rounding, I'll restore the original code around here.

jgehrig:
Marvelous! Thanks for your work: debugging, code, and explanation.

Looks ready for merge to me :+1:

PR #805, opened 2020-12-22
Short-term workaround for AppVeyor Windows breaks

jgehrig:
Upstream, Neovim has been forced to change CI:
https://github.com/neovim/neovim/issues/13312

For now, Windows Nighly Builds are unavailable.

We should test Windows against 0.4.4 in the interim.


jgehrig:
Breaks... Breaks! So many AppVeyor breaks, wow.

PR #807, opened 2020-12-25
Remove macro constant NVIM_QT_RUNTIME_PATH

### LINKED ISSUE #806: Unexpected runtime path C:\projects\neovim-qt\INSTALL\share\nvim-qt
In Windows, if a folder exists at `C:\projects\neovim-qt\INSTALL\share\nvim-qt`, Neovim-qt registers the directory to the runtimepath. If this folder is empty then Neovim fails to load neovim-gui-shim. In my personal case, there was an old neovim-gui-shim so I sometimes had unexpected behavior, e.g. my exit status improvement (#644) didn't work correctly as it required modified version of neovim-gui-shim.

The path is the INSTALL directory used inside CI like AppVayor.

I suspect following codes are the cause (and so I also suspect this is not Windows-specific but also affects other platforms except for macOS):

https://github.com/equalsraf/neovim-qt/blob/e8b37f069511ea7d7aa62b8ffbec17d3dc8689ef/src/gui/app_runtime.cpp#L16-L18

and `NVIM_QT_RUNTIME_PATH` is defined automatically by CMake at compile-time:

https://github.com/equalsraf/neovim-qt/blob/e8b37f069511ea7d7aa62b8ffbec17d3dc8689ef/src/gui/CMakeLists.txt#L58-L61

I think the paths only related to the build environment shouldn't be appear outside of the environment, but seeing it's disabled for macOS, it also appears to be intentional.

Anyway, I don't think hard-coding a path which is not related to the runtime environment is a good idea...

statiolake:
As an one of the easiest workaround for #806 , this PR removes macro constant NVIM_QT_RUNTIME_PATH. This works at least in my environment.

Closes #806.

### DIFF
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -55,11 +55,6 @@ add_library(neovim-qt-gui
 	${NEOVIM_RCC_SOURCES})
 target_link_libraries(neovim-qt-gui ${QTLIBS} qshellwidget neovim-qt)
 
-if(NOT APPLE)
-	set_property(SOURCE app_runtime.cpp PROPERTY COMPILE_DEFINITIONS
-		NVIM_QT_RUNTIME_PATH="${CMAKE_INSTALL_FULL_DATADIR}/nvim-qt/runtime")
-endif()
-
 add_executable(nvim-qt WIN32 MACOSX_BUNDLE main.cpp
 	${NEOVIM_RCC_SOURCES}
 	${RES_FILE}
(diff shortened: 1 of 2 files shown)


jgehrig:
I agree. I've merged this PR along with a refactoring of `app_runtime.cpp`: 8e167e1

Once `NVIM_QT_RUNTIME_PATH` is removed, we don't need the MacOS-specific implementation of `App::getRuntimePath`.

I don't think the path should be hard-coded on Windows. For example the package is compiled on a machine where `D:\Program Files` is used and consumed on a machine with the standard `C:\Program Files`.

You can make the argument this is safer on Linux/Unix, where the program directory structures are better defined. A contrived example is extracting a pre-compiled neovim-qt to `/opt`.

Thanks and keep the Pull Requests coming! :)

Issue #818, opened 2021-02-15
Support strikethrough highlighting


Melkster:
neovim-qt doesn't seem to support ~strikethrough~ highlighting. Is there any plan to add this?

The following command outputs strikethrough text when running neovim in terminal, but not in neovim-qt:

```vim
:hi Foo cterm=strikethrough gui=strikethrough | hi Foo
```

jgehrig:
The feature was added with: https://github.com/neovim/neovim/commit/3afb397407af3c94fc82d694186e8d451e625237

Yes, we should be able to easily support the feature. I was simply unaware of the new property, thanks for the ping.

jgehrig:
Merged, marking as closed.

The `guifont` `:s` option doesn't work currently, #824. Implementation for `:s` (and `:u`) is tricky, due to the way we render underline and strikethrough. Fortunately, I can't image many users want underline or strikeout on every character...

Please let me know if you notice any issues. Thanks!

p135246:
This is not working for me with vanilla nvim.

PR #819, opened 2021-02-15
Issue 818: Add strikethrough support

jgehrig:
**Issue #818:** Vim/Neovim added a new striketrough property. Add support.

See: https://github.com/neovim/neovim/commit/3afb397407af3c94fc82d694186e8d451e625237

We need to update `Cell`, `HighlightAttribute`, and `ShellWidget` to support the new property..

The strikethrough is renered as a Qt Canvas line halfway and one pixel up the cell in that cell's foreground color.


jgehrig:
Functional, but some code cleanup and more testing is required...

We should also add support for `s   - strikeout` in `guifont`.

I need to investigate how this interacts with `setStrikeOut` in `GetCellFont`. We may or may not want to add the property there too.

If anyone tests this, please let me know if this works or does not for you.

PR #821, opened 2021-02-18
Improve Font Page README.md

jgehrig:
The landing page could use improvement.

We should have a screenshot.

Several questions are asked repeatedly by users. We should make the questions/answers available in an obvious place: `GuiFont!`, `GuiTabline`, `GuiPopupmenu`, etc.

It would be nice to have a space to clearly describe common configuration options an features. Visuals would be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,3 @@
-
-Neovim client library and GUI, in Qt5.
-
 [![Build Status](https://circleci.com/gh/equalsraf/neovim-qt.svg?style=svg)](https://circleci.com/gh/equalsraf/neovim-qt)
 [![Build status](https://ci.appveyor.com/api/projects/status/c252f54mfjcuud8x/branch/master?svg=true)](https://ci.appveyor.com/project/equalsraf/neovim-qt/branch/master)
 [![Build Status](https://travis-ci.org/equalsraf/neovim-qt.svg?branch=master)](https://travis-ci.org/equalsraf/neovim-qt)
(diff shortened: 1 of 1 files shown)


jgehrig:
@jdhao

I've seen you file some bug reports on Windows for changes only available in `master`. Do you have a simple way of installing `neovim-qt` nighties on Windows?

If so, I would like to add this to the improved README. Thanks!

jdhao:
@jgehrig  I think the master build was downloaded from the CI. There is no automatic way to do this now. Maybe you can add a nightly build release for nvim-qt, just like what neovim currently does? Then it should be easy to install nvim-qt nightly version.

jgehrig:
@jdhao

Thanks for the confirmation, I do the same.

Getting automatic nightly releases from the CI is on my TODO list :)

README.md
```diff
@@ -1,99 +1,224 @@
...
+
+### Linux
+
+Neovim Qt is packaged for many distributions, use your favorite package manager.
+
+If your distribution is not listed here, please create an Issue with instructions.
+
+#### ArchLinux
+
+NeovimQt is available from this [AUR Package](https://archlinux.org/packages/community/x86_64/neovim-qt/).
```
fuelen, line 63: This is not AUR package.

jgehrig: Sorry, I don't run Arch...

Do you have better instructions/links you can provide? Thanks!

fuelen: `NeovimQt is available from this [AUR Package]` -> `NeovimQt is available in [community repository]`

Issue #833, opened 2021-04-13
german keyboard on MacOS


hjanuschka:
damn, the `--server` is a gamechanger!!!! huge props!

but, i have problems to enter some characters on my german macos qwertz keyboard, like `{` or `}` 

any ideas?

jgehrig:
Sure! We can get this resolved...

First thing to try is using a build from the latest commit:
```
$ brew tap equalsraf/neovim-qt
$ brew install --HEAD neovim-qt
```

I think there is a [typo](https://github.com/equalsraf/neovim-qt#macos) in my recent README update. Please let me know if the `--HEAD` above works for you.

You can check for the latest version with `--version`:
```
$ nvim-qt --version
NVIM-QT v0.2.17.9999
Build type: Release
Compilation:-march=native -O2 -pipe -Wall -Wextra -Wno-unused-parameter -Wunused-variable -std=c++11
Qt Version: 5.15.2
Environment: 
  nvim: nvim
  args: --cmd let &rtp.=',/usr/bin/../share/nvim-qt/runtime' --cmd set termguicolors
  runtime: /usr/bin/../share/nvim-qt/runtime
```

If your `{` key still doesn't work, I will need some output from a debug build to diagnose the issue:

https://github.com/equalsraf/neovim-qt/blob/621495a49df778d5f3e2625fb16d59c3483026e8/src/gui/shell.cpp#L1242-L1244

1. Un-comment the lines above
2. Build `neovim-qt` from source [MacOS Bulid Instructions](https://github.com/equalsraf/neovim-qt/wiki/Build-Instructions#macos)
2. Make sure to use `-DCMAKE_BUILD_TYPE=Debug`
3. Press `{` and `}` using your keyboard layout.
4. Paste the debug output here.

I'm working to slowly improve the input layer and add test coverage. I will use your debug output to write tests, and then fix the issue (if necessary).

Thanks!

hjanuschka:
this is what happens if i press `{`   - like on german keyboard -> OPTION+8


```

QKeyEvent ev: QKeyEvent(KeyPress, Key_Alt, AltModifier)
   ""
QKeyEvent ev: QKeyEvent(KeyPress, Key_8, AltModifier, text="{")
   "<A-{>"
QKeyEvent ev: QKeyEvent(KeyPress, Key_Control, ControlModifier)


```


this is what happens on the `}` - OPTION+9

```
[...]

PR #836, opened 2021-04-14
Issue 833: German Layout Handle {}

jgehrig:
**Issue #833:** German Layout Handle {}

We need to remove the `AltModifier` for `{}` keys on German Layouts.

Add handler and test coverage to prevent future breaks.


jgehrig:
Abandoning in favor of Pull Request #837.

PR #837, opened 2021-04-15
Issue 833: Handle German Layout AltModifier

hjanuschka:
hey there 🤚  - so since i had more troubles with more keys. it seems that it is a macOS specific thing, to the OPTION key

see the code change, the ifdef should prevent other OS's to break 👍 

ref: continue of https://github.com/equalsraf/neovim-qt/pull/836

p.s.: timezones and keyboard layouts where a big mistake!!


src/gui/input.cpp
```diff
@@ -195,6 +195,13 @@ QString convertKey(const QKeyEvent& ev) noexcept
 		return ToKeyString(GetModifierPrefix(mod), "Bslash");
 	}
 
+	// Issue#833: Handle MacOS Keyboard
+#ifdef __APPLE__
```
jgehrig, line 199: No `#ifdefs` please :)

It makes the code more difficult to test. Avoiding `#ifdefs` allows us to test Windows/Linux/Mac behavior on one machine.

Instead, you can define a function in `input.h`, and implement it for each platform in `src/gui/input_{platform}.cpp`.

Alternatively, it may be easier to do platform specific fixup in `CreatePlatformNormalizedKeyEvent(...)`.

src/gui/input.cpp
```diff
@@ -195,6 +195,13 @@ QString convertKey(const QKeyEvent& ev) noexcept
...
+	if (mod & Qt::AltModifier ) {
```
jgehrig, line 200: I think this is too broad. Does this pass Unit Tests?

Consider the case where a user wants to map `<A-X>` on a US layout.

Make sure to run the unit tests:
```
cd build
cmake --build .
ctest -R tst_input*
```

jgehrig:
Thanks! I'll leave this fix up to you :+1:

I've added a few comments, I think we need to narrow the scope.

I'd recommend a whitelist of characters, for which we remove `Qt::AltModifier`.

Psuedocode:
```
std::vector<QChar> removeAltCharList{ "{", "|", "}", ... }.

if (Key != Text && Text is not in removeAltCharList) {
 mod &= ~Qt::AltModifier;
}
```

Just a suggestion, perhaps you can come up with something better.

hjanuschka:
@jgehrig i re-did it - no more ifdef
and tests-pass (w/o changes) - this seem to be all OPTION+ keys i have on my german qwertz macos keyboard.


needed to use `QString`  instead of `QChar`  - let me know if you'd need me to change anything more.

jgehrig:
You've picked a hard change... The input layer is a minefield of gotcha's! :sweat_smile: 

The only values necessary in your map are:
`{ "[", "]", "|", "{", "}", "~", "@", "\" }`

The  conditional `c.unicode() >= 0x80 && c.isPrint()` above makes all other values unnecessary.

We need unit tests coverage for this issue. Can you add the `tst_input_mac` changes back and expand upon them?

Test coverage is *REALLY* important here. There are too many cases to reason through manually.

We should also cover false positives and Windows/Linux `QKeyEvents`. I'm happy to provide the Windows/Linux input if needed.

I suspect this will trigger false positives matches on non-german layouts.

Consider the following:
```
QKeyEvent ev: QKeyEvent(KeyPress, Key_BracketLeft, AltModifier, text="[")
   "<A-[>" // This is expected behavior on US and other layouts
```

Sorry for being so pedantic. This area is a frequent source of bug reports. The code is tricky :/

jgehrig:
I really appreciate your assistance! Thanks :+1:

src/gui/input_mac.cpp
```diff
@@ -48,13 +48,27 @@ QKeyEvent CreatePlatformNormalizedKeyEvent(
 	Qt::KeyboardModifiers mod,
 	const QString& text) noexcept
 {
+	
+	std::vector<QString> removeAltCharList  {"¡","“","¶","¢","[","]","|","{","}","≠","¿","±","‘","–","…","∞","µ","~","∫","√","ç","≈","¥","≤","å","‚","∂","ƒ","©","ª","º","∆","@","œ","æ","‘","±","•","π","⁄","¨","Ω","†","®","€","∑","«","„","\"","¿"};
```
jgehrig, line 52:
```
static const std::vector<QChar> removeAltCharList{  '{', ''}' ... }; /*constexpr is even better, but may not work?*/
...
if (std::find(removeAltCharList.begin(), removeAltCharList.end(), c ) != removeAltCharList.end()) {
```

`QChar` should be considerably faster/smaller than `QString`.

No heap allocation, much smaller memory footprint.

hjanuschka: done

hjanuschka:
so - well - infact i guess it will break languages, and overall this is not a solution for various other languages et-all


i added a check for the current input locale - and only applied the removal of altmodifier if locale is de_DE and mac - so the us keyboard should not be affected. maybe the approach with querieng the input locale and have mappings based on that  is generally a good idea.


however, i suck at cpp - not sure where to put 

```
QInputMethod * im = QGuiApplication::inputMethod();
```


to not execute it on every keypress, ideally this is done somewhere on main() and only a const property is defined.
can you guide me - to a sample area where such a thing is done?

jgehrig:
In C++, you can accomplish caching with a static local variable:
```
// Assuming this is inside a function/method
static const QInputMethod* im{ QGuiApplication::inputMethod() }; // Note: this is not thread safe.
```

However... Is it safe to cache the return value of that function?
What happens if the user switches between keyboard layouts?

I'm not sure what the performance implications are for `QGuiApplication::inputMethod()`  or if it can return `nullptr`. We should understand both.

If we use `QInputMethod`, we will also need to mock for test coverage. It is important to be able to validate unit tests regardless of the host platform/localization.

We also need to fix this on Linux/Windows.

Here is Linux `QKeyEvent` output:
```
QKeyEvent ev: QKeyEvent(KeyPress, Key_BraceLeft, GroupSwitchModifier, text="{")
   "{"
QKeyEvent ev: QKeyEvent(KeyPress, Key_BracketLeft, GroupSwitchModifier, text="[")
   "["
QKeyEvent ev: QKeyEvent(KeyPress, Key_BracketRight, GroupSwitchModifier, text="]")
   "]"
QKeyEvent ev: QKeyEvent(KeyPress, Key_BraceRight, GroupSwitchModifier, text="}")
   "}"
```

I will find a Windows machine to generate `QKeyEvent` for Windows + de_DE.

hjanuschka:
i  added `currentLocale`  and `currentInputMethod` to the `Shell` class. and this one fetches the locale, and also updates it with the `localeChanged` signal - in case you change your keyboard layout.

and then passes the Qlocale down to the keyboard logic.


can you please let me know if this approach is feasable for you?
adding another arg to the `convertKey` is discussable - but i didn't find another/better way. so that we still can connect signals, and not having to run any logic (besides the `==`  to a string;de_DE) on every keystroke.


i worked a whole day withit, so i can fully confirm that german keyboard on macos works perfectly.
if you can confirm that this poc is ok, i ll continue and try to fix the unittests, and also - if possible add new tests (never did cpp tests tbh)

jgehrig:
Generally, I like the locale approach! It is much cleaner than trying to write code that works for all localizations :+1:

I prefer adding a parameter to `convertKey`. We should avoid adding state. Keeping the function pure/stateless makes it much easier to test.

One API suggestion:
```C++
// input.h
QString convertKey(const QKeyEvent& ev, QLocale* locale = nullptr) noexcept;
```

I would recommend using a `QLocale*` as an optional parameter with a default `nullptr` value provided: pointer as optional.

If a `locale` is provided, we can check localization specific fix-up. If `nullptr`, skip the localization specific fixes entirely.

This makes your changes API compatible with old callers, no code changes for old callers required.

---
And... Don't forget unit tests! :smile: 

Any changes we make here should be should designed for testability. All scenarios must be testable from a arbitrary machine.

For example, we shouldn't need a Mac to test the Mac input layer; the same applies to german and other localizations. 

The one potential blocker on this approach is: How do we generate `QLocale` values for test?

Probably possible, I just don't know the answer...

src/gui/shell.h
```diff
@@ -230,6 +232,8 @@ private slots:
 	QPoint m_mouse_pos;
 	// 2/3/4 mouse click tracking
 	QTimer m_mouseclick_timer;
+	QLocale currentLocale;
```
jgehrig, line 235: `m_` prefix for member variables.

We should also provide a sane default value:
QLocale m_currentLocale{ QLocale QLocale::system() };

nitpick: `QInputMethod* currentInputMethod`

If we don't need `QInputMethod` the signal/slot finishes, we should avoid caching the value. Less state, less bugs.

src/gui/shell.cpp
```diff
@@ -1237,11 +1245,12 @@ void Shell::keyPressEvent(QKeyEvent *ev)
...
 	}
 
-	const QString inp{ Input::convertKey(*ev) };
-
+	const QString inp{ Input::convertKey(*ev, currentLocale) };
+	
 	// Uncomment for key input debugging and unit test writing.
 	// qDebug() << "QKeyEvent ev:" << ev;
 	// qDebug() << "  " << inp;
+	// qDebug() << " Keyboard Locale: " << currentLocale.name();
```
jgehrig, line 1253: Good call! :+1:

src/gui/shell.cpp
```diff
@@ -55,13 +56,20 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 {
 	setAttribute(Qt::WA_KeyCompression, false);
 
+	currentInputMethod = QGuiApplication::inputMethod();
+	currentLocale = currentInputMethod->locale();
```
jgehrig, line 60: Move this up to the initializer list:
```
  currentInputMethod{ QGuiApplication::inputMethod() };
  currentLocale{ currentInputMethod->locale() };
```

Using that `{}` init syntax helps the compiler catch type errors. C++ is weird!

hjanuschka: done

src/gui/shell.h
hjanuschka: m_prefix done, not sure if i understand, we need the ref to QInputMethod so that dynaic locale change can happen?!

hjanuschka:
@jgehrig  feedback addressed, and refactored the de_DE handler a bit. - old test still - compile and pass. 


mocking a Qlocale should be easy like:


```
QLocale germany(QLocale::Germany);
```


full list seems to be here: https://doc.qt.io/qt-5/qlocale.html#Country-enum

i ll start adding tests, tomorrow.

src/gui/input.h
```diff
@@ -5,14 +5,15 @@
...
+#include <QLocale>
 
 namespace NeovimQt { namespace Input {
 
 /// Returns the QMap of all NeoVim-recognized special keys.
 const QMap<int, QString>& GetSpecialKeysMap() noexcept;
 
 /// Convert Qt key input into Neovim key-notation. See QKeyEvent.
-QString convertKey(const QKeyEvent& ev) noexcept;
+QString convertKey(const QKeyEvent& ev, QLocale* locale = nullptr) noexcept;
```
jgehrig, line 16: `const QLocale*` (if possible).

Sorry, my bad I missed this in my API suggestion.

src/gui/input_mac.cpp
```diff
@@ -42,19 +43,45 @@ QString GetModifierPrefix(Qt::KeyboardModifiers mod) noexcept
...
+	
 	if (!text.isEmpty())
 	{
 		const QChar c{ text.at(0) };
 		if (c.unicode() >= 0x80 && c.isPrint()) {
 			mod &= ~Qt::AltModifier;
 		}
+		
+		// German/Austrian/Swiss MacOS QWERTZ keyboard, remove AltModifier (OPTION KEY)
+		if(locale != nullptr) {
```
jgehrig, line 75: `if (locale)`

No need for `== nullptr` or `!= nullptr` in C++.

src/gui/input_mac.cpp
```diff
@@ -42,19 +43,45 @@ QString GetModifierPrefix(Qt::KeyboardModifiers mod) noexcept
...
+			if(locale->name() == "de_DE" || locale->name() == "de_AT" || locale->name() == "de_CH") {
```
jgehrig, line 76: Totally optional: `locale->name().startsWith("de_", /*? Qt::CaseSensitive*/`

Something to consider if this applies to any german layout... Not strictly necessary.

Perhaps we only want call out certain german layouts?

src/gui/shell.cpp
```diff
@@ -1288,6 +1299,20 @@ void Shell::mousePressEvent(QMouseEvent *ev)
...
+/** Update Current Locale */
+void Shell::updateCurrentLocale() {
+	m_currentInputMethod = QGuiApplication::inputMethod();
+	if(m_currentInputMethod != nullptr) {
+		m_currentLocale =  m_currentInputMethod->locale();
+	} 
+}
+/** Keyboard Locale has changed */
+void Shell::localeChanged()
+{
```
jgehrig, line 1312: I could be wrong... If this doesn't work, ignore my comment :)

No need for `updateCurrentLocale` or `m_currentInputMethod`.

I think you can write:
```C++
void Shell::localeChanged() noexcept
{
  const QInputMethod* inputMethod{ QGuiApplication::inputMethod() };
  if (!inputMethod)
  {
    qDebug() << "Unexpected: QGuiApplication::inputMethod() returned nullptr!"
    return;
  }

  m_currentLocale = inputMethod->locale();
}
```

Less code, less state.

src/gui/shell.h
```diff
@@ -123,6 +124,8 @@ protected slots:
 	void neovimExited(int);
 	void neovimResizeFinished();
 	void mouseClickReset();
+	void localeChanged();
```
jgehrig, line 127: All new functions should be marked as `noexcept`... Yet another weird C++ thing.

jgehrig:
Looks great! I'd consider this check-in ready once tests are added :+1: 

Let me know if you'd like to fix Windows/Linux, or if I should handle that afterwards.

I can provide you with the necessary `qDebug()` output for both platforms.

src/gui/input.h
hjanuschka: done

src/gui/input_mac.cpp
hjanuschka: done

src/gui/input_mac.cpp
hjanuschka: i think its good to do with de_ - and see if any regression pops up.

src/gui/shell.cpp
hjanuschka: initially i neede to call the logic from that method on 2 places
  - the "change" signal.
  - on startup

the startup part moved into the initializer - so yes, i am going to move it

hjanuschka: ok that doesnt work out, one needs the ref to the inputMethod for the signal and so on, so i  think  we'll keep it.

hjanuschka:
so i think i am done. added tests for macos - let me know if i did that correctly.


i would love to add linux and windows, i guess it's pretty much the same, just OPTION swapped with ALTGR - please post the qDebug, as i don't really have access to linux/windows ( well would work with a VM ) - but i have no non-macOS keyboards at home..... and being in the office is not happening in the near future :/

src/gui/input_mac.cpp
```diff
@@ -1,5 +1,6 @@
 #include "input.h"
-
+#include <QInputMethod>
+#include <QApplication>
```
jgehrig, line 3: I think these headers no longer necessary in `input_mac.cpp`?

src/gui/input_mac.cpp
```diff
@@ -42,19 +43,45 @@ QString GetModifierPrefix(Qt::KeyboardModifiers mod) noexcept
 	return modprefix;
 }
 
+// Fixup Modifiers, Apples OPTION key
+Qt::KeyboardModifiers de_DE(Qt::KeyboardModifiers mod, QChar c) noexcept
```
jgehrig, line 47: Let's use a more descriptive name.

`FixInputForGermanLayouts`?

src/gui/shell.cpp
```diff
@@ -52,16 +53,22 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 	: ShellWidget{ parent }
 	, m_nvim{ nvim }
 	, m_options{ GetShellOptionsFromQSettings() }
+	, m_currentInputMethod{ QGuiApplication::inputMethod() }
+	, m_currentLocale { m_currentInputMethod->locale() }
 {
 	setAttribute(Qt::WA_KeyCompression, false);
 
+
```
jgehrig, line 61: nit: remove extra line

src/gui/shell.cpp
```diff
@@ -1237,11 +1244,15 @@ void Shell::keyPressEvent(QKeyEvent *ev)
 		this->setCursor(Qt::BlankCursor);
 	}
 
-	const QString inp{ Input::convertKey(*ev) };
-
+	QLocale* useLocale { &m_currentLocale };
+	const QString inp{ Input::convertKey(*ev, useLocale) };
```
jgehrig, line 1248: Consume directly, no need for `useLocale`:
`const QString inp{ Input::convertKey(*ev, &m_currentLocale) };`

test/tst_input_mac.cpp
```diff
@@ -12,8 +13,50 @@ private slots:
 	void SpecialKeys() noexcept;
 	void KeyboardLayoutUnicodeHexInput() noexcept;
 	void CtrlCaretWellFormed() noexcept;
+	void GermanKeyBoard() noexcept;
 };
 
+void TestInputMac::GermanKeyBoard() noexcept
```
jgehrig, line 19: nit: Keyboard

jgehrig: nit: remove the double line returns below.

src/gui/shell.cpp
jgehrig: Ahhh I see the issue, Signal/Slot connection. It should be safe to call QGuiApplicatoin::inputMethod() directly instead.

From my quick read, the function returns a single instance every time:
https://code.woboq.org/qt5/qtbase/src/gui/kernel/qguiapplication.cpp.html#_ZN15QGuiApplication11inputMethodEv

The perf on that function should be good, no need for `m_currentInputMethod`.

If the object returned were to change, we'd have a bug: `connect(m_currentInputMethod, ...` is not updated to the new object.

Patch attached for your reference:
[RemoveCurrentInputMethodMemberPatch.txt](https://github.com/equalsraf/neovim-qt/files/6332135/RemoveCurrentInputMethodMemberPatch.txt)

`git apply RemoveCurrentInputMethodMemberPatch.txt`

src/gui/input_mac.cpp
hjanuschka: i thought of sticking to de_DE and so on might be easier to handle if there is others comming up like pt_PT (portugal) and so on.

hjanuschka:
feedback addressed, only one conv. open

jgehrig:
Looks great! I'm comfortable merging this as-is.

I added a few nit-picky comments have been added since we'll also fix this for Windows/Linux. Sorry! :sweat_smile: 

Here is some Linux output for now
```
QKeyEvent ev: QKeyEvent(KeyPress, Key_BraceLeft, GroupSwitchModifier, text="{")
   "{"
QKeyEvent ev: QKeyEvent(KeyPress, Key_BracketLeft, GroupSwitchModifier, text="[")
   "["
QKeyEvent ev: QKeyEvent(KeyPress, Key_BracketRight, GroupSwitchModifier, text="]")
   "]"
QKeyEvent ev: QKeyEvent(KeyPress, Key_BraceRight, GroupSwitchModifier, text="}")
   "}"
```

I'll generate a complete equivalent set to what you've used for MacOS... Windows is giving me some trouble, DLL linker issues.

src/gui/shell.cpp
hjanuschka: patch manually applied!

jgehrig:
We may have a problem here... Does the `QInputMethod::localeChanged` signal ever fire for you?

On my Linux machine, the signal does not fire when I switch between `us` and `de` keyboard layouts.

This looks like an issue with Qt:
https://stackoverflow.com/questions/17346849/how-to-catch-keyboard-layout-change
https://bugreports.qt.io/browse/QTBUG-27681
https://www.qtcentre.org/threads/68763-How-to-get-the-current-layout-of-keyboard

And Qt has marked the issue as closed? The Qt dev's comments suggest they won't fix this bug :confused: 

Also, we should update the `Uncomment` section in `shell.cpp`. Here is what I used:
```C++
// Uncomment for key input debugging and unit test writing.
// qDebug() << "QKeyEvent + QLocale:" << ev << m_currentLocale;
// qDebug() << "  " << inp;
```

Attached is output from above on my Linux machine: [output.txt](https://github.com/equalsraf/neovim-qt/files/6332674/output.txt)

It looks like the behavior is slightly different than MacOS. I went down my keyboard pressing each key with `AltGr`/`Right Alt` held. Let me know if you need additional key-combinations.

hjanuschka:
ok - so you thing we should remove the signal et-al? - it doesnt fire, in fact.


about the keycodes you recorded, thos show up as US keyboard layout, can you run it with german?

hjanuschka:
ok - i tried to do in theorie.


i would like to merge this PR - and finish it. i found an old keyboard down in the basement - going to bootup some VM's and do a seperate UNIX and another windows PR.



it is to risky to do it,  in one run without having it tested/tried myself.

jgehrig:
Without a working `QEvent::KeyboardLayoutChange` or `localeChanged()`, the `QLocale` approach is not viable. Sorry.

And it was a very clever/elegant solution too!

If not for the Qt bug, it would be a BIG improvement...

> it is to risky to do it, in one run without having it tested/tried myself.

The input layer test coverage is quite good. As long as accurate `QKeyEvents` are added as test-cases, I am comfortable.

The worst case is we miss something, and have the opportunity to add more test coverage. I will also do testing before merge.

> i would like to merge this PR - and finish it.

Unfortunately, the `QLocale` bug  is a blocker...

Pending the `QLocale` issue resolution, I am happy to merge a MacOS-only fix.

We can address Linux/Windows with a follow-up PR.

hjanuschka:
> Without a working QEvent::KeyboardLayoutChange or localeChanged(), the QLocale approach is not viable. Sorry.

i don't fully understand, the "only" bug is - that a dynamically changed keyboard locale - does not apply, which just means you'd need to restart neovim-qt after keyboard layout change - not perfect for sure.


i think - changing layout is not the most day-to-day task one does.


without accepting the MR - it renders neovim-qt useless for de_ keyboards on MacOS (and probably windows/unix too) - doing it without querieng the local et all - ends up in a minefield of spagetti code if/elseing the key combos.




Linux/Windows german keyboard is the same, if you don't accept the QLocale - it won't be able to fix without the minefield!



let me know what you think.

jgehrig:
Sorry for the merge conflict... Fixed!

I used this as an excuse experiment with GitHub's 'allow maintainer edits' option. Neat feature!

hjanuschka:
yeah - great feature.

ok i get the point, so from observations it looks like Qlocale is in fact not the locale of the input method (which the variable names suggest) - but it is the Locale of the OS - so if i change my ubuntu VM's language (not keyboard layout) it works fine.


i still believe in the - get a clean string like iso code - and do the IF - dance with that. instead of combo-guessing - its like u fix one you break 3 game......


what about this, we find native ways - to querie the operating system for keyboard layout - transform whatever the OS reports back into a standardized, string and work from there, not passing down QLocale but some "NVIMKeyboardLayout".



like on MacOS there seems to be an api that works flawless, even changing (with restart/rerun - for sure; but that is what i could totally live with)



```c
#include <Carbon/Carbon.h>
int main() {
    char layout[128];
    memset(layout, '\0', sizeof(layout));
    TISInputSourceRef source = TISCopyCurrentKeyboardInputSource();
    // get input source id - kTISPropertyInputSourceID
    // get layout name - kTISPropertyLocalizedName
    CFStringRef layoutID = TISGetInputSourceProperty(source, kTISPropertyInputSourceID);
    CFStringGetCString(layoutID, layout, sizeof(layout), kCFStringEncodingUTF8);
    printf("%s\n", layout);
    return 0;
}
```

it returns for German: `com.apple.keylayout.German`  or Austrian (which i am from): `com.apple.keylayout.Austria`  `com.apple.keylayout.US`  for US. so we could actually make a mapping hash from `com.apple.keylayout.US`  to `us_US`  and so on.


so we end up with a `NVIMLocale` => `de_DE`  - and implement a little native helper on all 3 platforms.


currently investigating how to reliable query the layout on linux, xkb seems to be way to go.

via shell this is easy `setxkblayout -query|grep 'layout:'|awk '{print $2}'` => `us` or `de`

this project also seems to have some code to extract xkb via cpp - avoiding a shell out. https://github.com/nonpop/xkblayout-state


for windows, which is like another planet to me this maybe a way:  https://stackoverflow.com/a/50575608/1945875

hjanuschka:
if the platform, does not return anything, or is not queryable, i would just take the system locale as a default, i guess in *most* cases system locale equals keyboard layout.

jgehrig:
> i still believe in the - get a clean string like iso code - and do the IF - dance with that. instead of combo-guessing - its like u fix one you break 3 game......

In an ideal world, I too would prefer `QLocale` or `QEvent::KeyboardLayoutChange`... But, Qt is not perfect.

The test coverage has made the key-combo-if-dance much more manageable. As long as the fixes are properly scoped (not too broad), it has worked very well in practice. The key is robust test coverage and readable/commented code.

> if the platform, does not return anything, or is not queryable, i would just take the system locale as a default, i guess in most cases system locale equals keyboard layout.

I am hesitant to do this. It adds one more variable to evaluate when diagnosing input layer issues.

For what benefit? We can do a better job (albeit with more manually written logic) of interpreting `QKeyEvents` without `QLocale`.

From what I've observed, `QLocale` isn't reliable signal for the meaning of a specific `QKeyEvent`. Unfortunate.

> what about this, we find native ways - to querie the operating system for keyboard layout - transform whatever the OS reports back into a standardized, string and work from there, not passing down QLocale but some "NVIMKeyboardLayout".

Such a solution could quickly creep in size and become a maintainability problem:
 - FreeBSD? Linux? MacOS? Windows? Others?
 - Which versions of each platform do we support?
 - What libraries should we use and test against? Min/Max Version?
 - Communicating these new dependencies to package maintainers?

Staying within the Qt framework keeps dependencies and testing simple. I want to avoid complexity, unless it is necessary.

hjanuschka:
ok adding native/platform specific helpers, will for sure introduce, complexity and code to maintain.

totally agree that QT should fix that, but from how i read that issues/posts - that will not happen :)


i ll just stick to my fork now - that solves my MacOS-DE/AT issues, let me know if you find better approaches and wan't me to continue this PR in whatever way.

jgehrig:
Closing this PR.

I merged your test coverage + the non-`QLocale` fix in #841. Thanks for driving this!

Sorry again for the run-around... I would have preferred this solution if Qt had better support for on-the-fly keyboard layout changes. :/

Issue #731, opened 2020-08-05
Improve fork behaviour on Mac OS X, fork when launched from terminal


equalsraf:
Back in https://github.com/equalsraf/neovim-qt/issues/203 we added the nofork cli argument. This is a followup to improve the default behavior on Mac OS X.

The current behavior with regards to forking the process on startup is:

1. in unix (but not mac os) we fork by default. A --nofork argument can be used to disable this.
2. in windows we dont fork (because we cant)
3. in mac os x we dont fork, because the launcher expects the application to stay in the foreground

Point 3 is not ideal if you are launching nvim-qt from a terminal. The request is to allow nvim-qt to fork when launched from a terminal. Or alternatively to enable the same default behavior as in UNIX but ensure when launched via the launcher it does not fork.

x37v:
I use nvim-qt on a mac and launch from the terminal and background with `&` .. i find that pretty frequently after i quit nvim-qt i still have a `nvim` process running. I wonder if it is related to the forking behavior?

equalsraf:
> I use nvim-qt on a mac and launch from the terminal and background with & .. i find that pretty frequently after i quit nvim-qt i still have a nvim process running. I wonder if it is related to the forking behavior?

It should not be related to this issue though, this one is about nvim-qt forking its own process, and in mac os x we dont really do that anyway.

Do you mind opening another issue for this? Also you might be able to reattach to the nvim process to check what is going on.

PR #838, opened 2021-04-16
Mac OS: fork process when started from a tty

equalsraf:
The goal here is to improve the mac os x fork behavior: https://github.com/equalsraf/neovim-qt/issues/731

I did not have a chance to test this is Mac OS yet, and I dont have an idea for a test here.

### DIFF
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -3,6 +3,10 @@
 #include <QProcess>
 #include <QFile>
 #include <QCommandLineParser>
+#if defined (Q_OS_MAC)
+# include <unistd.h>
+# include <stdio.h>
+#endif
 #include "neovimconnector.h"
 #include "app.h"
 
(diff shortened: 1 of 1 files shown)


equalsraf:
Tested in Mac OS, --no-fork (or lack of it) now works the same as in other unixes when running from terminal. I still have to check the case when it is not running from a terminal.

equalsraf:
I am done testing. It looks ok, tested from

1. the dock, by pinning a manually started binary
2. by piping /dev/null into nvim-qt stdin

jgehrig:
I'd help test, but I don't own a Mac.

One idea, you could get rid of the `#ifdefs` using the same pattern as `input_*.cpp`. No conditional includes, just a simple function call `bool IsTTYStandardInput()` (with a better name). Implement once for MacOS, and one `_emptyimpl.cpp` for Win32/Unix which always returns false. Share the header definition across all platforms.

Definitely not essential... Feel free to ignore.

I think `#ifdef` is another one of those evils specific to C/C++ software! :wink:

equalsraf:
I agree, but i am inclined to leave it for now.

I think if i were to refactor to remove the ifdefs then i would have to remove both of them, something like a platform specific main() implementation where

- unix defaults to nofork=false, checks arguments and picks the cli or gui main depending on the value
- mac os is identical to unix except the default nofork is conditional on the stdin tty
- windows ignores nofork and just goes into the ui version

I am also wondering if we can adopt the mac os behavior in all unix and avoid the ifdef entirely, but i would need to think a bit further about it.

PR #841, opened 2021-04-25
Issue 833: Handle MacOS German Keyboard AltModifier

jgehrig:
**Issue #833**: Fix German MacOS AltModifier low-ascii keys.

Some layouts require Alt to input low-ascii characters (German "{").
Other keyboard layouts do not require Alt (US "{").

`AltModifier` should be removed for the former but not the later. Unfortunately, Qt does not provide robust keyboard layout detection. See #836.

We can indirectly detect this scenario by comparing `Qt::Key` and text values. `AltModifier` will be removed for a subset of known characters whenever `Qt::Key` does not directly map to text.

### DIFF
--- a/src/gui/input_mac.cpp
+++ b/src/gui/input_mac.cpp
@@ -42,18 +42,53 @@ QString GetModifierPrefix(Qt::KeyboardModifiers mod) noexcept
 	return modprefix;
 }
 
+// Some keyboard locales require Alt to input low-ascii characters (German "{").
+// Other keyboard locales do not require Alt (US "{"). Remove Alt on layouts where
+// the key modifies text input. Qt does not provide a robust layout-detection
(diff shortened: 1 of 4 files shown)


jgehrig:
FYI @hjanuschka 

I'm sorry for the `QLocale` runaround... Darn Qt bugs!

Here is how we will fix the change. Please let me know if this doesn't work for you.

I copped some of your code (and you really drove this fix), so I will give you commit credit when this gets merged.

jgehrig:
We still need to add Windows test coverage.

Generating a build with AppVeyor now...

hjanuschka:
💯 gave it a try, works flawless - many thx!

jgehrig:
For reference, Win32 `qDebug()` output for the German keyboard layout: [win32-qdebug-log.txt](https://github.com/equalsraf/neovim-qt/files/6415954/win32-qdebug-log.txt)

PR #840, opened 2021-04-22
Support builds against Qt6

equalsraf:
See https://github.com/equalsraf/neovim-qt/issues/839 for the list of issues

In terms of code this is ready for review. There are still runtime issues to be checked (font integer metrics and possibly imfocus). But since we continue to default to qt5, I think it is ok to proceed and handle those separately.

Issue #839, opened 2021-04-22
Qt6 support


equalsraf:
Tracking issues for supporting builds against Qt6 - I was able to build against Qt6 after some changes, but I could not test, here is the list of issues

- [x] CMake can support mixed Qt5/6 builds https://doc-snapshots.qt.io/qt6-dev/cmake-qt5-and-qt6-compatibility.html
- [x] windeployqt is failing in qt6 windows
- [ ] static builds (windows) are failing for qt6 - we also lack CI for this option

APIs removed in Qt6, need to check if these can be replaced (current minimal version is 5.8):

- [x]  QLayout::setMargin
- [x] QStringRef -> QStringView
- [x] QStringView.toInt()/toFloat() does not exist in Qt5
- [x]  Qt::MidButton
- [ ]  Qt::ImMicroFocus
- [x]  QPalette::Background/Foreground
- [x] QDesktopWidget -> QScreen
- [ ] QFont::ForceIntegerMetrics
- [x] QProcess/QTcpSocket/QLocalSocket::error(...errortype) signal is not available in qt6
- [ ] qRegisterMetaTypeStreamOperators was removed from Qt6, need to test QSettings serialization (see neovimconnector.cpp)

Changes:

- [x] QTextCodec is now part of the Core5Compat module and this module is not part of the default install (windows?) - just remove it - we always hardcode utf8 anyway!
- [x] dependency on QtSvg moves to QtSvgWidgets - and consumers of these dlls e.g. https://github.com/neovim/neovim/blob/667e3bfb6e9dcd3287f2082edd83068d5428f3af/src/nvim/CMakeLists.txt#L507-L568

Other issues

- [x] some includes had to be adjusted
- [x] const operator needed adjustment
- [x] QList and `{}` fail to build tests, there is also an error with const in the related for loops
- [ ] new deprecation warnings
- [ ] test failure in the shell widget, related to the font (invalid font weight)

jgehrig:
> dependency on QtSvg moves to QtSvgWidgets

I just thought of this... We should watch for DLL name changes on Windows.

Neovim hard-codes DLLs for copy:
https://github.com/neovim/neovim/blob/667e3bfb6e9dcd3287f2082edd83068d5428f3af/src/nvim/CMakeLists.txt#L507-L568

```
https://github.com/neovim/neovim/issues/12928
```

The response from Neovim will probably be to remove the bundled `neovim-qt`. They almost did that last time :/

[...]

PR #842, opened 2021-04-25
Cleanup deprecated code before enabling qt6 builds

equalsraf:
While porting for Qt6 https://github.com/equalsraf/neovim-qt/issues/839 I came across some APIs that were deprecated during the Qt5 cycle and removed in Qt6. I have extracted these from https://github.com/equalsraf/neovim-qt/pull/840 and they correspond to changes in Qt6 that would not build, and should not alter the current behavior (i.e. the easy stuff).

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -543,7 +543,7 @@ void ShellWidget::paintForegroundTextBlock(
 
 		// When the cursor IS within the glyph run, decompose individual characters under the cursor.
 		const int cursorGlyphRunPos { cursorPos - glyphsRendered };
-		const QString textGlyphRun{ QStringRef{ &text, glyphsRendered, sizeGlyphRun }.toString() };
+		const QString textGlyphRun{ QStringView{text}.mid(glyphsRendered, sizeGlyphRun).toString() };
 
 		// Compares a glyph run with and without ligatures. Ligature glyphs are detected as differences
 		// in these two lists. A non-empty newCursorGlyphList indicates glyph substitution is required.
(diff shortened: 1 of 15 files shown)


test/tst_input.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+// A class to hold test data, an event type/key/modifiers
+// as used in QKeyEvent and a matching Neovim input string.
+class InputTest
```
jgehrig, line 5: `struct InputTest`?

We're using this like a struct, and a class with all public members is basically a struct anyways.

test/tst_input.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+// A class to hold test data, an event type/key/modifiers
```
jgehrig, line 3: an -> and?

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -1007,7 +1007,9 @@ QVariant ShellWidget::TryGetQFontFromDescription(const QString& fdesc) const noe
 	for (const auto& attr : attrs) {
 		if (attr.size() >= 2 && attr[0] == 'h') {
 			bool ok{ false };
-			qreal height = attr.midRef(1).toFloat(&ok);
+			// TODO: the additional toString() call is needed to be compatible with
```
jgehrig, line 1010: What is still left `TODO` here?

Should we file an Issue and reference it here so the fix-up isn't lost to the void?

jgehrig (approved):

src/gui/shellwidget/shellwidget.cpp
equalsraf: I've updated the comment.

The extra toString() call is needed when Qt < 5.15, but it can be removed for Qt 5.15 or Qt6. Once we the minimum supported Qt version reaches 5.15 this can be removed.

test/tst_input.h
equalsraf: yup, updating.

test/tst_input.h
equalsraf: I did meant to use "An". "An event (type/key/modifier) and a matching neovim input string"

PR #845, opened 2021-04-30
Generate installer in Appveyor builds

equalsraf:
Still WIP, but to make life a little easier I was trying to get Appveyor to generate .exe NSIS installers for neovim-qt.

The installer packages the same files as the zip file, plus an NSIS uninstaller.

Note: Appveyor is unable to get the neovim nightly builds (404) since today, that is why i added the second commit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,4 +120,3 @@ if (PYTHONINTERP_FOUND)
 endif()
 
 include(CPackOptions)
-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,8 +7,6 @@ endif (POLICY CMP0048)
 # 9999 = Development Pre-Release
 project(neovim-qt VERSION 0.2.17.9999)
 
-INCLUDE(CPack)
-
 if(NOT EXISTS ${NEOVIM_EXEC})
 	set(NEOVIM_EXEC nvim)
 endif()
@@ -121,4 +119,5 @@ if (PYTHONINTERP_FOUND)
 		)
 endif()
 
+include(CPackOptions)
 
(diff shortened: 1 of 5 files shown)


equalsraf:
The issue we saw with nightly neovim builds is gone. This is good to go.

jgehrig:
Quickly tested on one of my Windows machines. Works great! :+1: 

A few comments:
1) Do we want to include the version? Will this complicate updates? `C:\Program Files\neovim-qt 0.2.17`
2) We should update the [Wiki Page](https://github.com/equalsraf/neovim-qt/wiki/Windows-AppVeyor-Builds) after checkin.

I wonder if we should bundle `nvim`? Probably not... That might get tricky and Chocolatey properly pokes the `%PATH%`.

jgehrig:
Also...

If neovim does pull the trigger after 0.5.0 and drops the bundled neovim-qt. We should probably investigate how to register a package with Chocolatey.

equalsraf:
> Do we want to include the version? Will this complicate updates? C:\Program Files\neovim-qt 0.2.17

The Installer was supposed to handle this gracefully, using the UUID in the cmake file to identify other installs.

> We should update the Wiki Page after checkin.
> I wonder if we should bundle nvim? Probably not... That might get tricky and Chocolatey properly pokes the %PATH%.

I'm still thinking about these. The PATH in particular is the reason why i dont just drop the exe in the build artifacts and update the wiki - it would annoy people to install a broken setup.

It might be worth it to clearly error the case when nvim is not in path.

The other option here that i have not checked is having  the installer check if neovim is already installed and error out if it is not.

Technically it is also possible to have Wix install dependencies, but bundling neovim might be too much.

> If neovim does pull the trigger after 0.5.0 and drops the bundled neovim-qt. We should probably investigate how to register a package with Chocolatey.

Yup, I guess there is also the new winget https://github.com/microsoft/winget-pkgs

equalsraf:
I'll continue thinking about it in https://github.com/equalsraf/neovim-qt/issues/852

PR #848, opened 2021-05-12
Adding support for multiple windows in the same process on macOS

### LINKED ISSUE #850: Adding support for multiple windows in the same process on macOS
I'm openning this to continue the conversation from #848 so that we can keep the conversation separate.

-----

On macOS, it's more common for windows to belong to the same process than the other way around. VimR works the same way, and all the common macOS apps. The only notable example that doesn't do that is Alacrity.

Having the windows belong to the same window makes window navigation much simpler with app exposè or "Cmd+`". This behavior would only be specific to macOS because having multiple windows on the other platforms provides no such benefits (I'm not 100% sure about this for Linux though. It's been a while since I was there.).

Here's how I think this would work.

- Multiple window support is only enabled on macOS.
- From a running nvim-qt instance, a new window can be launched with a command. And this command would only be available for macOS, or it would exist but be no-op on other platforms. I think it not existing is better.
- I'm not so sure about how to handle `open -a nvim-qt`. My gut feeling is that it would open the file either in the first window created, or in the one that was last activated. I don't have a technical understanding of how it works yet.

> I can think of a few cases when you want multiple nvim instances and or multiple top level windows, but I'm not sure if these match your motivation:

> 1. run multiple nvim instances, one in each container/machine/etc
> 2. neovim isolation is not perfect - e.g. some features are not 100% independent per tab , or i dont want to list all buffers just the buffers i am working on right now - this is probably addressable in neovim, however as a consequence people run multiple neovim instances anyway.
> 3. multiple screens - however in this case what people usually wan is one neovim instance for multiple top level windows

Similar motivations. I use different environments/variables to manage my sessions and having them all in one nvim session is not feasiable for me.

Also, ideally the new instance in a window would:

1- Inherit the environment if it's launched in the terminal. This would mean that I can launch different Neovim instances in the same process with different Python virtual environments or whatever else environment people use.
2- Inherit the environment of the previously active window if it's launched from an existing window with a command.

Furkanzmc:
This is my first contribution here. The end goal for me is to have the a single app contain multiple windows on macOS as an option. It makes window switching much more convenient with "Cmd+`". In order to do that, I first wanted to remove the NeovimConnector dependency from the app and have it owned by each window that might be created.

This kidn of breaks the `:cq` support when there's multiple windows, though. Before I dive much into it, I wanted to make sure I can get a proper PR in. I think I'll open an issue about the multiple window support because there are some things that has to be considered before it can be implemented.

Closes #850.

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -13,6 +13,65 @@
 
 namespace NeovimQt {
 
+struct ConnectorInitArgs {
+	enum Type {
+		Embed,
+		Server,
+		Spawn,
+		Default
+	};
+
+	const Type type;
+	const int timeout;
+	const QString server;
+	const QString nvim;
+	const QStringList positionalArgs;
+	const QStringList neovimArgs;
+};
+
+ConnectorInitArgs::Type getConnectorType(const QCommandLineParser &parser)
+{
+	if (parser.isSet("server")) {
+		return ConnectorInitArgs::Type::Server;
+	}
+
+	if (parser.isSet("embed")) {
+		return ConnectorInitArgs::Type::Embed;
+	}
+
+	if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
+		return ConnectorInitArgs::Type::Spawn;
+	}
+
+	return ConnectorInitArgs::Type::Default;
+}
+
+NeovimConnector *connectoToRemoteNeovim(const ConnectorInitArgs &args)
+{
+	NeovimConnector *connector{nullptr};
+	if (args.type == ConnectorInitArgs::Type::Embed) {
+		connector = NeovimQt::NeovimConnector::fromStdinOut();
+	}
+
+	if (args.type == ConnectorInitArgs::Type::Server) {
+		connector = NeovimQt::NeovimConnector::connectToNeovim(args.server);
+	}
+
+	if (args.type == ConnectorInitArgs::Type::Spawn) {
+		assert(!args.positionalArgs.isEmpty());
+		connector = NeovimQt::NeovimConnector::spawn(args.positionalArgs.mid(1),
+				args.positionalArgs.at(0));
+	}
+
+	if (!connector) {
+		connector = NeovimQt::NeovimConnector::spawn(
+				args.neovimArgs + args.positionalArgs, args.nvim);
+	}
+
+	connector->setRequestTimeout(args.timeout);
+	return connector;
+}
+
 /// A log handler for Qt messages, all messages are dumped into the file
 /// passed via the NVIM_QT_LOG variable. Some information is only available
 /// in debug builds (e.g. qDebug is only called in debug builds).
@@ -104,7 +163,15 @@ void App::showUi() noexcept
 		win->show();
 	}
 #else
-	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(m_connector.get());
+	auto connector = connectoToRemoteNeovim(ConnectorInitArgs{
+		getConnectorType(m_parser),
+		m_parser.value("timeout").toInt(),
+		m_parser.value("server"),
+		m_parser.value("nvim"),
+		m_parser.positionalArguments(),
+		getNeovimArgs()
+	});
+	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(connector);
 
 	// delete the main window when closed to emit `destroyed()` signal to
 	// support `:cq` return codes (Pull#644).
@@ -234,16 +301,6 @@ void App::checkArgumentsMayTerminate(QCommandLineParser& parser) noexcept
 	}
 }
 
-void App::setupRequestTimeout() noexcept
-{
-	if (!m_connector)
-	{
-		return;
-	}
-
-	m_connector->setRequestTimeout(m_parser.value("timeout").toInt());
-}
-
 /*static*/ QString App::getRuntimePath() noexcept
 {
 	QString path{ QString::fromLocal8Bit(qgetenv("NVIM_QT_RUNTIME_PATH")) };
@@ -272,38 +329,6 @@ void App::setupRequestTimeout() noexcept
 
 	return { "--cmd", QString{ "let &rtp.=',%1'" }.arg(runtimePath),
 		"--cmd","set termguicolors" };
-}
-
-void App::connectToRemoteNeovim() noexcept
-{
-	if (m_parser.isSet("embed")) {
-		m_connector = std::unique_ptr<NeovimConnector>{ NeovimQt::NeovimConnector::fromStdinOut() };
-		setupRequestTimeout();
-		return;
-	}
-
-	if (m_parser.isSet("server")) {
-		QString server = m_parser.value("server");
-		m_connector = std::unique_ptr<NeovimConnector>{ NeovimQt::NeovimConnector::connectToNeovim(server) };
-		setupRequestTimeout();
-		return;
-	}
-
-	if (m_parser.isSet("spawn") && !m_parser.positionalArguments().isEmpty()) {
-		const QStringList& args = m_parser.positionalArguments();
-		m_connector = std::unique_ptr<NeovimConnector> { NeovimQt::NeovimConnector::spawn(args.mid(1), args.at(0)) };
-		setupRequestTimeout();
-		return;
-	}
-
-	QStringList neovimArgs{ getNeovimArgs() };
-
-	// Append positional file arguments to nvim.
-	neovimArgs.append(m_parser.positionalArguments());
-
-	m_connector = std::unique_ptr<NeovimConnector>{  NeovimQt::NeovimConnector::spawn(neovimArgs, m_parser.value("nvim")) };
-	setupRequestTimeout();
-	return;
 }
 
 static QString GetNeovimVersionInfo(const QString& nvim) noexcept
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -32,6 +32,17 @@ struct ConnectorInitArgs {
 
 	}
 
+	ConnectorInitArgs(QStringList nvimArgs)
+		: type{Type::Default}
+		, timeout{2000}
+		, server{""}
+		, nvim{"nvim"}
+		, positionalArgs{}
+		, neovimArgs{std::move(nvimArgs)}
+	{
+
+	}
+
 	const Type type;
 	const int timeout;
 	const QString server;
@@ -375,6 +386,14 @@ void App::showVersionInfo(QCommandLineParser& parser) noexcept
 	out << GetNeovimVersionInfo(nvimExecutable) << QT_ENDL;
 
 	PrintInfo(versionInfo);
+}
+
+void App::openNewWindow() noexcept
+{
+	auto connector = connectToRemoteNeovim(ConnectorInitArgs{getNeovimArgs()});
+	Q_ASSERT(connector);
+	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(connector);
+	win->show();
 }
 
 void App::mainWindowClosing(int status) {
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -105,9 +105,10 @@ void onWindowActiveChanged(MainWindow& window)
 	s_lastActiveWindow = &window;
 }
 
-void onFileOpenEvent(MainWindow& window, const QList<QUrl>& files)
+void onFileOpenEvent(const QList<QUrl>& files)
 {
-	window.shell()->openFiles(files);
+	Q_ASSERT(s_lastActiveWindow);
+	s_lastActiveWindow->shell()->openFiles(files);
 }
 
 void onWindowClosing(int status) {
@@ -129,9 +130,6 @@ MainWindow* createWindow(NeovimConnector* connector)
 	App *app = qobject_cast<App *>(App::instance());
 	Q_ASSERT(app);
 
-	QObject::connect(
-			app, &App::openFilesTriggered, app,
-			[win](const QList<QUrl> &files) { onFileOpenEvent(*win, files); });
 	QObject::connect(win, &MainWindow::closing, app, onWindowClosing);
 	QObject::connect(win, &MainWindow::destroyed, app, [win]() {
 		Q_ASSERT(std::find(s_windows.cbegin(), s_windows.cend(), win) != s_windows.cend());
@@ -219,8 +217,6 @@ bool App::event(QEvent *event) noexcept
 		if(fileOpenEvent) {
 			emit openFilesTriggered({fileOpenEvent->url()});
 		}
-
-		return true;
 	}
 	else if (event->type() == QEvent::Quit) {
 		for (auto window : s_windows) {
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -255,8 +255,6 @@ void App::showUi() noexcept
 		win->show();
 	}
 #else
-	// FIXME: On macOS, Cmd+q kills the first window that was created.
-	// There doesn't seem to be a good order to how windows are closed.
 	auto win{createWindow(connectToRemoteNeovim(ConnectorInitArgs{ m_parser, getNeovimArgs() }))};
 
 	// delete the main window when closed to emit `destroyed()` signal to
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -126,10 +126,11 @@ void onWindowDestroyed() {
 MainWindow* createWindow(NeovimConnector* connector)
 {
 	Q_ASSERT(connector);
-	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(connector);
+
+	NeovimQt::MainWindow *win{new NeovimQt::MainWindow(connector)};
 	win->setAttribute(Qt::WA_DeleteOnClose);
 
-	App *app = qobject_cast<App *>(App::instance());
+	App *app{qobject_cast<App *>(App::instance())};
 	Q_ASSERT(app);
 
 	QObject::connect(win, &MainWindow::closing, app, onWindowClosing);
@@ -248,8 +249,7 @@ void App::showUi() noexcept
 #else
 	// FIXME: On macOS, Cmd+q kills the first window that was created.
 	// There doesn't seem to be a good order to how windows are closed.
-	auto *win = createWindow(
-			connectToRemoteNeovim(ConnectorInitArgs{m_parser, getNeovimArgs()}));
+	auto win{createWindow(connectToRemoteNeovim(ConnectorInitArgs{ m_parser, getNeovimArgs() }))};
 
 	// delete the main window when closed to emit `destroyed()` signal to
 	// support `:cq` return codes (Pull#644).
@@ -464,9 +464,9 @@ void App::openNewWindow(const QVariantList& args) noexcept
 		}
 	}
 
-	auto connector = connectToRemoteNeovim(ConnectorInitArgs{
-		type, 2000, std::move(server), std::move(nvim), getNeovimArgs()});
-	auto win = createWindow(connector);
+	NeovimConnector* connector{connectToRemoteNeovim(
+		ConnectorInitArgs{ type, 2000, std::move(server), std::move(nvim), getNeovimArgs() })};
+	MainWindow* win{createWindow(connector)};
 	win->resize(s_lastActiveWindow->size());
 	win->delayedShow();
 }
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -1,114 +1,127 @@
 #include "app.h"
 
+#include <algorithm>
 #include <QDir>
 #include <QFileInfo>
 #include <QFileOpenEvent>
 #include <QSettings>
 #include <QTextStream>
+#include <vector>
 
 #include "arguments.h"
 #include "mainwindow.h"
 #include "printinfo.h"
 #include "version.h"
 
-#include <vector>
-#include <functional>
+namespace NeovimQt {
 
 namespace {
-NeovimQt::MainWindow* s_lastActiveWindow{nullptr};
-std::vector<std::reference_wrapper<NeovimQt::MainWindow>> s_windows;
-int s_exitStatus{0};
 
-struct ConnectorInitArgs {
-	enum class Type {
+MainWindow* s_lastActiveWindow;
+std::vector<MainWindow*> s_windows;
+
+int s_exitStatus{ 0 };
+
+struct ConnectorInitArgs
+{
+	enum class Type
+	{
 		Embed,
 		Server,
 		Spawn,
 		Default
 	};
 
-	ConnectorInitArgs(const QCommandLineParser& parser, QStringList nvimArgs) noexcept
-		: type{ getConnectorType(parser) }
-		, timeout{ parser.value("timeout").toInt() }
-		, server{ parser.value("server") }
-		, nvim{ parser.value("nvim") }
-		, positionalArgs{ parser.positionalArguments() }
-		, neovimArgs{ std::move(nvimArgs) }
-	{
-
-	}
-
-	ConnectorInitArgs(
-		Type _type, int _timeout, QString _server, QString _nvim, QStringList nvimArgs) noexcept
-		: type{ _type }
-		, timeout{ _timeout }
-		, server{ std::move(_server) }
-		, nvim{ std::move(_nvim) }
-		, positionalArgs{}
-		, neovimArgs{std::move(nvimArgs)}
-	{
-	}
-
-	const Type type;
-	const int timeout;
+	const Type type{ Type ::Default };
+	const int timeout{ 10000 };
 	const QString server;
 	const QString nvim;
 	const QStringList positionalArgs;
 	const QStringList neovimArgs;
 
-	static Type getConnectorType(const QCommandLineParser& parser) noexcept
-	{
-		if (parser.isSet("server")) {
-			return ConnectorInitArgs::Type::Server;
-		}
+	ConnectorInitArgs(const QCommandLineParser& parser, QStringList nvimArgs) noexcept;
 
-		if (parser.isSet("embed")) {
-			return ConnectorInitArgs::Type::Embed;
-		}
-
-		if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
-			return ConnectorInitArgs::Type::Spawn;
-		}
-
-		return ConnectorInitArgs::Type::Default;
-	}
+	ConnectorInitArgs(
+		Type _type, int _timeout, QString _server, QString _nvim, QStringList nvimArgs) noexcept;
 };
 
-NeovimQt::NeovimConnector* connectToRemoteNeovim(const ConnectorInitArgs& args) noexcept
+ConnectorInitArgs::Type getConnectorType(const QCommandLineParser& parser) noexcept
 {
-	using namespace NeovimQt;
-	NeovimConnector *connector{nullptr};
-	if (args.type == ConnectorInitArgs::Type::Embed) {
-		connector = NeovimConnector::fromStdinOut();
+	if (parser.isSet("server")) {
+		return ConnectorInitArgs::Type::Server;
 	}
 
-	if (args.type == ConnectorInitArgs::Type::Server) {
-		connector = NeovimConnector::connectToNeovim(args.server);
+	if (parser.isSet("embed")) {
+		return ConnectorInitArgs::Type::Embed;
 	}
 
-	if (args.type == ConnectorInitArgs::Type::Spawn) {
-		Q_ASSERT(!args.positionalArgs.isEmpty());
-		connector = NeovimConnector::spawn(args.positionalArgs.mid(1),
-				args.positionalArgs.at(0));
+	if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
+		return ConnectorInitArgs::Type::Spawn;
 	}
+
+	return ConnectorInitArgs::Type::Default;
+}
+
+ConnectorInitArgs::ConnectorInitArgs(const QCommandLineParser& parser, QStringList nvimArgs) noexcept
+	: type{ getConnectorType(parser) }
+	, timeout{ parser.value("timeout").toInt() }
+	, server{ parser.value("server") }
+	, nvim{ parser.value("nvim") }
+	, positionalArgs{ parser.positionalArguments() }
+	, neovimArgs{ std::move(nvimArgs) }
+{
+}
+
+ConnectorInitArgs::ConnectorInitArgs(
+	Type _type, int _timeout, QString _server, QString _nvim, QStringList nvimArgs) noexcept
+	: type{ _type }
+	, timeout{ _timeout }
+	, server{ std::move(_server) }
+	, nvim{ std::move(_nvim) }
+	, neovimArgs{ std::move(nvimArgs) }
+{
+}
+
+NeovimConnector& connectToRemoteNeovim(const ConnectorInitArgs& args) noexcept
+{
+	NeovimConnector* connector{ nullptr };
+
+	switch (args.type)
+	{
+		case ConnectorInitArgs::Type::Embed:
+			connector = NeovimConnector::fromStdinOut();
+			break;
+
+		case ConnectorInitArgs::Type::Server:
+			connector = NeovimConnector::connectToNeovim(args.server);
+			break;
+
+		case ConnectorInitArgs::Type::Spawn:
+			if (args.positionalArgs.size() >= 2)
+			{
+				connector = NeovimConnector::spawn(args.positionalArgs.mid(1), args.positionalArgs.at(0));
+			}
+			break;
+
+		case ConnectorInitArgs::Type::Default:
+			break;
+	};
 
 	if (!connector) {
-		connector = NeovimConnector::spawn(
-				args.neovimArgs + args.positionalArgs, args.nvim);
+		connector = NeovimConnector::spawn(args.neovimArgs + args.positionalArgs, args.nvim);
 	}
 
 	connector->setRequestTimeout(args.timeout);
-	return connector;
+	return *connector;
 }
 
-void onWindowActiveChanged(NeovimQt::MainWindow& window) noexcept
+void onWindowActiveChanged(MainWindow& window) noexcept
 {
 	s_lastActiveWindow = &window;
 }
 
 void onFileOpenEvent(const QList<QUrl>& files) noexcept
 {
-	Q_ASSERT(s_lastActiveWindow);
 	s_lastActiveWindow->shell()->openFiles(files);
 }
 
@@ -117,44 +130,34 @@ void onWindowClosing(int status) noexcept
 	s_exitStatus = status;
 }
 
-void onWindowDestroyed() noexcept
+void onWindowDestroyed(QObject* obj) noexcept
 {
+	s_windows.erase(std::remove(s_windows.begin(), s_windows.end(), (MainWindow*)obj));
+
 	if (s_windows.empty()) {
 		qApp->exit(s_exitStatus);
 	}
 }
 
-NeovimQt::MainWindow* createWindow(NeovimQt::NeovimConnector* connector) noexcept
+MainWindow& createWindow(const ConnectorInitArgs& args) noexcept
 {
-	using namespace NeovimQt;
-	Q_ASSERT(connector);
+	NeovimConnector& conn{ connectToRemoteNeovim(args) };
+	MainWindow* win{ new MainWindow(&conn) };
 
-	MainWindow *win{new MainWindow(connector)};
 	win->setAttribute(Qt::WA_DeleteOnClose);
 
-	App *app{qobject_cast<App *>(App::instance())};
-	Q_ASSERT(app);
+	App* app{ qobject_cast<App*>(App::instance()) };
 
 	QObject::connect(win, &MainWindow::closing, app, onWindowClosing);
-    QObject::connect(win, &MainWindow::destroyed, app, [win]() {
-        auto foundIt = std::find_if(
-            s_windows.cbegin(), s_windows.cend(),
-            [win](const std::reference_wrapper<NeovimQt::MainWindow>& rw) {
-                return win == &rw.get();
-            });
-		Q_ASSERT(foundIt != s_windows.cend());
-		s_windows.erase(foundIt);
-		onWindowDestroyed();
-	});
+	QObject::connect(win, &MainWindow::destroyed, app, onWindowDestroyed);
 	QObject::connect(win, &MainWindow::activeChanged, app, onWindowActiveChanged);
 
 	s_lastActiveWindow = win;
-	s_windows.push_back(*win);
-	return win;
+	s_windows.emplace_back(win);
+	return *win;
 }
-} // namespace
 
-namespace NeovimQt {
+} // namespace
 
 /// A log handler for Qt messages, all messages are dumped into the file
 /// passed via the NVIM_QT_LOG variable. Some information is only available
@@ -223,15 +226,15 @@ App::App(int &argc, char ** argv) noexcept
 	processCommandlineOptions(m_parser, arguments());
 }
 
-bool App::event(QEvent *event) noexcept
+bool App::event(QEvent* event) noexcept
 {
-	if( event->type()  == QEvent::FileOpen) {
-		QFileOpenEvent * fileOpenEvent = static_cast<QFileOpenEvent *>(event);
-		onFileOpenEvent({fileOpenEvent->url()});
+	if (event->type() == QEvent::FileOpen) {
+		QFileOpenEvent* fileOpenEvent{ static_cast<QFileOpenEvent*>(event) };
+		onFileOpenEvent({ fileOpenEvent->url() });
 	}
 	else if (event->type() == QEvent::Quit) {
 		for (auto window : s_windows) {
-			if (!window.get().close()) {
+			if (!window->close()) {
 				event->setAccepted(false);
 			}
 		}
@@ -255,7 +258,9 @@ void App::showUi() noexcept
 		win->show();
 	}
 #else
-	auto win{createWindow(connectToRemoteNeovim(ConnectorInitArgs{ m_parser, getNeovimArgs() }))};
+
+	ConnectorInitArgs args{ m_parser, getNeovimArgs() };
+	MainWindow* win{ &createWindow(args) };
 
 	// delete the main window when closed to emit `destroyed()` signal to
 	// support `:cq` return codes (Pull#644).
@@ -455,24 +460,24 @@ void App::showVersionInfo(QCommandLineParser& parser) noexcept
 void App::openNewWindow(const QVariantList& args) noexcept
 {
 	QString server;
-	QString nvim{"nvim"};
-	ConnectorInitArgs::Type type{ConnectorInitArgs::Type::Default};
-	if (args.size() > 1) {
-		Q_ASSERT(args.at(1).type() == QVariant::Type::Map);
-		auto initMap = args.at(1).toMap();
-		if (initMap.find("nvim") != initMap.end()) {
-			nvim = initMap["nvim"].toString();
+	QString nvim{ "nvim" };
+	ConnectorInitArgs::Type type{ ConnectorInitArgs::Type::Default };
+	if (args.size() > 1 && args.at(1).type() == QVariant::Type::Map) {
+
+		const QVariantMap initMap{ args.at(1).toMap() };
+
+		if (initMap.contains("nvim")) {
+			nvim = initMap.value("nvim").toString();
 		}
 
-		if (initMap.find("server") != initMap.end()) {
-			server = initMap["server"].toString();
+		if (initMap.contains("server")) {
+			server = initMap.value("server").toString();
 			type = ConnectorInitArgs::Type::Server;
 		}
 	}
 
-	NeovimConnector* connector{connectToRemoteNeovim(
-		ConnectorInitArgs{ type, 2000, std::move(server), std::move(nvim), getNeovimArgs() })};
-	MainWindow* win{createWindow(connector)};
+	ConnectorInitArgs params{ type, 2000, std::move(server), std::move(nvim), getNeovimArgs() };
+	MainWindow* win{ &createWindow(params) };
 	win->resize(s_lastActiveWindow->size());
 	win->delayedShow();
 }
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -21,30 +21,41 @@ struct ConnectorInitArgs {
 		Default
 	};
 
+	ConnectorInitArgs(const QCommandLineParser& parser, QStringList nvimArgs)
+		: type{getConnectorType(parser)}
+		, timeout{parser.value("timeout").toInt()}
+		, server{parser.value("server")}
+		, nvim{parser.value("nvim")}
+		, positionalArgs{parser.positionalArguments()}
+		, neovimArgs{std::move(nvimArgs)}
+	{
+
+	}
+
 	const Type type;
 	const int timeout;
 	const QString server;
 	const QString nvim;
 	const QStringList positionalArgs;
 	const QStringList neovimArgs;
+
+	Type getConnectorType(const QCommandLineParser &parser)
+	{
+		if (parser.isSet("server")) {
+			return ConnectorInitArgs::Type::Server;
+		}
+
+		if (parser.isSet("embed")) {
+			return ConnectorInitArgs::Type::Embed;
+		}
+
+		if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
+			return ConnectorInitArgs::Type::Spawn;
+		}
+
+		return ConnectorInitArgs::Type::Default;
+	}
 };
-
-ConnectorInitArgs::Type getConnectorType(const QCommandLineParser& parser)
-{
-	if (parser.isSet("server")) {
-		return ConnectorInitArgs::Type::Server;
-	}
-
-	if (parser.isSet("embed")) {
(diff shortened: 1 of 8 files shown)


jgehrig:
@Furkanzmc

Thanks for the Pull Request! :+1:

I agree. The multi-window behavior on MacOS is in need of improvement, and this is a valuable feature to work on.

> This kind of breaks the :cq support when there's multiple windows, though.

Some additional details would be helpful here. Why can't we support both?

I don't think breaking Linux/Windows `:cq` support is a good trade-off for a new MacOS-only feature.

Ideally, we should add MacOS multi-window support without breaking `:cq`. If this is not possible on MacOS, we should at least avoid regressing `:cq` on Windows/Linux.

> I think I'll open an issue about the multiple window support because there are some things that has to be considered before it can be implemented.

Yes, please do this. That will give us a place to discuss the design.

It would be helpful to understand the "big picture" before merging the trivial refactor:
 - Why is this refactor necessary?
 - How will we support this feature?
 - What changes are required?
 - How will we avoid breaking non-MacOS platforms?

equalsraf:
From what I understand you want neovim-qt to support multiple top-level windows - primarily because Mac OS X has special handling for multiple windows in the same process (for window switching).

Internally you want to launch multiple connectors+mainwindows in the same process i.e. our GUI would get the ability to serve multiple neovim instances.

The reason why I think that this is so hard is that in Mac OS X window grouping behaves differently, if:

1. Multiple windows belong to the same process get one icon in the dock (switch with Cmd+`)
2. Multiple windows from different process get multiple icons in the dock (switch with Cmd+Tab) even if both process are different instances of the same application

(Are my assumptions here wrong? This is as far as i understand Mac OS X - is there a way to change this behavior per app?)

Like you said, I don't see a way to do this without breaking `:cq`. Even if we introduced a flag or a command to enable this mode, calling `:q` would then close a window and kill one neovim process - the rest would remain. This is probably ok-ish as long as it is opt-in.

I am not sure if this breaks any assumptions in the code, I had a quick look for:

- calls to QApplication::quit or similarly, calls to QApplication/QCoreApplication
- instances that are not cleaned on exit (you dont want leave a dangling connection/qprocess)

QFileOpenEvent was the first issue that came to mind. Maybe App::showUi is affected too.

---------------

Some thoughts

1. how do we start a new window? Vim commands?
2. Calls to `open -a nvim-qt filename` are currently opening a new buffer in the current instance. What will happen to this with a change like this?
3. support for non-Mac OS, this depends a lot on how you answer the the first question

This sounds like a big structural change for something that is pretty ingrained in Mac OS X (one window per app - or window grouping). It would be nice to have an idea on how other apps in Mac OS X handle this issue too.

I can think of a few cases when you want multiple nvim instances and or multiple top level windows, but I'm not sure if these match your motivation:

1. run multiple nvim instances, one in each container/machine/etc
2. neovim isolation is not perfect - e.g. some features are not 100% independent per tab , or i dont want to list all buffers just the buffers i am working on right now - this is probably addressable in neovim, however as a consequence people run multiple neovim instances anyway.
3. multiple screens - however in this case what people usually wan is one neovim instance for multiple top level windows

Just for context - on the opposite direction - one of the features in neovim that was supposed to help here
was [multigrid](https://neovim.io/doc/user/ui.html#ui-multigrid) - i.e. the UI chooses how to layout neovim windows into actual UI windows - it could be 1-1 or something else, but this retains a single nvim process.

Furkanzmc:
The commit messages should have answere these questions, sorry @jgehrig.
I edited the commit messages to include this information as well.

> It would be helpful to understand the "big picture" before merging the trivial refactor:
> 
>     * Why is this refactor necessary?

It's necessary to be able to create multiple connections per window. Right now the app owns it and destroys it when it exists. But when there's multiple windows, there can't be one connection and the app can't own the connection since the first window might be closed before the newly created windows.

>     * How will we support this feature?

I think this would be specific to macOS as I don't see any use for this on other platforms. Do I understand your question correctly? Does this answer it?

>     * What changes are required?

I'll need to investigate this more with a future PR because simply having the ability to create multiple windows is not enough. I'll put the possible changes in a new issue.

>     * How will we avoid breaking non-MacOS platforms?

I think we should probably only compile this feature for macOS.

-----

@equalsraf 

> The reason why I think that this is so hard is that in Mac OS X window grouping behaves differently, if:
> 
>     1. Multiple windows belong to the same process get one icon in the dock (switch with Cmd+`)
> 
>     2. Multiple windows from different process get multiple icons in the dock (switch with Cmd+Tab) even if both process are different instances of the same application
> 
> (Are my assumptions here wrong? This is as far as i understand Mac OS X - is there a way to change this behavior per app?)

These assumptions are correct. Option number one also enables app expose.

> Like you said, I don't see a way to do this without breaking `:cq`. Even if we introduced a flag or a command to enable this mode, calling `:q` would then close a window and kill one neovim process - the rest would remain. This is probably ok-ish as long as it is opt-in.
> 
> I am not sure if this breaks any assumptions in the code, I had a quick look for:
> 
>     * calls to QApplication::quit or similarly, calls to QApplication/QCoreApplication
> 
>     * instances that are not cleaned on exit (you dont want leave a dangling connection/qprocess)
> 
> 
> QFileOpenEvent was the first issue that came to mind. Maybe App::showUi is affected too.

Those are the exact same places I looked for as well. But I haven't really ironed out the implementation in my mind yet. I wanted to get my feet wet with a PR and open up a discussion before I invest in it.

I'll answer the other question on a separate issue #850 so that I don't pollute this PR. I don't think the PR breaks anything even if the multi window support is not implemented.

jgehrig:
Just to be clear :)

This is not merge-ready, since it breaks `:cq` support. We should not change behavior during the refactor.

Later, we can change `:cq` behavior as needed for the feature-work.

Furkanzmc:
> Just to be clear :)
> 
> 
> 
> This is not merge-ready, since it breaks `:cq` support. We should not change behavior during the refactor.
> 
> 
> 
> Later, we can change `:cq` behavior as needed for the feature-work.

I know. I haven't had a chance to fix the tests yet. 😄

Furkanzmc:
I assumed that the code was not reviewed because of the failing test so I just force pushed my changes. I hope that's not against the rules, I didn't find anything against it.

jgehrig:
Definitely feel free to `git push -f`!

Actually, I prefer it... I find the code easier to read/merge when the commits are tidy.

src/gui/mainwindow.cpp
```diff
@@ -24,16 +24,10 @@ MainWindow::MainWindow(NeovimConnector* c, QWidget* parent)
 
 void MainWindow::init(NeovimConnector *c)
 {
-	if (m_shell) {
-		m_shell->deleteLater();
-		m_stack.removeWidget(m_shell);
-	}
+	assert(!m_shell);
```
jgehrig, line 27: This isn't functionally equivalent. The code before will gracefully handle `!m_shell`. This code will abort with an error.

I think this block is necessary for when the `ErrorWidget` displays on connection failure?

src/gui/mainwindow.cpp
```diff
@@ -249,9 +244,21 @@ void MainWindow::neovimGuiCloseRequest(int status)
 
 void MainWindow::reconnectNeovim()
 {
+	assert(m_nvim);
+	assert(m_shell);
```
jgehrig, line 248: Is this safe?

Why not write defensive code and explicitly handle `!m_nvim || !m_shell`?

src/gui/app.cpp
```diff
@@ -13,6 +13,65 @@
...
+
+	const Type type;
+	const int timeout;
+	const QString server;
+	const QString nvim;
+	const QStringList positionalArgs;
+	const QStringList neovimArgs;
+};
+
+ConnectorInitArgs::Type getConnectorType(const QCommandLineParser &parser)
```
jgehrig, line 32: nitpick: `CommandLineParser& parser`

src/gui/app.cpp
```diff
@@ -13,6 +13,65 @@
 
 namespace NeovimQt {
 
+struct ConnectorInitArgs {
+	enum Type {
```
jgehrig, line 17: `enum class Type : int`

With enum class the compiler with enforce stronger type checks.

src/gui/app.cpp
```diff
@@ -13,6 +13,65 @@
...
+		Embed,
+		Server,
+		Spawn,
+		Default
+	};
+
+	const Type type;
+	const int timeout;
```
jgehrig, line 25: Provide a reasonable default value:
`const int timeout{ ??? };`

Necessary for primitive types.

src/gui/app.cpp
```diff
@@ -13,6 +13,65 @@
...
+	const QString server;
+	const QString nvim;
+	const QStringList positionalArgs;
+	const QStringList neovimArgs;
```
jgehrig, line 29: Let's provide a CTOR and delete the default CTOR.

```C++
ConnectorInitArgs(const QCommandLineParser& parser, const QStringList& neovimArgs) noexcept
{
..
}
```

Use the prefix `m_` for member variables.

Consider a class, and moving this to its own file if complexity increases. A struct here is fine for now.

src/gui/app.cpp
```diff
@@ -13,6 +13,65 @@
...
+	}
+
+	if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
+		return ConnectorInitArgs::Type::Spawn;
+	}
+
+	return ConnectorInitArgs::Type::Default;
+}
+
+NeovimConnector *connectoToRemoteNeovim(const ConnectorInitArgs &args)
```
jgehrig, line 49: `NeovimConnector*` and `connectToRemoteNeovim`

src/gui/app.cpp
Furkanzmc: Is there a clang-format or some other format configuration file I can use? I didn't know what the style was so just ended up using the default clang-format abs adapting it a little bit so it looks  similar to the rest of the code. I missed this one.

src/gui/app.cpp
Furkanzmc: Isn't the default value always provided by the command line parser?

src/gui/app.cpp
Furkanzmc: I was intending this to be a very simple  data structure to hold on to the initialization invariants. If I add a ctor that takes a parser, I would need to create another ctor later when a different configuration is used (e.g multiple windows). 

How about a function that returns this struct from a parser?

src/gui/mainwindow.cpp
Furkanzmc: I usually avoid that pattern as it blurs the line between a bug and an error. But I won't go into a philosophical debate about it. :) I'll make the change.

Thank you for reviewing my code.

Furkanzmc:
I don't understand why the Appveyor build is failing. Can you point me in the right direction there?

src/gui/mainwindow.cpp
jgehrig: I completely agree philosophically :)

What I'm not sure: will this change trigger a segfault that did not occur before? Seems possible.

Generally, I like to use a reference for non-nullable values and pointers as optional values.

equalsraf:
Cant have a deep look right now, but the failure is this one (in tst_shell.cpp)

```
6: QWARN  : NeovimQt::Test::startVarsMainWindow() qt.svg: Cannot open file ':/neovim.svg', because: No such file or directory
6: PASS   : NeovimQt::Test::startVarsMainWindow()
6: PASS   : NeovimQt::Test::guiExtTablineSet()
6: PASS   : NeovimQt::Test::gviminit()
6: QWARN  : NeovimQt::Test::guiShimCommands() qt.svg: Cannot open file ':/neovim.svg', because: No such file or directory
6: FAIL!  : NeovimQt::Test::guiShimCommands() 'SPYWAIT(cmd_gf2)' returned FALSE. ()
6: C:\projects\neovim-qt\test\tst_shell.cpp(175) : failure location
```

I dont know if the svg warning was happening before, but it is not the cause of the test failure (we should still check).

The failure happens when executing a GuiFont command - and apparently the command did not finish. The timeout is 30s and a previous similar command did work. But the exact same issue happens in both executions so i dont think it is a timeout/fluke.

I would need to look into the patch, could there be something severing the signal or replacing the neovimObject?

equalsraf:
If anyone is looking into that failure, I have built windows artifacts https://ci.appveyor.com/project/equalsraf/neovim-qt/builds/39169706/job/r18xr0v0tjuk13e3/artifacts

jgehrig:
> I dont know if the svg warning was happening before, but it is not the cause of the test failure (we should still check).

This is unrelated it has existed for some time. The issue is related to missing resources in the CMake logic for tst_shell.cpp.

We could fix this by providing `src/gui/data.qrc` to tst_shell. Safe to ignore.

src/gui/mainwindow.cpp
Furkanzmc: > What I'm not sure: will this change trigger a segfault that did not occur before? Seems possible.

Just looking at the related code, `ErrorWidget::reconnectNeovim` is emitted when the retry button is clicked. And right after creating the widget `MainWindow::init` is already called. So, it seems that there's no way that it would and if it does it's certainly a bug.

I think it's safe to leave it as it is, but if you still think I should add the additional check, I will.

src/gui/mainwindow.cpp
Furkanzmc: I think the initialization code should just do that. Handling the case where `m_shell` and `m_nvim` is not properly uninitialized could hide some other bugs.

> I think this block is necessary for when the ErrorWidget displays on connection failure?

It doesn't seem to effect it. Like I said in my other comment, by the time the error widget is shown the init is called and if it didn't do what it's supposed to, retrying properly releases the previous resources and creates new ones.

src/gui/app.cpp
jgehrig: A few comments:
1. I'd strongly encourage a class w/ getters/setters. A struct is harder to expand in the future. If you prefer a simple struct, that's fine.
2. A CTOR will not increase the complexity. Multiple CTORs is fine. A CTOR is the correct C++ mechanism for what you're doing.
3. Yeah, I suppose `m_` is a little awkward for public struct members. Feel free to ignore `m_`.

> How about a function that returns this struct from a parser?

Well, that's a CTOR... Right?

Writing a CTOR instead of a function returning the object allows the compiler to perform additional optimizations. Your code is all great, it simply needs to be moved into the CTOR.

```
struct struct ConnectorInitArgs
{
    static Type getConnectorType(const QCommandLineParser& parser) noexcept;

    ConnectorInitArgs() noexcept = delete;

    ConnectorInitArgs(const QCommandLineParser& parser, const QStringList& neovimArgs) noexcept
        : server{ ... }
	, nvim{ ... }
	, positionalArgs{ ... }
	, neovimArgs{ ... }
    { ... }
    ...
```

A CTOR helps prevent contributors in the future from introducing bugs.

Ex) It is very easy to accidentally switch the ordering of variables and break your code.

src/gui/app.cpp
jgehrig: Yes. Your code is logically sound.

This is about writing future-proof code. Primitive types such as `int` will not be zero initialized. Providing a default value helps future authors from accidentally introducing non-deterministic behavior. Hard to catch, hard to debug.

src/gui/mainwindow.cpp
jgehrig: Please use the old logic here.

Let's not change things we don't have to during the refactor.

src/gui/mainwindow.cpp
```diff
@@ -249,9 +244,21 @@ void MainWindow::neovimGuiCloseRequest(int status)
...
-		init(m_nvim->reconnect());
+		m_shell->deleteLater();
+		m_stack.removeWidget(m_shell);
+		m_shell = nullptr;
+
+		auto connector = m_nvim->reconnect();
+		m_nvim->deleteLater();
+		m_nvim = nullptr;
+
+		init(connector);
```
jgehrig, line 259: Please revert this block, no need to change the existing logic.

src/gui/mainwindow.cpp
jgehrig: The `m_nvim` assert is valid I think, since we deference `m_nvim` below.

There is no need for the `m_shell` assert. The existing code is valid, and less strict. A future author could easily make a mistake with the new structure and introduce a case that crashes. Let's avoid this.

`Q_ASSERT` is prefered in Qt code. It provides better error handling, and will be disabled in `Release` mode.

Furkanzmc: > There is no need for the m_shell assert. The existing code is valid, and less strict. A future author could easily make a mistake with the new structure and introduce a case that crashes. Let's avoid this.

I'll make the changes. I just want to comment as a professional curiosity.

Isn't a case that crashes better than a case that hides one? If that mistake is made, and it's hidden the system is basically in an unstable state because there's no telling when the crash is going to exhibit itself. 

Again, I don't want to turn it into a big thing. Just taking this as an opportunity to learn.

jgehrig: No big deal, I'm happy to discuss. Discussion is good :)

Yes, that's generally good practice.

Here, you're adding a requirement to the code which did not exist before. The requirement that m_shell/m_nvim is or is not nullptr is not obvious from the API (a non-null ptr should be a reference). The requirement is easy to miss.

What does this change accomplish? The old code is more tolerant of fault cases and does not leak. The new code has cases where a possible state will trigger asset(false). Right?

Furkanzmc: > What does this change accomplish? The old code is more tolerant of fault cases and does not leak. The new code has cases where a possible state will trigger asset(false). Right?

That's right. And it's exactly the reason why it's useful. The requirement that an initialization should only happen if it has not been initialized before is a pretty obvious one. Otherwise, what's stopping me from calling `init` in a loop over and over again because for some reason my code worked if I called it twice but not when I called it once? Assertions make it easier to not write lazy code.

I find that assertions make a code base more robust and less error prone. My life has been much easier when QA or another developer comes to be with an assertion than with a description of a problem. Not that you don't agree with this already...

-----

On another note, I fixed all your comments and pushed the commits. This has been a great experience as a first time contributor, I appreciate it.

jgehrig:
@Furkanzmc 

Looking at this code one last time before merge...

Is there a reason we need `ConnectorInitArgs` at all?

It seems like we can further simply this code by making `connectToRemoveNeovim` `static`. We simply return `m_connector` instead, and pass `m_parser` as a parameter. Simple change, less risk.

If we were writing this from scratch today, I would do something similar to your code. However, I think there is value in minimizing the amount of code we change. The existing code/logic is proven.

Is the following compatible with your planned MacOS window changes?
https://github.com/jgehrig/neovim-qt/commit/e912272994ee35c2437749a9c9810854de270ff6

Let me know and I can merge https://github.com/jgehrig/neovim-qt/commit/e912272994ee35c2437749a9c9810854de270ff6.

Furkanzmc:
> @Furkanzmc 
> 
> 
> 
> Looking at this code one last time before merge...
> 
> 
> 
> Is there a reason we need `ConnectorInitArgs` at all?
> 
> 
> 
> It seems like we can further simply this code by making `connectToRemoveNeovim` `static`. We simply return `m_connector` instead, and pass `m_parser` as a parameter. Simple change, less risk.

I will need that struct in order to spawn a new window with new initialization arguments. If it only takes a parser, I would need to create a new function for the multiple window support or change that one. 

I've been using the refactored version for a while and haven't encountered any problems.

jgehrig:
> I've been using the refactored version for a while and haven't encountered any problems.

Acknowledged.

> I will need that struct in order to spawn a new window with new initialization arguments. If it only takes a parser, I would need to create a new function for the multiple window support or change that one.

Are you sure? Why wouldn't the following work?

```
NeovimConnector* connector1{ connectToRemoteNeovim(m_parser) };
NeovimQt::MainWindow* win1{ new NeovimQt::MainWindow(connector) };
...

NeovimConnector* connector2{ connectToRemoteNeovim(m_parser) };
NeovimQt::MainWindow* win2{ new NeovimQt::MainWindow(connector) };
...
```

From my read, `ConnectorInitArgs` simply copies content from `m_parser`. This copy and parsing are both non-trivail. Why not simply call the existing function (minor modification to make it static) with `m_parser` directly?

Such a solution is simpler, faster, uses proven code, and requires minimal modification.

--- 

Alternatively.

> I've been using the refactored version for a while

It sounds like you already have something written.

Please point me to a commit, a prototype working or not working or not is fine.

Seeing where you're taking with this would make me more comfortable with a merge.

Furkanzmc:
> > I will need that struct in order to spawn a new window with new initialization arguments. If it only takes a parser, I would need to create a new function for the multiple window support or change that one.
> 
> 
> 
> Are you sure? Why wouldn't the following work?
> 
> 
> 
> ```
> 
> NeovimConnector* connector1{ connectToRemoteNeovim(m_parser) };
> 
> NeovimQt::MainWindow* win1{ new NeovimQt::MainWindow(connector) };
> 
> ...
> 
> 
> 
> NeovimConnector* connector2{ connectToRemoteNeovim(m_parser) };
> 
> NeovimQt::MainWindow* win2{ new NeovimQt::MainWindow(connector) };
> 
> ...
> 
> ```
> 
> 
> 
> From my read, `ConnectorInitArgs` simply copies content from `m_parser`. The copy and parsing are both non-trivial. Why not simply call the existing function (minor modification to make it static) with `m_parser` directly?
> 
> 
> 
> Such a solution is simpler, faster, uses proven code, and requires minimal modification.

Because it would launch a copy of the existing window which is not what we necessarily want. What if they connected to a server and are now creating a new window that's separate? Or a new window that connects to a separate server. It limits us to the arguments that were passed in when launching the program. But we want to be able to modify so that we Dan decide what kind of new window we want to create. 
> 
> --- 
> 
> 
> 
> Alternatively.
> 
> 
> 
> > I've been using the refactored version for a while
> 
> 
> 
> It sounds like you already have something written.
> 
> 
> 
> Please point me to a commit, a prototype working or not is fine.
> 
> 
> 
> Seeing where you're taking with this would make me more comfortable with a merge.

I meant I've been using the build from my branch for this PR. There's nothing separate.

jgehrig:
> I meant I've been using the build from my branch for this PR. There's nothing separate.

Sorry, I misunderstood.

> Because it would launch a copy of the existing window which is not what we necessarily want. What if they connected to a server and are now creating a new window that's separate? Or a new window that connects to a separate server. It limits us to the arguments that were passed in when launching the program. But we want to be able to modify so that we Dan decide what kind of new window we want to create.

Your changes are difficult to follow without a more detailed multi-window design/plan. I'm assuming you'll want to re-use the existing argument parsing code. Maybe this wasn't your intent?

Perhaps it makes sense to lay out a rough plan of your design in #850? That would help me understand...

I still think my change above is capable of all of the things you've listed. Here is a more detailed example:
```
void CreateNewWindowWithArbitraryArguments() noexcept
{
	QCommandLineParser p;
	QStringList arbitraryArguments{ "-arg1", "value1", "-arg2", "value2", "-arg3" }
	NeovimQt::App::processCommandlineOptions(p, arbitraryArguments);
	
	// The line below is tricky... It exits on error. You'll need to address this in either approach (later).
	// Probably return an error code enum class instead of exiting within the function?
	NeovimQt::App::checkArgumentsMayTerminate(p);

	NeovimConnector* connector1{ connectToRemoteNeovim(p) };
	NeovimQt::MainWindow* newWinArbitraryArguments{ new NeovimQt::MainWindow(connector) };
}
```

Thanks for the continued dialogue and patience!

Furkanzmc:
> > Because it would launch a copy of the existing window which is not what we necessarily want. What if they connected to a server and are now creating a new window that's separate? Or a new window that connects to a separate server. It limits us to the arguments that were passed in when launching the program. But we want to be able to modify so that we Dan decide what kind of new window we want to create.
> 
> Your changes are difficult to follow without a more detailed multi-window design/plan. I'm assuming you'll want to re-use the existing argument parsing code. Maybe this wasn't your intent?
> 
> Perhaps it makes sense to lay out a rough plan of your design in #850? That would help me understand...

My changes only allows a decoupling of the command line arguments from the initialization arguments. The while reason for this refctor was to enable a way to create multiple Neovim connections. Before this PR, the connection is unique and bound to the app instance, whereas in this one there's multiple and is bound to the window. I'm not sure what else to add to the issue on top of the existing information there. I think you and equalsraf also summed up the situation there. I haven't had time to respond to those comments but I will.

> I still think my change above is capable of all of the things you've listed. Here is a more detailed example:
> 
> ```
> void CreateNewWindowWithArbitraryArguments() noexcept
> {
> 	QCommandLineParser p;
> 	QStringList arbitraryArguments{ "-arg1", "value1", "-arg2", "value2", "-arg3" }
> 	NeovimQt::App::processCommandlineOptions(p, arbitraryArguments);
> 	
> 	// The line below is tricky... It exits on error. You'll need to address this in either approach (later).
> 	// Probably return an error code enum class instead of exiting within the function?
> 	NeovimQt::App::checkArgumentsMayTerminate(p);
> 
> 	NeovimConnector* connector1{ connectToRemoteNeovim(p) };
> 	NeovimQt::MainWindow* newWinArbitraryArguments{ new NeovimQt::MainWindow(connector) };
> }
> ```

It does support it, but to me creating a command line parser for initialization of an object doesn't seem right. I don't see the parser as a way to hold data but to transfer data from the launch arguments to the program. It looks like a misuse to me. Why do I have to copy these string arguments to another function and then create a command line parser when I'm actually not parsing the command line arguments? Why wouldn't I just pass those arguments directly instead of doing an unnecessary class construction? Having that struct to initialize new connection seems like a better and more statically correct way to me.

If you still prefer it that way, I can disagree and commit. :D

jgehrig:
All sounds reasonable to me. But you need to explain what your plan is before we re-write widely used code :)

How do you propose creating new windows if not via the existing commandline argument mechanisms?

Please do one of the following:
- Explain your overall plans in more detail here or in #850. We'll merge this once a plan is formulated.
- Assume this commit is good and build off it in this PR. We'll merge once more code is written, and I understand the plan.

The code is reasonable, I don't see any major issues. However, we should all be on the same page before we start merging to `master`. Feel free to force-push and edit history on this branch.

Does this sound reasonable?

Furkanzmc:
Sorry it took a while to finish the PR. I've been using the implementation locally on my machine to see If it works as we intended, and I'm happy to say I have not encountered any problems.

One thing you'll notice is the `Cmd+Q` behavior when there's a window with an unsaved buffer. It'll cause the default prompt to show up (save/cancel/don't save), and if it's cancelled the window won't be closed and as a result the application won't be closed.

`open -a` works as we discussed.

src/gui/app.cpp
```diff
@@ -1,18 +1,150 @@
...
-#include "mainwindow.h"
-#include "printinfo.h"
-#include "version.h"
+#include <vector>
 
 namespace NeovimQt {
 
+MainWindow* s_lastActiveWindow{ nullptr };
+std::vector<MainWindow*> s_windows;
+int s_exitStatus{ 0 };
```
jgehrig, line 19: Let's avoid globals whenever possible... Globals are tricky, and it is easy to accidentally introduce bugs.

You've also hit a few C++ gotchas here:

1. `s_` should only be used for `static` vars.
2. We should avoid globals with non-trival CTORs: `std::vector`.
3. Globals should be marked `static` to prevent this variable from being accessed outside of this compilation unit.

Can we define these variables somewhere else?

src/gui/app.cpp
```diff
@@ -1,18 +1,150 @@
...
+
+	QObject::connect(win, &MainWindow::closing, app, onWindowClosing);
+	QObject::connect(win, &MainWindow::destroyed, app, [win]() {
+		Q_ASSERT(std::find(s_windows.cbegin(), s_windows.cend(), win) != s_windows.cend());
+		s_windows.erase(std::find(s_windows.cbegin(), s_windows.cend(), win));
+		onWindowDestroyed();
+	});
+	QObject::connect(win, &MainWindow::activeChanged, app, onWindowActiveChanged);
+
+	s_lastActiveWindow = win;
```
jgehrig, line 143: I don't think you need this variable?

You should be able to access the first/last element of the vector directly:
 - https://en.cppreference.com/w/cpp/container/vector/front
 - https://en.cppreference.com/w/cpp/container/vector/back

src/gui/app.cpp
```diff
@@ -104,17 +246,14 @@ void App::showUi() noexcept
 		win->show();
 	}
 #else
-	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(m_connector.get());
+	// FIXME: On macOS, Cmd+q kills the first window that was created.
```
jgehrig, line 249: No FIXME or un-tracked TODO comments please :)

Two options here:
1. Fix the underlying issue now (preferred).
2. File an issue and reference the issue number: `// TODO Issue#??: On macOS, CMD+q ...`

Is this a bad enough issue that we should fix it now, before check-in?

jgehrig:
@Furkanzmc 

Looks great! Thanks for all of your work :+1:

What is the state of this PR? Do you think the feature is ready for merge and general use?

---

I have added a few minor comments that should be addressed before merge.

I also have a few nit-picky syntax comments: init with `{}`, diff minimization, etc. All very trivial.

Two options:
1. I quickly push some trivial changes to your branch before I merge (probably easiest).
2. I can add comments and you can fix.

I will try to do some quick testing on a MacOS VM once you think this is ready.

src/gui/app.cpp
Furkanzmc: I usually use the anonymous namespace for this. I think I got visually tricked by the namespace above, which I don't normally do, and just put the variables there. I can move them to the anonymous namespace if that's acceptable to you.

src/gui/app.cpp
Furkanzmc: This is not about accessing the first or last, but the last active window. I don't change the order of the windows inside the vector when a different window becomes active. So I have to keep track of the active window explicitly.

src/gui/app.cpp
Furkanzmc: My bad. I actually fixed that but forgot to remove that FIXME. I'll remove it.

Furkanzmc:
> What is the state of this PR? Do you think the feature is ready for merge and general use?

I think it's ready for general use. Worst case there will be some bugs but I doubt there will be any blocking ones. I've been using this branch on my day to day work, and it all seems fine.

> I also have a few nit-picky syntax comments: init with `{}`, diff minimization, etc. All very trivial.

If you don't think it's going to take a lot of time to point them out for me, I would prefer that so I know what's expected from a PR for the project. Otherwise, I'm fine with you making the changes as well.

> I will try to do some quick testing on a MacOS VM once you think this is ready.

The feature is ready, I think you can do your testing before fixing those comments.

src/gui/app.cpp
```diff
@@ -1,18 +1,150 @@
...
+		Default
+	};
+
+	ConnectorInitArgs(const QCommandLineParser& parser, QStringList nvimArgs)
+		: type{ getConnectorType(parser) }
+		, timeout{ parser.value("timeout").toInt() }
+		, server{ parser.value("server") }
+		, nvim{ parser.value("nvim") }
+		, positionalArgs{ parser.positionalArguments() }
+		, neovimArgs{ std::move(nvimArgs) }
```
jgehrig, line 37: Are the move semantics here intentional?

You're copying for the param, and then moving the copied param.

I'd recommend:
1. `const QStringList nvimArgs` and get rid of the `std::move`.
2. Use `QStringList&&` to move from the caller. (better if possible).

src/gui/app.cpp
```diff
@@ -1,18 +1,150 @@
...
+	}
+
+	ConnectorInitArgs(
+		Type _type, int _timeout, QString _server, QString _nvim, QStringList nvimArgs)
+		: type{ _type }
+		, timeout{ _timeout }
+		, server{ std::move(_server) }
+		, nvim{ std::move(_nvim) }
+		, positionalArgs{}
+		, neovimArgs{ std::move(nvimArgs) }
```
jgehrig, line 48: Same here. Please double-check your move semantics.

Generally you should use:
1. `T&& val` with `m_var{ std::move(val) }`
2. `const T& val` with `m_var{ val }`

The move here doesn't do much, since your parameter is copied by value.

Based on looking at the caller, I think you want `T&&` as the type instead.

src/gui/app.cpp
jgehrig: Doh! Sorry I missed this.

Agreed, that makes sense.

src/gui/app.cpp
jgehrig: No worries, C++ is tricky. It helps to have multiple pairs of eyes reading the code.

Here's my order of preference:
1. statics live within the function they are used
2. statics live within a class
3. statics live at the top of the compilation unit (separate namespace is a nice touch)

The other issues still apply inside a namespace:
1. These need to be `static` so the linker does not expose them outside of this file.
2. `std::vector` should be avoided as a global. The CTOR gets called from startup code, this can be (silently) expensive.

src/gui/app.cpp
jgehrig: Excellent! :+1:

jgehrig:
I'll add my nit-picky comments so you are aware for the future. I apologize in advance :)

Eventually, I would like to catch these issue automatically with clang-tidy or similar. If you don't get them all, I will do some quick clean-up before merge.

I will do some testing, and we will get this merged after the remaining few comments are addressed.

src/gui/app.cpp
```diff
@@ -1,18 +1,150 @@
...
+
+		if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
+			return ConnectorInitArgs::Type::Spawn;
+		}
+
+		return ConnectorInitArgs::Type::Default;
+	}
+};
+
+NeovimConnector* connectToRemoteNeovim(const ConnectorInitArgs& args)
```
jgehrig, line 77: nitpick:

All new functions and CTORs should be marked as `noexcept`.

Here and elsewhere.

src/gui/shell.cpp
```diff
@@ -858,21 +864,27 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
...
+		}
+		else if (guiEvName == "AdaptiveFont") {
 			handleGuiAdaptiveFont(args);
-		} else if (guiEvName == "AdaptiveStyle") {
+		}
+		else if (guiEvName == "AdaptiveStyle") {
 			handleGuiAdaptiveStyle(args);
-		} else if (guiEvName == "AdaptiveStyleList") {
+		}
+		else if (guiEvName == "AdaptiveStyleList") {
```
jgehrig, line 887: nitpick:

Please keep the diff to a minimum. These lines are changing, so they can be reverted.

src/gui/app.cpp
```diff
@@ -348,12 +445,29 @@ void App::showVersionInfo(QCommandLineParser& parser) noexcept
...
+			server = initMap["server"].toString();
+			type = ConnectorInitArgs::Type::Server;
+		}
+	}
 
-void App::exitWithStatus() {
-	exit(m_exitStatus);
+	auto connector = connectToRemoteNeovim(
+		ConnectorInitArgs{ type, 2000, std::move(server), std::move(nvim), getNeovimArgs() });
+	auto win = createWindow(connector);
```
jgehrig, line 468: niptick:

Use uniform init `{}` instead of `=` when defining variables.

The compiler will enforce more aggressive type-checking with `{}`. Helps catch bugs.

src/gui/app.cpp
```diff
@@ -348,12 +445,29 @@ void App::showVersionInfo(QCommandLineParser& parser) noexcept
...
+
+		if (initMap.find("server") != initMap.end()) {
+			server = initMap["server"].toString();
+			type = ConnectorInitArgs::Type::Server;
+		}
+	}
 
-void App::exitWithStatus() {
-	exit(m_exitStatus);
+	auto connector = connectToRemoteNeovim(
```
jgehrig, line 466: nitpick:

Avoid `auto` if the type is simple. `auto` is good for iterators, complex templates.

`auto` is bad for simple types, the type has to be inferred by the reader. Fine for compiler, hard for humans.

src/gui/app.cpp
```diff
@@ -1,18 +1,150 @@
...
-
-#include "arguments.h"
-#include "mainwindow.h"
-#include "printinfo.h"
-#include "version.h"
+#include <vector>
 
 namespace NeovimQt {
 
+MainWindow* s_lastActiveWindow{ nullptr };
```
jgehrig, line 17: nitpick:

Try to avoid `T*` if `nullptr` is not an expected value.

Ex) `&` for required parameters/values and `*` for optional parameters/values. 

This fix might be hard for `s_lastActiveWindow` as you've written the code... If so, NBD.

I think this can be easily applied to `s_windows`, `std::vector<MainWindow&>`

src/gui/app.cpp
Furkanzmc: > The other issues still apply inside a namespace:

I don't think those issues apply If I put them inside the unnamed namespace. Or are you saying this because you don't want me using the unnamed namespace?

> std::vector should be avoided as a global. The CTOR gets called from startup code, this can be (silently) expensive.

It`s simple container of pointers. This container is going to be used throughout the application's lifetime and it's useful as soon as it's launched. I don't think it would affect a lot, but I'm happy to change it if you can tell me your preferred alternative.

jgehrig: > It`s simple container of pointers.

Is it a big performance hit? No.

However, it is a faux-paux in C++. You can very easily introduce expensive code at startup with a global `std::vector`. Please don't do this, let's avoid future headaches.

> I don't think those issues apply If I put them inside the unnamed namespace.

As I understand, yes the `static` linkage issues still apply.

Here is an example that avoids all of the issues above:
```
class App
{
...
static std::vector<const MainWindow&> m_windowList;
...
void AppendNewWindow(...) noexcept
{

}
void RemoveNewWindow(...) noexcept;
MainWindow* GetLastActiveWindow(...) noexcept;
}
```

Another alternative:
```
static std::vector<MainWindow&> GetWindowVectorList(...) noexcept;
{
    static std::vector<MainWIndow&> s_windowList{};
    // Do stuff?
    return s_windowList;
}
```

The compiler add some implicit code in both cases to initialize the vector on first use. Both are also easier to modify in the future.

src/gui/app.cpp
Furkanzmc: I use this approach with simple types that I always will copy. There's no condition that changes the fact that I will not be using `_nvim`. This is pattern so that I don't have to have one ctor with the move only parameter and one that takes a const ref. Stephen Meyers has this in his book as well. 

Do you prefer that I get rid of this? Or create a new ctro that takes `T&&` and one that takes `const T&`?

src/gui/shell.cpp
Furkanzmc: All these come from the patch that the CI provided for clang-format. I did not change these files. The build was rejected because of the formatting check.

src/gui/app.cpp
jgehrig: Sorry for the confusion. The `_varName` is fine.

I'm talking about the `std::move`. It doesn't do anything as written.

This will move the values:
```
..., QString&& _server, ...
...
, server{ std::move(_server) }
...
```

This code is probably not what you intended, but it is simpler and functionally equivalent to what you wrote:
```
..., const QString& _server, ...
...
, server{ _server }
...
```

src/gui/app.cpp
Furkanzmc: > Here's my order of preference:
> 
>   - statics live within the function they are used
>   - statics live within a class
>   - statics live at the top of the compilation unit (separate namespace is a nice touch)

That's a good rule of thumb. I'll keep that in mind. Thanks.

> You've also hit a few C++ gotchas here:
>     - s_ should only be used for static vars.

When put inside the anonymous name space, those variables are the same as statics. So, I don't see a
problem with the naming scheme here.

>     - We should avoid globals with non-trival CTORs: std::vector.

Correct, but in this case the lifetime of these variables are actually very tightly connected to the
life time of the application. If these were static class memebers, then the initialization order
would only change slightly. But I don't feel strongly about it, not that big of a difference. I'll
use your suggested solution.

>     - Globals should be marked static to prevent this variable from being accessed outside of this compilation unit.

Again, once they are in the anoynmous name space, they are static and only available to the current
complication unit.



>> I don't think those issues apply If I put them inside the unnamed namespace.
>
> As I understand, yes the static linkage issues still apply.

If you are talking about the static linkage issues you laid out above, I don't think they do. The
only time that difference might matter is when I put that static variable in a method, but then the
variable is used as soon as the application launches, I don't think it really doesn't make a
difference. But again, I don't feel strongly about them and it doesn't really make a mountain of
difference this way or the other. I'll make the changes.

src/gui/shell.cpp
jgehrig: Ahhhh, clang-format is being weird... That is a new check.

Let me investigate.

src/gui/app.cpp
jgehrig: C++ is complicated, there are so many ways to solve the same problem.

Okay... I learned something new here about anonymous namespaces. Thanks! :smile:

Any solution works for me, as long as the linkage is static. However you think this code flows best.

jgehrig: Feel free to keep `std::vector`, although I would recommend `MainWindow&` and `const` if possible.

There isn't a case where we expect a `nullptr`, so use a reference. References are better than the implicit never-null pointers with `assert` everywhere.

src/gui/app.cpp
jgehrig: I have no issues with the `_name` pattern.

If I'm understanding your comment correctly... This does in fact `std::move` here. Right?

I prefer being explicit with about `T&&` to avoid accidental copy (you aren't actually using copy, only move here right?). If you do need the copy and move, feel free to keep as-is.

I've never seen this pattern before. I'll pull out my copy of Scott Meyers "Effective Modern C++" and read more.

Thanks for the clarification. Again... I learned something new today! :smile:

src/gui/shell.cpp
jgehrig: Ohhh, and feel free to ignore the `clang-format` and CodeCov tests. Consider them both soft failures.

Clang-format is new, and I am still ironing out the bugs. Open to any suggestions.

Personally, I don't like:
 - All of the header changes/re-ordering.
 - The modifications to un-changed lines here.

src/gui/app.cpp
Furkanzmc: > If I'm understanding your comment correctly... This does in fact std::move here. Right?

Yes, it indeed does a move.

> I prefer being explicit with about `T&&` to avoid accidental copy (you aren't actually using copy, only move here right?). If you do need the copy and move, feel free to keep as-is.

I prefer this as well, but in this case there's really not that big of a gain from having two different ctors one of which takes `T&&` and the other `const T&`. So, I'll keep it as it is. 

> Thanks for the clarification. Again... I learned something new today! 😄

Well, I'm glad the tables turned and I was able to be useful to you. :D

src/gui/app.cpp
```diff
@@ -11,8 +11,148 @@
...
+	}
+
+	const Type type;
+	const int timeout;
+	const QString server;
+	const QString nvim;
+	const QStringList positionalArgs;
+	const QStringList neovimArgs;
+
+	Type getConnectorType(const QCommandLineParser& parser) noexcept
```
jgehrig, line 62: This should be `static`, if `this` is not used.

src/gui/app.cpp
```diff
@@ -11,8 +11,148 @@
...
+
+		if (parser.isSet("spawn") && !parser.positionalArguments().isEmpty()) {
+			return ConnectorInitArgs::Type::Spawn;
+		}
+
+		return ConnectorInitArgs::Type::Default;
+	}
+};
+
+NeovimConnector* connectToRemoteNeovim(const ConnectorInitArgs& args) noexcept
```
jgehrig, line 80: This function should be `static`, it is not used outside of this file (I think).

This applies to most new functions in this PR. Generally, functions not defined in a header should be `static`.

src/gui/app.cpp
```diff
@@ -11,8 +11,148 @@
...
+{
+	if (s_windows.empty()) {
+		qApp->exit(s_exitStatus);
+	}
+}
+
+MainWindow* createWindow(NeovimConnector* connector) noexcept
+{
+	Q_ASSERT(connector);
+	NeovimQt::MainWindow *win = new NeovimQt::MainWindow(connector);
```
jgehrig, line 132: uniform init `{}`

src/gui/app.cpp
```diff
@@ -11,8 +11,148 @@
...
+	win->setAttribute(Qt::WA_DeleteOnClose);
+
+	App *app = qobject_cast<App *>(App::instance());
```
jgehrig, line 135: uniform init `{}`

src/gui/app.cpp
```diff
@@ -11,8 +11,148 @@
...
+	Q_ASSERT(app);
+
+	QObject::connect(win, &MainWindow::closing, app, onWindowClosing);
+    QObject::connect(win, &MainWindow::destroyed, app, [win]() {
+        auto foundIt = std::find_if(
+            s_windows.cbegin(), s_windows.cend(),
+            [win](const std::reference_wrapper<NeovimQt::MainWindow>& rw) {
```
jgehrig, line 142: Curious... Why `reference_wrapper` here? Shouldn't a plain `const MainWindow&` work?

Also no need for `NeovimQt`.

src/gui/app.cpp
Furkanzmc: You are right. Same reason as to how I forgot to move the variables to the unnamed namespace. Apparently I forgot these too. Sorry. I'll fix it.

src/gui/app.cpp
Furkanzmc: We can't have vector of references since references are not assignable. That's why the use of `std::reference_wrapper`.

src/gui/shell.cpp
Furkanzmc: Dropped the commit with the formatting changes.

tucnak:
Wow this is promising, we shall hope this goes thru and lands ASAP. Go go Fyrkanzmc!

src/gui/runtime/plugin/nvim_gui_shim.vim
```diff
@@ -21,6 +21,14 @@ function! GuiClose() abort
   call rpcnotify(0, 'Gui', 'Close', v:exiting)
 endfunction
 
+" Open new window
+function! GuiNewWindow(...) abort
+	let server = get(a:, 0, 0)
+	let nvim = get(a:, 1, 0)
+
+	call rpcnotify(0, 'Gui', 'NewWindow', server, nvim)
+endfunction
```
jgehrig, line 30: This is a function change. I've made the parameters optional.

Is that what you intended?

Before this, I was unable to call `:call GuiNewWindow()`.

jgehrig:
@Furkanzmc 

I've made some changes to get this ready for merge. Please let me know if you have any concerns.

Most of the changes were trivial syntax changes (sorry :sweat_smile:).

One small functional change was made. See comment. 

This looks great! I haven't found any bugs during my testing...

**Tasks TODO:**
1. Write quick documentation with usage examples. README or Wiki.
2. Add basic tests to `tst_shell`, or perhaps write a new `tst_window`

If you'd like, I am hapy to merge this now. We can do a follow-up PR for the above tasks.

src/gui/runtime/plugin/nvim_gui_shim.vim
Furkanzmc: Yes, that was my intention. I just did that to get it working but then forgot to do it the right way. Thanks for fixing it.

Furkanzmc:
> @Furkanzmc
> 
> I've made some changes to get this ready for merge. Please let me know if you have any concerns.
> 
> Most of the changes were trivial syntax changes (sorry 😅).
> 
> One small functional change was made. See comment.

All looks good. Thanks.

> 
> This looks great! I haven't found any bugs during my testing...
> 
> **Tasks TODO:**
> 
> 1. Write quick documentation with usage examples. README or Wiki.
> 2. Add basic tests to `tst_shell`, or perhaps write a new `tst_window`
> 
> If you'd like, I am hapy to merge this now. We can do a follow-up PR for the above tasks.

I think that makes sense. I can work on the tests in a separate PR.

jgehrig:
@Furkanzmc 

Pull Request Merged! :tada:

Thank you for your contributions and great work! I look forward to more in the future :)

I'd also like to thank you for working through and addressing all of my comments. Not easy, but the resulting code is usually better after such review.

Also, I learned a few new tricks (anonymous namespaces, copy-move-forwarding, etc) from our discussions. Most appreciated! :+1: 

**Outstanding Tasks:**
#899 - Documentation
#900 - Clazy Warning (Global Static)
#901 - Thread Safety
#902 - Test Coverage

Please let me know if I can help with any of these tasks...

Furkanzmc:
Hi @jgehrig . Sorry for being quiet for so long. This month has been very hectic for me.

Thanks for merging. I'm looking forward to making more contributions.

> I'd also like to thank you for working through and addressing all of my comments. Not easy, but the resulting code is usually better after such review.
> 
> Also, I learned a few new tricks (anonymous namespaces, copy-move-forwarding, etc) from our discussions. Most appreciated! 👍

It goes both ways. Thank you. I'm sure the next PRs will be easy now that I've gotten a bit more used to the coding style in the project.

> **Outstanding Tasks:**
> #899 - Documentation
> #900 - Clazy Warning (Global Static)
> #901 - Thread Safety
> #902 - Test Coverage
> 
> Please let me know if I can help with any of these tasks...

Is there a way you can assign those issues to me so I can filter for them? It would also help others in case they decide to fix any of them.

jgehrig:
Not a problem! My activity here also comes in waves.

> Is there a way you can assign those issues to me so I can filter for them?

Unfortunately no :/

The GitHub Issue Tracker leaves a lot to be desired. I can only assign to myself of equalsraf (users with push access).

`https://github.com/isaacs/github/issues/100`

I'm surprised this feature isn't available to contributors at least... It seems like a huge gap in functionality.

Issue #851, opened 2021-05-14
Highlight does not respect guisp for underline


YorickPeterse:
Support for the undercurl setting was added back in https://github.com/equalsraf/neovim-qt/commit/1d9c076e2992705c860f094433b704bae9ef177e. At the time the value for `guisp` was hard coded. It seems this may have been solved in https://github.com/neovim/neovim/pull/4633. I believe at least goneovim and neovide support this setting, though I could be mistaken.

If possible, it would be nice for neovim-qt to support this setting.

jgehrig:
How is this supposed to behave? Do you have a non-working example?

From what I can tell, we are correctly processing the colors as sent by `nvim`.

For example:
`:hi FooBar guisp=Blue`
`QMap(("special", QVariant(qulonglong, 255)))`

We received and register the correct `guisp` value.

Interestingly, I'm not able to generate a `:hi` with property `undercurl`:
`:hi FooBar cterm=undercurl guisp=Blue`
`QMap(("special", QVariant(qulonglong, 255)))`

YorickPeterse:
@jgehrig The `guisp` parameter specifies the color for just the underline/undercurl that is drawn. For example, this is using the TUI:

![Screenshot from 2021-05-14 15-54-20](https://user-images.githubusercontent.com/86065/118280646-b3ee5080-b4cc-11eb-82d3-7d85a840b296.png)

Here the highlight group is `SpellBad       xxx ctermbg=224 gui=bold,underline guifg=#cc3e28 guisp=blue`.

Here is what that currently looks like in neovim-qt:

![Screenshot from 2021-05-14 15-56-05](https://user-images.githubusercontent.com/86065/118280889-eac46680-b4cc-11eb-95b2-2ef9fd7bcf8a.png)

This is using the same highlight group, except neovim-qt seems to ignore the `guisp` setting, always rendering the underline using the text color.

jgehrig:
@YorickPeterse

Okay, I see the issue...

The value for `undercurl` is correct:
![image](https://user-images.githubusercontent.com/11207308/118861217-41a1b580-b8aa-11eb-9d9c-bd0628a0efc1.png)
`:hi SpellBad ctermbg=224 gui=bold,undercurl guifg=#cc3e28 guisp=blue`

The value for `underline` is incorrect:
![image](https://user-images.githubusercontent.com/11207308/118861440-7f9ed980-b8aa-11eb-92cb-14a205aa8a69.png)
[...]

PR #856, opened 2021-05-19
Issue 851: Underline should render with special

jgehrig:
**Issue #851:** Undercurl is rendered with the wrong color.

Currently, we render `:Highlight` with color `foreground`.

According to the Vim spec, underline should be rendered with color `special`. We will share the `QPen` color lookup logic with `paintUndercurl`.

image
`:hi SpellBad ctermbg=224 gui=bold,underline guifg=#cc3e28 guisp=green`

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -229,7 +229,7 @@ void ShellWidget::paintUnderline(
 		return;
 	}
 
-	p.setPen(getForegroundPen(cell));
+	p.setPen(getSpecialPen(cell));
 
 	p.drawLine(GetUnderline(cellRect));
 }
@@ -274,18 +274,7 @@ void ShellWidget::paintUndercurl(
 		return;
 	}
 
-	QPen pen;
-	if (cell.GetSpecialColor().isValid()) {
-		pen.setColor(cell.GetSpecialColor());
-	} else if (special().isValid()) {
-		pen.setColor(special());
-	} else if (cell.GetForegroundColor().isValid()) {
-		pen.setColor(cell.GetForegroundColor());
-	} else {
-		pen.setColor(foreground());
-	}
-
-	p.setPen(pen);
+	p.setPen(getSpecialPen(cell));
 
 	p.drawPath(GetUndercurlPath(cellRect));
 }
@@ -373,6 +362,22 @@ QPen ShellWidget::getForegroundPen(const Cell& cell) noexcept
 {
 	QPen pen;
 	if (cell.GetForegroundColor().isValid()) {
+		pen.setColor(cell.GetForegroundColor());
+	} else {
+		pen.setColor(foreground());
+	}
+
+	return pen;
+}
+
+QPen ShellWidget::getSpecialPen(const Cell& cell) noexcept
+{
+	QPen pen;
+	if (cell.GetSpecialColor().isValid()) {
+		pen.setColor(cell.GetSpecialColor());
+	} else if (special().isValid()) {
+		pen.setColor(special());
+	} else if (cell.GetForegroundColor().isValid()) {
 		pen.setColor(cell.GetForegroundColor());
 	} else {
 		pen.setColor(foreground());
--- a/src/gui/shellwidget/shellwidget.h
+++ b/src/gui/shellwidget/shellwidget.h
@@ -181,6 +181,7 @@ private:
 
 	QFont GetCellFont(const Cell& cell) const noexcept;
 	QPen getForegroundPen(const Cell& cell) noexcept;
+	QPen getSpecialPen(const Cell& cell) noexcept;
 
 	ShellContents m_contents{ 0, 0 };
 	QSize m_cellSize;


Shatur95:
@jgehrig, but the color should fallback to `foreground` if `guisp` is missing. Now it renders `red` color:
изображение

jgehrig:
@Shatur95 

Do you have an `:hi ...` example that reproduces the issue?

I would expect the  code to use foreground as a fallback already. If not, this sounds like a bug.

No new code here, I re-used the existing undercurl color-selection logic:
```C++
QPen pen;
if (cell.GetSpecialColor().isValid()) {
	pen.setColor(cell.GetSpecialColor());
} else if (special().isValid()) {
	pen.setColor(special());
} else if (cell.GetForegroundColor().isValid()) {
	pen.setColor(cell.GetForegroundColor());
} else {
	pen.setColor(foreground());
}
```

Shatur95:
> No new code here, I re-used the existing undercurl color-selection logic:

@jgehrig, You reused the code from `undercurl` that results to red color by default.
Any highlighting group without `guisp` and `style=underline` will trigger the issue.

Shatur95:
> I agree, this behavior is rather odd. My expectation without reading the man page would be always render with the foreground color; as we did originally.

Both of these behaviors are wrong. Yes, you should use `guisp` for undercurl and underline. But if `guisp` is not specified, then the color should fallback the background color.

YorickPeterse:
@jgehrig I believe that @Shatur95 is correct here: when `guisp` is not set, the fallback color should be the foreground color. This matches what the NeoVim TUI does. 

To reproduce, you can do something like the following:

1. `set ft=ruby`
2. type `"foo"`
3. run `hi String gui=underline`

Assuming that in your setup `rubyString` is linked to `String` (which IIRC is the case by default), you should now have the following:

- The underline is the same color as the text in the TUI (= correct)
- The underline is red in nvim-qt (= incorrect)

I didn't catch this earlier because for the few highlight elements that I use underlines/undercurls for, I also have `guisp` set. Oops!

jgehrig:
Thanks @YorickPeterse  and @Shatur95!

ACK. Understood, the proposed logic sounds reasonable:
``` C++
QPen pen;
if (cell.GetSpecialColor().isValid()) {
	pen.setColor(cell.GetSpecialColor());
} else if (cell.GetForegroundColor().isValid()) {
	pen.setColor(cell.GetForegroundColor());
} else {
	pen.setColor(foreground());
}
```

**However**...

I think we should first understand what `gVim` does in this scenario. Generally, I think we should match either `gVim` or `TUI` behavior.

I tried `:help guisp` in gVim 8.2 (a somewhat old version?):
```
guifg={color-name}					*highlight-guifg*
guibg={color-name}					*highlight-guibg*
guisp={color-name}					*highlight-guisp*
	These give the foreground (guifg), background (guibg) and special
	(guisp) color to use in the GUI.  "guisp" is used for undercurl and
	strikethrough.
```

Interestingly, they don't use `special` at all... I'm going to get the latest version and re-test.

> To reproduce, you can do something like the following:
>    set ft=ruby
    type "foo"
    run hi String gui=underline

I don't see the same behavior...

In this scenario, my `TUI` only responds to `cterm`, `ctermfg`, and `ctermbg`. I probably don't have some color setting enabled in my terminal, `Konsole`.

What terminal emulator are you using?

Shatur95:
> In this scenario, my TUI only responds to cterm, ctermfg, and ctermbg. I probably don't have some color setting enabled in my terminal, Konsole.

I also use Konsole, but I have `termguicolors` enabled.

YorickPeterse:
@jgehrig I'm using the GNOME terminal, with `set termguicolors`, and `TERM` set to `xterm-256color` (if it's of any use).

jgehrig:
This behavior is Neovim specific, and was introduced to the TUI in https://github.com/neovim/neovim/commit/c4006a621eff1761208b0aa38057f4668b4a6ba8

Vim does not behave this way. It uses our previous logic, which renders `underline` in `guifg`.

It looks like another Vim/Neovim deviation exists in `strikethrough`:
https://github.com/vim/vim/blob/130cbfc31235c6cb52ffe718ea0a5bb50fbbc9fd/runtime/doc/syntax.txt#L5079-L5081
```
These give the foreground (guifg), background (guibg) and special
(guisp) color to use in the GUI.  "guisp" is used for undercurl and
strikethrough.
```

My terminal is Konsole (Yakuake) with `:set termguicolors` and `TERM` of `xterm-256color`

Interesting. I am unable to to test this scenario:
```
:version
NVIM v0.5.0-dev+1119-gf2fc44d50-dirty
Build type: Gentoo
```
image
image
image
image


It looks like `undercurl` is not supported by TUI on my system?

jgehrig:
**Here's what we will do**:

It sounds like everyone is comfortable with the newly proposed fix/logic.

@jgehrig

I will create a PR with this logic:
```
QPen pen;
if (cell.GetSpecialColor().isValid()) {
	pen.setColor(cell.GetSpecialColor());
} else if (cell.GetForegroundColor().isValid()) {
	pen.setColor(cell.GetForegroundColor());
} else {
	pen.setColor(foreground());
}
```

I will look into merging the Vim/Neovim documentation for `strikethrough` consistency.

I will test the behavior for `strikethrough` with gVim, nvim-qt.

My TUI output is above. Something is misconfigured or not-supported; I will need testing help.

@Shatur95 @YorickPeterse 

Please test upcoming PR with your configuration.

It would be helpful if you can describe the TUI behavior you see for `strikethrough` too.

YorickPeterse:
@jgehrig When using the TUI, `strikethrough` defaults to the foreground color. The same applies to `undercurl`:

undercurl
underline

Shatur95:
> It looks like undercurl is not supported by TUI on my system?

Same here.

> Interesting. I am unable to to test this scenario:

But according to your screenshot you have the described behavior - undercurl defaults to the foreground color. This is what we are talking about.

jgehrig:
@YorickPeterse 

Thanks for the pictures, that matches my understanding of what you've both described.

I'm also assuming running `:hi SpellBad guifg={X} guisp={Y} gui=underline`  sets text color `{X}` and underline color `{Y}`.

@Shatur95 

> But according to your screenshot you have the described behavior - undercurl defaults to the foreground color. This is what we are talking about.

No, I don't think so?

The screenshots are from Konsole Neovim TUI.

Notice `guisp=Red` is set in my screenshot. My TUI uses `guifg` regardless of the `guisp` value.

---

Anyways...

The logic we've discussed above sounds reasonable:
1. Cell Special
2. Cell Foreground
3. Default Foreground

My "broken" change here uses:
1. Cell Special
2. Default Special
3. Cell Foreground
4. Default Foreground

Let's do some behavior comparisons and testing over at #860.

Thanks for the help you two! :smile:

Shatur95:
> Notice guisp=Red is set in my screenshot. My TUI uses guifg regardless of the guisp value.

Oh, missed it, sorry, you are right. Yes, I have the same behavior my TUI always uses `guifg` too.

> The logic we've discussed above sounds reasonable:

Yep, thank you!

Shatur:
> The logic we've discussed above sounds reasonable:
> Cell Special
> Cell Foreground
> Default Foreground

I think it skips `Cell Foreground`. Here is my screenshot from konsole (left) and Neovim Qt (right):

изображение

Shatur:
@equalsraf @jgehrig sorry for bothering, but the following logic:

> The logic we've discussed above sounds reasonable:
> Cell Special
> Cell Foreground
> Default Foreground
 
Doesn't work in the latest release. I believe that there was a regression. In works like this:

> My "broken" change here uses:
>  Cell Special
>  Default Special
>  Cell Foreground
>  Default Foreground

PR #857, opened 2021-05-21
QStringView Compiler Error Ubuntu Bionic (18.04)

jgehrig:
Error:
```
./src/neovimconnector.cpp:358:14: error: ‘QStringView’ was not declared in this scope
   int port = QStringView{addr}.mid(colon_pos+1).toString().toInt(&ok);
```

We should `#include <QStringView>` before we use it.

### DIFF
--- a/src/neovimconnector.cpp
+++ b/src/neovimconnector.cpp
@@ -1,11 +1,14 @@
 #include "neovimconnector.h"
-#include <QtGlobal>
-#include <QMetaMethod>
+
 #include <QLocalSocket>
+#include <QMetaMethod>
+#include <QStringView>
 #include <QTcpSocket>
+#include <QtGlobal>
+
+#include "msgpackiodevice.h"
 #include "msgpackrequest.h"
 #include "neovimconnectorhelper.h"
-#include "msgpackiodevice.h"
 
 namespace NeovimQt {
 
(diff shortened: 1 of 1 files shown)


jgehrig:
@equalsraf 

Hmmm... It looks like `QStringView` was added in Qt 5.10. This isn't compatible with our minimum version, Qt 5.8.

Your change 00f6ba877e9f3e451ce3b199d1450408ed54a552 is invalid on Qt 5.9 and Qt 5.8.

We should also add some additional CI coverage... I thought we had adequate old-Qt coverage, but it appears this is not the case.

jgehrig:
Similar issues related to the Qt6 port in other areas:
```
./src/gui/shellwidget/shellwidget.cpp:520:3: error: ‘qsizetype’ was not declared in this scope
```

equalsraf:
For the sake of sanity - and since i don't have a solution for the other qt6 issues yet - it is fine to roll back the changes I merged earlier.

It looks like qsizetype also comes from 5.10.

equalsraf:
I opened https://github.com/equalsraf/neovim-qt/pull/858 for a CI using bionic.

jgehrig:
Reverting this PR in favor of #858.

> It looks like qsizetype also comes from 5.10.

Yes.

For `qsizetype`, we can probably use `auto`. While I find `auto` makes code harder to understand for simple types, this seems like a good application. The Qt API is not consistent across versions, and we want to use whatver `QVector<T>::size()` returns.

For `QStringView` we'll need to come up with something more clever. A revert is fine with me for now.

PR #858, opened 2021-05-21
Add azure CI build for ubuntu 18.04

equalsraf:
Recently I introduced a change that breaks builds in older versions
of Qt (<5.10) - this was not spotted in CI builds since none of the
builders uses a version under 5.10. This enables a Ubuntu 18.04 build
that should use an older Qt version.

### DIFF
--- a/contrib/azure-pipelines.yml
+++ b/contrib/azure-pipelines.yml
@@ -40,6 +40,25 @@ jobs:
         cmakeArgs: --build .
       displayName: CMake Build
 
+  - job: UbuntuBionic
+    pool:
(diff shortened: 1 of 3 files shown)


jgehrig:
I think this will gracefully fix the `qsizetype` issues:

`auto sizeGlyphRun{ glyphPositionList.size() };`

I can't comment outside of the PR diff. Weird...

equalsraf:
Yup i saw you comment in the other PR but I had already pushed the second revert :D. I'm pushing a commit with `auto` right now.

equalsraf:
Looks good on the CI side, pressing the merge button.

PR #866, opened 2021-06-03
Set initial editor size properly

### LINKED ISSUE #799: Initial editor size is different from the actual window size
This is what I get by `nvim-qt -- -O a b`:

image

The expected behavior is to open two panes of equal size side by side.

The reason for this unexpected behavior is the initial editor size is hard-coded relative to screen size, not window size:

https://github.com/equalsraf/neovim-qt/blob/d3f4b74021897157f3efd66a2d47791ce70c8250/src/gui/shell.cpp#L225-L227

To fix this, we need to know the actual window size before the Neovim's initialization. But current implementation does opposite: main window is waiting for Neovim's initialization, to fix other issue #70 (in commit 22aebb8e8eaf7b5799360416ff297da09c388469).

I tried removing this delaying behavior completely in my fork,  https://github.com/statiolake/neovim-qt/commit/a0ab4fda0f0be3dd87fbb6c07182a983eca3eadc and *for me* it works very well (including issue in #70 - actually removing delay does change things a bit, I can now see the black window just a moment until the Neovim get initialized, but it's not frustrating for me.)

statiolake:
When you run `nvim-qt -- -O a b` in your console, you'll see inequally divided two editor pains in the opened Neovim-Qt. This is because the initial editor size (lines and columns) is not based on the actual window size.

To resolve this issue, we need to know the window size and cell size before initializing Neovim. However it's impossible with the current implementation, because the window is shown only after Neovim initialization (delayed show mechanism first introduced in #70). I tried to see if I could get Qt to calculate the layout before the window is shown, but I couldn't find a way.

Therefore this commit tweaks three things:

- Show window (and shell widget) first and attach to neovim after the main window is shown.
- Use actual window size to calculate initial attaching size instead of the hard-coded value (the current code is assuming 60% of display size as a initial height and width, regardless of the actual window size).
- Remove delayed show mechanism entirely.

In addition to those, prevent init() from being called multiple times. This is because init() is now called from showEvent() but showEvent() is emitted not only when the window is newly created but also minimized window is restored.

Note that this PR entirely removes delaying mechanism. Even though I don't think the original issue #70 will appear again because Neovim-qt now restores the previous geometry on startup, I think we should check this fix doesn't make startup issues (not only #70 but also others like #198) before merging (but how?). Personally I *sometimes* do notice a flicker on startup but it's acceptable (considering that setting up colorscheme can also cause the similar flicker).

Fix #799.

### DIFF
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -129,11 +129,11 @@ void App::showUi() noexcept
 	}
 
 	if (m_parser.isSet("fullscreen")) {
-		win->delayedShow(NeovimQt::MainWindow::DelayedShow::FullScreen);
+		win->showFullScreen();
 	} else if (m_parser.isSet("maximized")) {
-		win->delayedShow(NeovimQt::MainWindow::DelayedShow::Maximized);
+		win->showMaximized();
 	} else {
-		win->delayedShow();
+		win->show();
 	}
 #endif
 }
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -160,8 +160,6 @@ bool MainWindow::neovimAttached() const
 /** The Neovim process has exited */
 void MainWindow::neovimExited(int status)
 {
-	showIfDelayed();
-
 	if (m_nvim->errorCause() != NeovimConnector::NoError) {
 		m_errorWidget->setText(m_nvim->errorString());
 		m_errorWidget->showReconnect(m_nvim->canReconnect());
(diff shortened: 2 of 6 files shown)


statiolake:
@jgehrig I think it's ready except Codecov report. However the decrease in the coverage seems to be simply due to a decrease in the total lines of code.

jgehrig:
@statiolake 

From a quick read, this looks good. I will review in more detail later this evening.

I will probably apply this patch to my machine for a few days, just to make sure everything works well before merge.

> I think it's ready except Codecov report.

No worries. Generally, we can ignore the CodeCov report. It would be very hard to unit-test this change.

PR #869, opened 2021-06-06
Issue #868: Always update `guifont` when needed

### LINKED ISSUE #868: `guifont` not updated with `:GuiFont` if the font is already saved in QSettings
## Summary

If the font specified by `:GuiFont` command in init.vim is already saved in QSettings, `:set guifont?` shows empty string unless the different font is manually set.

Steps to reproduce:

1. Create the following file and save it as `repro.vim` in the current dir:
  ```vim
  " Change font name if needed
  autocmd VimEnter * Guifont! Consolas:h15
  ```
2. Run `nvim-qt -- -u repro.vim` once.
  At this stage, `set guifont?` correctly returns the font name you specified unless the above font is what you usually use.
3. Close the previous window, then run `nvim-qt -- -u repro.vim` again.
4. `set guifont?` gives an empty string (`set guifont=`)

## Details (cause)

With cafc9ca9f03e7f4104cc81c0330ae220e5f23497, the font is now saved to use as a default font in the next launch. The font is restored here in the initialization:

https://github.com/equalsraf/neovim-qt/blob/e1fc9d3684cbbdf29716cce212d52022e326b014/src/gui/shell.cpp#L80-L85

Inside `setGuiFont()`, we are updating `guifont` option, but it doesn't seem to be actually updated. I doubt it is because during initialization the Neovim instance is actually not ready yet. `guifont` is still not updated when init.vim is loaded, as we are skipping changing the font if `fontDesc()` is already the same with the new font:

https://github.com/equalsraf/neovim-qt/blob/e1fc9d3684cbbdf29716cce212d52022e326b014/src/gui/shell.cpp#L124-L126

I think we need to treat initial font differently. For example, we can introduce a new flag variable that indicates the current font is automatically restored from QSettings.

~~By the way, is it OK to call `NeovimConnector::api0()` before the Neovim is ready? I saw the warning at `api0()`~~ I'm sorry, but it was checked and only executed when Neovim was attached. No problem.

statiolake:
Try to update option `guifont` even if the fontDesc is unchanged.

Close #868.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -81,7 +81,10 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 	if (guiFont.canConvert<QString>())
 	{
 		QString fontDescription{ guiFont.toString() };
-		setGuiFont(fontDescription, true /*force*/, true /*updateOption*/);
+		// We shouldn't update `guifont` because this is just a default font,
+		// not something specified explicitly by the user. In addition to
+		// that, we can't set the option here because Neovim is not ready yet.
+		setGuiFont(fontDescription, true /*force*/, false /*updateOption*/);
 	}
 
 	if (!m_nvim) {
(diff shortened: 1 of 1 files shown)


jgehrig:
@statiolake Generally looks good. :+1: 

This is an error that I have been aware of for a while.

It wasn't exactly introduced in https://github.com/equalsraf/neovim-qt/commit/cafc9ca9f03e7f4104cc81c0330ae220e5f23497. I modified`GuiFont` to update `guifont` in https://github.com/equalsraf/neovim-qt/commit/06b40dda38518a827f0e870d7768fcf36c453a4a. However, I did exposed the existing bug in a new way. I mentioned only so we extra historical search context in the future :)

Here is a simpler repro case:
1. Use neovim-qt `0.2.16.1`
2. Set `GuiFont! Some\ Font:h10` in `ginit.vim`
3. `nvim-qt -- -u NONE`
4. Call `:set guifont`
5. Call `:GuiFont`

Notice `GuiFont` reports the correct font, and `guifont` does not. Bug.

It looks like your code will fix both scenarios.

statiolake:
Thank you for checking!

I maybe confirmed the issue you pointed out with Neovim-qt 0.2.16.1, though the issue occurs only when the specified font is happens to be the same with the default font of the platform. (Such as `Consolas:h11` on Windows, as defined in shellwidget.cpp).

https://github.com/equalsraf/neovim-qt/blob/9f407f3efeb66984ffc824176879099dde07c3b5/src/gui/shellwidget/shellwidget.cpp#L34-L49

If the specified font differs in the name or the size, such as `Courier\ New:h11` or `Consolas:h12`, then `guifont` is successfully updated. If that's also true on your platform, I understand this patch can fix that too, as #868 and yours have the same root cause.

jgehrig:
I'm not too concerned with exact root cause. I only mentioned the above for search purposes in the future.

This issue is related to both cafc9ca and 06b40dd. 

> the issue occurs only when the specified font is happens to be the same with the default font of the platform.

I haven't done exhaustive testing... But this is not a requirement on my machine. With Linux and `v0.2.16.1`, `GuiFont` in `ginit.vim` does not update `guifont` both for the default font and non-default font.

No matter though, your patch fixes all called-out issues on my machine!

I'm comfortable with a merge once we tweak the code to minimize the changes.

jgehrig:
Closing this in favor of #875.

We can simplify the logic in `setGuiFont` if we use the Neovim API to read both variables and only update either as necessary (if changed). Loops will be prevented with an exit-early in `setGuiFont`.

Issue #870, opened 2021-06-25
undercurl doesn't render smoothly


YorickPeterse:
This is a bit nit picky, but here's how undercurls are rendered in GNOME terminal:

![terminal](https://user-images.githubusercontent.com/86065/123429856-bf529280-d5c7-11eb-9410-e8181ee4254d.png)

Note how the undercurl is rendered smoothly. nvim-qt on the other hand renders it in a more messy way:

![nvim-qt](https://user-images.githubusercontent.com/86065/123429909-cd081800-d5c7-11eb-9cb2-c1e6e952084e.png)

It looks like each character is given it's own undercurl, with the spacing in between caused by the spacing between characters; instead of there being a single smooth line below all the characters.

nvim version: `NVIM v0.5.0-dev+1405-g2f0e5e7e6`
nvim-qt version: `NVIM-QT v0.2.17.9999`
OS: Arch Linux

jgehrig:
Thanks for the bug report!

I thought these issues were resolved with #803...

A few questions:
1. What is the output of `nvim-qt --version`?
2. Is this the latest release from `master`?
3. Does the rendering vary with `:GuiRenderLigatures?
4. Do you have display scaling enabled?

Here is what I see:
![image](https://user-images.githubusercontent.com/11207308/123445663-1d6e7e00-d5a6-11eb-992e-abc2dfb9516c.png)

Related: #802

YorickPeterse:
@jgehrig 

> What is the output of nvim-qt --version?

```
NVIM-QT v0.2.17.9999
Build type: Release
Compilation:-D_FORTIFY_SOURCE=2 -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wall -Wextra -Wno-unused-parameter -Wunused-variable -std=c++11
Qt Version: 5.15.2
Environment: 
  nvim: nvim
  args: --cmd set termguicolors
  runtime: 

NVIM v0.5.0-dev+1405-g2f0e5e7e6
Build type: RelWithDebInfo
LuaJIT 2.0.5
[...]

PR #871, opened 2021-06-30
Scale undercurl and underline with font size.

gebner:
Currently, the undercurl and underline are drawn with a one-pixel line width, irrespective of the font size.  This results in extremely small lines on hidpi displays (small compared to the line width of the font):

undercurl_before

This PR scales the undercurl and underline size with the font size:

undercurl_after

See also #870 and #803.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -5,6 +5,7 @@
 #include <QPainterPath>
 #include <QPaintEvent>
 #include <QTextLayout>
+#include <QtMath>
 
 #include "helpers.h"
 
@@ -209,13 +210,13 @@ void ShellWidget::paintNeovimCursorForeground(
 	p.setClipping(oldClippingSetting);
 }
 
-static QLine GetUnderline(QRect cellRect) noexcept
+static QLine GetUnderline(QRect cellRect, qreal penWidth) noexcept
 {
 	QPoint start{ cellRect.bottomLeft() };
-	start.ry()--;
+	start.ry() -= penWidth - 1;
 
 	QPoint end{ cellRect.bottomRight() };
-	end.ry()--;
+	end.ry() -= penWidth - 1;
 
 	return { start, end };
 }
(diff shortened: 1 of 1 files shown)


jgehrig:
Looks much better! :+1: 

Have you tested with `:GuiRenderLigatures 1` and `:GuiRenderLigatures 0`?

I'll do some testing myself, and try to get this merged shortly.

gebner:
> Have you tested with `:GuiRenderLigatures 1` and `:GuiRenderLigatures 0`?

I have tested with both of the settings, but I didn't see a difference with any of the fonts I have.  (And for some reason neovim-qt doesn't want to load Fira Code here, claiming that it is "not a fixed pitch font".)

jgehrig:
@gebner 

Yes, the `Fira Code` issue you noticed is a Linux-only bug in Qt:
https://bugreports.qt.io/browse/QTBUG-83699?gerritIssueType=IssueOnly

We should add a friendly `+1` to the Qt bug. They should really fix this issue, FiraCode is also not present in the monospace-only Font Dialog.

This PR still needs some work. Here are some rendering artifacts I noticed with your build:
image
image

Notice the Anti-Aliasing ghosting and undercurl discontinuity between some cells.

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -570,6 +566,8 @@ void ShellWidget::paintEvent(QPaintEvent *ev)
 {
 	QPainter p(this);
 
+	p.setRenderHint(QPainter::Antialiasing);
```
jgehrig, line 567: I don't think this is valid as-written. I think this is the cause of ghosting in my screenshot?

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -1,12 +1,13 @@
 #include "shellwidget.h"
 
+#include "helpers.h"
```
jgehrig: `clang-format` is new, and I don't like what it is doing to the headers here.

I'm going to change it. Feel free to ignore the header-related format failures.

gebner:
> Here are some rendering artifacts I noticed with your build:
> Notice the Anti-Aliasing ghosting and undercurl discontinuity between some cells.

Unfortunately I can't reproduce either of those artifacts.  What platform/qt version are you on?

I've disabled the anti-aliasing setting again, so now there should only be a difference for cells with underline, undercurl, or strikethrough.

gebner:
> undercurl discontinuity between some cells.

After some playing around, I believe this happens at small font sizes.  (Without anti-aliasing you get different artifacts.)  We might need to fall back to the old code at those sizes.

jgehrig:
I'm on Gentoo + Plasma/KDE. I can check versions for Plasma/Qt and report back.

Removing the Anti-Aliasing line solves ghosting issues for me.

I'd guess the discontinuity is a result of display scaling. One pixel does not necessarily map to one unit on the Qt Canvas. It seems like a rounding error.

I'm not opposed to a mix-and match approach, as long as the code is reasonable.

Thanks again for your contributions! 👍

jgehrig:
Here's another idea to consider:

Undercurl/Underline should really be rendered as a continuous block. When `:GuiRenderLigatures 1` is set, we could render the entire line in one block.

This would be a more involved change though...

gebner:
> Undercurl/Underline should really be rendered as a continuous block. When :GuiRenderLigatures 1 is set, we could render the entire line in one block.

This would definitely be nicer, but as you say it's more involved.  And then you need to make sure that partial repaints work fine, etc.

I've made a few tweaks that fix the rendering for me at all font sizes (tested at 3pt, 4pt, 10pt, 30pt).  Can you test them again?
 - Drawing the line a bit further until `cellRect.right()+1` (instead of `cellRect.right()`) seems to fix the anti-aliasing rendering artifacts.
 - I've enabled anti-aliasing, but only for the undercurl.
 - I've added various maximum and minimum values to make sure that the pen width is large enough and the curl remains visible at small font sizes.
 - Both undercurl and underline are now one pixel lower than before (they're still inside the cell, if you move the cursor around, the undercurl now ends exactly at the bottom-most row.)

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -273,8 +266,13 @@ void ShellWidget::paintUndercurl(
 		return;
 	}
 
-	p.setPen(getSpecialPen(cell));
-	p.drawPath(GetUndercurlPath(cellRect));
+	QPen specialPen = getSpecialPen(cell);
+	p.setPen(specialPen);
+	// only enable anti-aliasing for the undercurl (causes rendering
+	// artifacts when rendering the cell background)
+	p.setRenderHint(QPainter::Antialiasing);
```
jgehrig, line 273: Since `p` is an out-param it has the side-effect of setting `QPainter::Antialiasing`.

Let's be extra cautious and set/un-set the value, since antialiasing causes issues elsewhere:
```
const auto renderHints { p.renderHints() };
... /* Your existing code */
p.setRenderHints(renderHints);
```

Do quickly verify this code actually works...

I didn't observe anything wrong with your code, just trying to prevent accidental bugs :)

jgehrig:
Both issues are fixed on my machine.

Everything looks good to me, thanks for the prompt fixes!

I'll merge this once I have a chance to test on MacOS and Windows.

---

I'm on KDE Plasma 5.20.5.

Here's my `neovim-qt --version` for reference:
```
NVIM-QT v0.2.17.9999
Build type: Release
Compilation:-march=native -O2 -pipe -Wall -Wextra -Wno-unused-parameter -Wunused-variable -std=c++11
Qt Version: 5.15.2
Environment: 
  nvim: nvim
  args: --cmd let &rtp.=',/usr/bin/../share/nvim-qt/runtime' --cmd set termguicolors
  runtime: /usr/bin/../share/nvim-qt/runtime

NVIM v0.5.0-dev+1119-gf2fc44d50-dirty
Build type: Gentoo
Lua 5.1
Compilation: /usr/bin/x86_64-pc-linux-gnu-gcc -march=native -O2 -pipe -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=1 -I/tmp/portage/app-editors/neovim-9999/work/neovim-9999_build/config -I/tmp/portage/app-editors/neovim-9999/work/neovim-9999/src -I/usr/include -I/tmp/portage/app-editors/neovim-9999/work/neovim-9999_build/src/nvim/auto -I/tmp/portage/app-editors/neovim-9999/work/neovim-9999_build/include
Compiled by portage@localhost

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "/etc/vim/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

src/gui/shellwidget/shellwidget.cpp
gebner: Seems like a good idea.  I've added it to the PR.

jgehrig:
Looks great on MacOS/Windows too... I've merged the changes.

Thanks for your contribution! :)

PR #875, opened 2021-07-06
Issue 868: GuiFont/guifont not updated correctly

jgehrig:
**Issue #868:** `:GuiFont` and `:set guifont` are not updated correctly.

Uses callbacks to set guifont and GuiFont any time the font is changed.

The values of both variables are read, and if they do not match
fontDesc(), they are updated.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -81,7 +81,7 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 	if (guiFont.canConvert<QString>())
 	{
 		QString fontDescription{ guiFont.toString() };
-		setGuiFont(fontDescription, true /*force*/, true /*updateOption*/);
+		setGuiFont(fontDescription, true /*force*/);
 	}
 
 	if (!m_nvim) {
@@ -114,20 +114,20 @@ void Shell::fontError(const QString& msg)
 
 /// Set the GUI font, or display the current font
 ///
-/// @param updateOption controls update of the guifont option, which should not
-/// be updated when the user calls from 'set guifont=...'.
-bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
+/// @param fdesc Neovim font description string, "Fira Code:h11".
+/// @param force used to indicate :GuiFont!, overrides mono space checks.
+/// @returns `true` if the font was successfully set.
+bool Shell::setGuiFont(const QString& fdesc, bool force) noexcept
 {
-	// Exit early if the font description does not change, prevent loops.
+	// Exit early if the font description has not changed
 	if (fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0) {
 		return false;
 	}
 
-	bool setShellFontSuccess{ false };
 	const bool isGuiDialogRequest{ fdesc == "*" };
 
 	if (isGuiDialogRequest) {
-		bool fontDialogSuccess = false;
+		bool fontDialogSuccess{ false };
 		QFont font{ QFontDialog::getFont(&fontDialogSuccess, QWidget::font(), this, {},
 			QFontDialog::MonospacedFonts) };
 
@@ -136,7 +136,9 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool updateOption)
 			return false;
 		}
 
-		setShellFontSuccess = setShellFont(font, force);
+		if (!setShellFont(font, force)) {
+			return false;
+		}
 	}
 	else {
 		QVariant varFont{ TryGetQFontFromDescription(fdesc) };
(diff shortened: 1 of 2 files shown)


src/gui/shell.cpp
```diff
@@ -210,6 +207,51 @@ bool Shell::setGuiFontWide(const QString& fdesc) noexcept
...
+	connect(getOption, &MsgpackRequest::finished, this, &Shell::handleGuiFontOption);
+
+	// Update `:GuiFont`, but only if value changes
+	MsgpackRequest* getVariable{ m_nvim->api0()->vim_get_var("GuiFont") };
+	connect(getVariable, &MsgpackRequest::finished, this, &Shell::handleGuiFontVariable);
+}
+
+void Shell::writeGuiFontQSettings() noexcept
+{
+	QSettings settings;
```
statiolake, line 227: Don't we need to specify arguments, as in `src/gui/shell.cpp`? Actually the font was not restored in the next startup in my environment without this change.
```suggestion
	QSettings settings { "nvim-qt", "nvim-qt" };
```

statiolake:
Thanks for the improvement!

I found some cases where it doesn't work as expected, but I haven't determine the root cause of them yet. I'll look more closely later.

src/gui/shell.cpp
jgehrig: Yes, currently.

Ideally, I'd like to add some `QSettings` test coverage. This is a byproduct of some soon-to-be-added changes, the new call structure is necessary for a test-mock.

Side note: How did you get this cool "Commit suggestion" dialog?

jgehrig:
> I found some cases where it doesn't work as expected

Thanks! If you can (quickly) describe the scenario's, I'll try to add some test coverage.

If it is easier for you to just debug yourself, that's fine too.

Hopefully, we can get some of these corner cases covered in `tst_shell`.

statiolake:
Most of unexpected behaviors were caused by ginit.vim I had forgot to remove after previous testing, so they were not related to this PR.

One important problem still remains however: this doesn't fix the original issue #868. Same reproduction steps will trigger the problem.

The cause of the issue #868 is that `guifont` cannot be updated during initialization: Neovim is not started when we load saved fonts from QSettings and set it as a default font. Then `guifont` won't be updated after `init.vim` is loaded, since in most cases there's no font changes on the shell side.

The below is the details of what I think is a problem. Assume the user configured `Consolas:h11` as their default font, i.e. they have following in their init.vim:

```vim
autocmd VimEnter * GuiFont! Consolas:h11
```

1. (The previous run, Neovim-qt saves `Consolas:h11` to QSettings.)
1. User launches `nvim-qt`.
1. Neovim-qt loads initial font from QSettings and set `Consolas:h11` to the shell font. We can't update `guifont` option because Neovim is not started yet.
1. Neovim-qt starts backend Neovim process. At this time, of course `guifont` is empty.
1. Neovim loads `init.vim` and run `VimEnter` autocmd when it's ready.
1. `GuiFont! Consolas:h11` is evaluated and Neovim send the event to the Neovim-qt.
1. During handling the event, in `setGuiFont()`, Neovim-qt compares `fontDesc()` (which is `Consolas:h11` loaded from QSettings) and the passed new font name `fdesc` (which is also `Consolas:h11` specified in the command). They are same, so the all of the rest process are skipped.
    https://github.com/equalsraf/neovim-qt/blob/b1b15517a3b5c4800f701b72505af02785a25f67/src/gui/shell.cpp#L124-L127
1. Since all process was skipped, `guifont` is not updated, so `guifont` is still empty.

To avoid this, I think we need to either specify the initial value for `guifont` (but is it possible?) or always update `guifont` (and this is #869).

src/gui/shell.cpp
statiolake: > Side note: How did you get this cool "Commit suggestion" dialog?

I think this is very useful feature too :)
Usage is very simple. When you add a comment to the code as usual, press the icon indicated below.

image

Then the commenting line will be automatically inserted as a code block of type `suggestion`.

image

You can make a suggestion by editing the inserted code block.

image

Finally you will get the following dialog.

image

jgehrig:
@statiolake 

Can you quickly test this again?

I think I've fixed the scenario you call out above:
https://github.com/equalsraf/neovim-qt/pull/875/commits/4993db2790899466158c48368a4d4492b86b934b

Since when we load from `QSettings`, we are not updating `:GuiFont` or `:set guifont`, we should call `updateGuiFontRegisters()` once Neovim is attached.

The code still needs some refactoring, and I may add more test coverage. Let me know if there are specific `:GuiFont` scenarios you feel should be explicitly test-covered.

statiolake:
Thanks! Very nice. After a quick check, it seems the latest code solves the issue.

I found one small difference between Neovim-qt and GVim with this implementation. The `guifont` will always be set even when user runs Neovim-qt with empty init.vim and ginit.vim. On the contrary, `guifont` is empty in that case in GVim. I don't think this is a big problem because problems caused by the difference in behavior will rarely occur. (We can also justify it by considering our QSettings as a *persistent user configurations* instead of *adapted default values*.)

Issue #808, opened 2020-12-29
Welcome message doesn't appear in nvim-qt


ndavidpsi:
The welcome message that shows the version of nvim, etc. doesn't appear in nvim-qt. Taking a closer look it seems that it briefly appears when nvim-qt is opened but then the screen clears right away.

jgehrig:
Personally, I like this behavior. I prefer the clean screen when I open my editor.

I'd guess the lack of welcome screen is a side effect of our current startup logic:
1. Start blank Qt MainWindow
2. Spawn `nvim`, initiate connection.
3. Once connected resize Neovim canvas to fit.

>Taking a closer look it seems that it briefly appears when nvim-qt is opened but then the screen clears right away.

How did you observe this? Do you see an empty screen flickering or a quick flash of the welcome message?

ndavidpsi:
I saw a quick flash of the welcome message. It's not even noticeable all the time but I have seen it a few times. So is it possible to make it show the welcome message or is it too much trouble?

jgehrig:
Okay. I don't think I can't see the flash, empty screen only.

Unfortunately, I don't currently have the time to investigate. For me, this is low priority.

If you'd like to investigate yourself, I'm open to a merge/fix and am happy to answer any questions you have about the code.

ndavidpsi:
Yes I understand, thanks. I'll look into it

jgehrig:
#197

Issue #614, opened 2019-12-11
Waiting for UI takes long time


statiolake:
Neovim-Qt takes 2 - 3 secs to startup.  This is a part of startup log obtained by `nvim-qt.exe -- --startuptime startup.log`:

```
times in msec
 clock   self+sourced   self:  sourced script
 clock   elapsed:              other lines

000.028  000.028: --- NVIM STARTING ---
021.439  021.411: locale set
022.292  000.853: inits 1
022.321  000.030: window checked
022.332  000.011: parsing arguments
022.523  000.190: expanding arguments
022.602  000.079: inits 2
026.355  003.753: init highlight
027.260  000.906: waiting for UI
2700.525  2673.265: done waiting for UI
2700.551  000.026: initialized screen early for UI
2700.622  000.071: --cmd commands

(...snip...)

3151.830  000.010: --- NVIM STARTED ---
```

Even if I use `--clean` flag the result is almost same (except ommited parts on above output, as `--clean` doesn't load my vimrc), so I believe this is not related to my Vim configs or plugins.  Also there's no problem with console version of Neovim, it takes only 30-40 msecs in `waiting for UI`.

## Environment

- OS: Windows 10 1903 18917.1000 (Insider Preview)
- Antivirus software: Windows Security (builtin)
    - I excluded `nvim.exe`, `nvim-qt.exe` and directory Neovim is installed in.
    - Anti-malware software doesn't seem to be active during Neovim-qt is starting up according to its CPU usage, so I think it is not due to it.
- Neovim version: NVIM v0.5.0-233-g76d1e9360
- Neovim-Qt version: 0.2.15
    - I downloaded it from "Release" tab on Github.
    - The version included in Neovim distribution also has the same issue.

jgehrig:
This sounds similar to Issue #606

Can you downgrade nvim.exe and see if this issue persists?
`nvim-qt --nvim {path to old version of nvim}`

statiolake:
I retried measurement with the same setting (Neovim 0.5.0 with Neovim-Qt 0.2.15):

```
045.724  001.054: waiting for UI
3558.373  3512.649: done waiting for UI
```

I don't know why it is slower than yesterday, but for now I focus on the difference between various versions.

I downloaded Neovim 0.4.3, 0.3.8 and 0.3.0. Then I ran the bundled version of Neovim-qt.

Neovim 0.4.3 (current stable):

```
062.247  003.056: waiting for UI
[...]

Issue #197, opened 2016-10-11
No intro message when opening nvim-qt


wandersoncferreira:
Hi, 
I checked the set shortmess? option and it looks identical to my Gvim application which is working normally:

set shortmess=filnxtToO

But still there is only a blank screen when I open the nvim. I am on Fedora 24 KDE.

justinmk:
This is a known nvim bug. I'll link the issue later.

PR #898, opened 2021-07-25
Set initial editor size properly

jgehrig:
**Pull Request #866:** Set initial editor size properly

Created to rebase + merge #866.

Slight modifications required due to #875.

### DIFF
--- a/contrib/appveyor.yml
+++ b/contrib/appveyor.yml
@@ -21,7 +21,7 @@ matrix:
 install:
 - ps: appveyor DownloadFile -FileName Neovim.zip "https://github.com/neovim/neovim/releases/download/nightly/nvim-win64.zip"
 - 7z x Neovim.zip
-- set PATH=%PATH%;%CD%\Neovim\bin;
+- set PATH=%PATH%;%CD%\nvim-win64\bin;
 - nvim --version
 # sh.exe must not be in the PATH
 - if "%PlatformToolset%"=="MinGW-Win64" set PATH=%PATH%;C:\Qt\5.15.1\mingw81_64\bin;C:\Qt\Tools\mingw810_64\bin;C:\Program Files\Git\usr\bin
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -283,11 +283,11 @@ void App::showUi() noexcept
 	}
 
 	if (m_parser.isSet("fullscreen")) {
-		win->delayedShow(NeovimQt::MainWindow::DelayedShow::FullScreen);
+		win->showFullScreen();
 	} else if (m_parser.isSet("maximized")) {
-		win->delayedShow(NeovimQt::MainWindow::DelayedShow::Maximized);
+		win->showMaximized();
 	} else {
-		win->delayedShow();
+		win->show();
 	}
 #endif
 }
@@ -481,7 +481,7 @@ void App::openNewWindow(const QVariantList& args) noexcept
 	ConnectorInitArgs params{ type, 2000, std::move(server), std::move(nvim), getNeovimArgs() };
 	MainWindow* win{ &createWindow(params) };
 	win->resize(s_lastActiveWindow->size());
-	win->delayedShow();
+	win->show();
 }
 
 } // namespace NeovimQt
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -126,8 +126,6 @@ void MainWindow::init(NeovimConnector *c)
 /** The Neovim process has exited */
 void MainWindow::neovimExited(int status)
 {
-	showIfDelayed();
-
 	if (m_nvim->errorCause() != NeovimConnector::NoError) {
 		m_errorWidget->setText(m_nvim->errorString());
 		m_errorWidget->showReconnect(m_nvim->canReconnect());
@@ -142,8 +140,6 @@ void MainWindow::neovimExited(int status)
 }
 void MainWindow::neovimError(NeovimConnector::NeovimError err)
 {
-	showIfDelayed();
-
 	switch(err) {
 	case NeovimConnector::FailedToStart:
 		m_errorWidget->setText("Unable to start nvim: " + m_nvim->errorString());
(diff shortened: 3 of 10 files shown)


jgehrig:
@statiolake

This is a rebase-only... Your changes unmodified.

I did notice some flicker on startup:
Startup

Do you see this flicker? I don't think this fixes #212...

With the PR above, startup was flicker-free. The changes as written are not compatible with each other.

Ideally, we would fix both issues. If we have to pick only one, proper startup size is definitely more important.

statiolake:
Does "flicker" in your screenshot mean the initial intro message shown for a moment? If so, I don't notice such a flicker on Windows, but that might be hidden under its opening window animation. Note that usually I'm using `shortmess+=I` so I'm not bothered by it in daily life.

Instead I see the blank window on startup and I see it for a longer time with my patch. I didn't cared about it because it is common in Windows applications to have a blank window right after opening. GVim also has this kind of flicker. (As a side note, GVim has more "dynamic" flicker, since it keeps the lines and columns instead of window size. `set guifont` causes window resize.)

At least, in terms of flicker, this PR doesn't improve anything. It even removes the delaying mechanism introduced to reduce flickers.

For now I don't have a better way to achieve both of flickerless startup and proper editor size. We need to postpone displaying the window until everything gets ready to remove flicker. But we need to show the window before starting Neovim instance to know exact window size to determine the initial editor size.

A possible hacky way is to use saved geometry or explicitly specified value to estimate the next window size. We can expect the next window size is same with the previous window, as we are already trying to restore geometry on startup. But the window size request is not always satisifiable so actual window size may differ from the expectation. In that case the editor size will be still broken.

jgehrig:
@statiolake 

Thanks for the detailed explanation. This all sounds reasonable.

Yes, I think `shortmess+=I` would eliminate the flicker I see.

> Instead I see the blank window on startup and I see it for a longer time with my patch.

Not a problem. I think proper startup size is much more important than increased blank-window time.

I view blank-window display as a non-issue. The critical metric (IMO) is startup time, and I don't think your changes have a significant effect on overall startup time.

**Side Note:**
At some point, I would like to write a CI rule that measures the startup-time differences between baseline and PR.
Measuring startup is critical to improving/preventing scenarios like your Issue #614.

> At least, in terms of flicker, this PR doesn't improve anything.

Not a problem, just curious.

> A possible hacky way

No need, I'd rather have clean code with some flicker.

jgehrig:
@statiolake 

During some pre-merge testing, I discovered we don't account for space taken by `GuiScrollBar`, `GuiTabline`, or `GuiTreeView`.

We need to expand upon cafc9ca9f03e7f4104cc81c0330ae220e5f23497, and include additional options such as `Gui/ScrollBar`. Once we ensure that last known state for each of these controls is set, we can determine the correct size of `Shell` on startup.

Some flicker is caused by GUI widgets (we can fix this). The rest is due to the intro message, #808 #197 (we can ignore this).

Let me know how you'd like to handle these fixes.

statiolake:
I've completely forgot about them since I don't show them at all...

Let me confirm my understanding. GUI widgets shown after initialization through init.vim's `:Gui*` commands change the editor size after Neovim splits the editor. For example, `GuiTreeviewShow` expands tree view from the left side of the editor, so the vertical split bar seems off to the right as a result. (I couldn't confirm it on my side because `GuiTreeviewShow` always filled up entire window size... Maybe I did something wrong.)

If that is true, as you said, just including their states in QSettings automatically solves this issue too.

I wish I could implement restoring those widgets by myself, but I'm not sure I have enough time to do it...

jgehrig:
@statiolake

> I've completely forgot about them since I don't show them at all...

Yes, me too... Otherwise, I would have caught this sooner.

Your understanding of widget behavior sounds correct to me.

That `:GuiTreeView` behavior is a bug. The widget should not behave like that. I think the issue is gone in `master`?

> I wish I could implement restoring those widgets by myself, but I'm not sure I have enough time to do it...

Not a problem :)

I'll take care of the issues and merge your changes from this PR. I wanted to ask before modifying your code...

statiolake:
> Not a problem :)
>
> I'll take care of the issues and merge your changes from this PR. I wanted to ask before modifying your code...

Thank you as always. Although I think changing QSettings should be in another PR, I'm fine with modifications by you.

jgehrig:
> Although I think changing QSettings should be in another PR

Yes, I'll split things up. Probably all within the same PR, but different commits. Same end result :)

Assuming all goes smoothly with the `:Gui` issues, I'll try to get this merged within the next few days.

Thank you for yet another great (and much needed) improvement!

jgehrig:
@statiolake 

Update, I haven't forgotten about this PR :)

I've noticed some issues when the window is restored from a previously maximized state... The initial size reported by Qt is not correct. I am trying to understand exactly what is going wrong, but it is not yet clear to me.

I'm delaying the merge until I better understand a root cause.

statiolake:
Sorry for late reply.

> I've noticed some issues when the window is restored from a previously maximized state...

Could you tell me how to test it? Just now I tested branch `jgehrig:jg-editor-size` on Windows but I didn't notice the problem. I confirmed that `nvim-qt -- -O a b` correctly shows the maximized window divided vertically.

jgehrig:
@statiolake

No worries.

The "issue" here is subtle. I'm not even sure it results in a perceivable difference...

If you look at the debug logs, you multiple calls to `handleResize: X Y` with different values before settling at a final size.

**Repro Steps:** (tested on Linux)
1. Open nvim-qt, maximize window, close
2. Re-open nvim-qt
3. Notice the `handleResize` messages report multiple different sizes. The first is incorrect.

If you do the same set of steps without the window previously maximized, the size is reported correctly immediately.

Note: Windows doesn't print `qDebug()` messages by default. I think you'll need to set `NVIM_QT_LOG`.

jgehrig:
@statiolake 

Sorry for the LONG delay here...

I've merged your fix. I know lots of users will appreciate this, it has appeared in the Issue Tracker several times.

Thanks for your contributions! :+1:

Issue #923, opened 2021-08-23
Regression in highlight rendering antialiasing artifacts


jgehrig:
**Repro Steps**
1. `nvim-qt -- -u NONE`
2. `:badd $VIMRUNTIME/mswin.vim`
3. `:bnext`
4. `ggVG`

Notice there are highlight rendering issues:

![image](https://user-images.githubusercontent.com/11207308/130488361-3d9fe505-1be7-4d1f-83a2-2d25d516aa57.png)
---

@gebner 

A `git bisect` suggests this issue was introduced with your change 683e7c6ee58a805a86a1b62ba07ed6a6dd886088.

I'd guess that this line is leaking out of it's intended scope, or is otherwise causing issues:
https://github.com/equalsraf/neovim-qt/blob/9b1ed37b02f3ceb768563ecdd97cf56d1c567dbb/src/gui/shellwidget/shellwidget.cpp#L275

PR #934, opened 2021-09-21
Issue 923: Antialiasing artifact fix

jgehrig:
**Issue #923:** Antialiasing flag is not unset after rendering undercurl.

The API for `QPainter::renderHints()` does not work as we expected.

`QPainter::setRenderHints` clears or sets all flags present. It does not set the aggregate value.

We need to manually set/clear the `Antialiasing` flag.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -269,12 +269,13 @@ void ShellWidget::paintUndercurl(
 
 	QPen specialPen = getSpecialPen(cell);
 	p.setPen(specialPen);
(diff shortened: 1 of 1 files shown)


gebner:
> QPainter::setRenderHints clears or sets all flags present. It does not set the aggregate value.

Good catch!

Sorry for not noticing this earlier.  I could never reproduce the rendering artifacts (also not with the repro from #923).

jgehrig:
@gebner

No worries!

Yes, I think display scaling is required to reproduce the artifacts.

Yeah, that API is not obvious... I think the bug is from code I suggested offhand without explicit testing. We both missed this one.

PR #946, opened 2021-10-30
Debug messages should only work in 'Debug' builds

### LINKED ISSUE #945: Build should not produce debug messages for build type RelWithDebInfo
Currently Debug output is enabled for all builds except for CMAKE_BUILD_TYPE == **Release**. However some packagers expect to build with **RelwithDebInfo** (for release packages with separate debug symbols).

The relevant code is

https://github.com/equalsraf/neovim-qt/blob/19b08c430c93dd1a78f5d2fdd081bbc6f530a17d/CMakeLists.txt#L154

In the interest of "no surprises", I would suggest enabling these only for Debug builds.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index fa9e2a4..a20f749 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,7 +168,7 @@ else()
 endif()
 include_directories(${MSGPACK_INCLUDE_DIRS})
 
-if(CMAKE_BUILD_TYPE STREQUAL "Release")
+if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
     add_definitions(-DQT_NO_DEBUG_OUTPUT)
 endif()
```

If that is too much, we can just add RelwithDebInfo to that check.

equalsraf:
closes #945

As a side note, in my system (cmake 3.21 in Linux) the default build type is  RelWithDebInfo and even if I force CMAKE_BUILD_TYPE=Release the build is still RelWithDebInfo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,7 +151,7 @@ else()
 endif()
 include_directories(${MSGPACK_INCLUDE_DIRS})
 
-if(CMAKE_BUILD_TYPE STREQUAL "Release")
+if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
     add_definitions(-DQT_NO_DEBUG_OUTPUT)
 endif()
 


jgehrig (approved):

jgehrig:
> As a side note, in my system (cmake 3.21 in Linux) the default build type is RelWithDebInfo and even if I force CMAKE_BUILD_TYPE=Release the build is still RelWithDebInfo.

Maybe check your environment variables, I think you can set `CMAKE_BUILD_TYPE`?

Although, I would expect the `-D` command line options to override that... Perhaps your distro patches or provides some extra config to CMake.

Issue #954, opened 2021-12-06
[Question] neovim-qt frameless


yyancy:
Hi, I want to know if there is a way to make neovim-qt frameless. 
Thanks.

jgehrig:
Are you looking for something like `:call GuiWindowMaximized(1)`.

Or are you looking for a non-maximized window without the border/frame?

If you can put together a PR for this, I am willing to review and merge. See the implementation of `GuiWindowMaximized` in `/src/gui/runtime/plugin/nvim_gui_shim.vim`.

That would need to call a new function in `src/gui/mainwindow.cpp`, which sets and un-sets the flag `Qt::FramelessWindowHint`.

See:
https://doc.qt.io/qt-5/qwindow.html#setFlag
https://doc.qt.io/qt-5/qt.html#WindowType-enum

yyancy:
Thanks for your reply, I will give it a try!

khalidchawtany:
I am using this patch https://github.com/khalidchawtany/neovim-qt/commit/a99fa3cb392e327b27d15220b5e6bc7cec9a4728 to allow FrameLess windows in macOS. It works great. Everything works from dragging to resizing etc. 

![2020-01-28 09 45 00](https://user-images.githubusercontent.com/1681606/73241465-1f66be80-41b3-11ea-937a-35e49f527a68.gif)

yyancy:
Thanks your advice. but I I'm not very good at C++，so I may not be able to implement it.

PR #959, opened 2022-01-25
Implement frameless window

f-squirrel:
Changes:
* Add function `GuiWindowFrameless`
* Update neovim with `g:GuiWindowFrameless`
* Update help
* Update tests

Fixes https://github.com/equalsraf/neovim-qt/issues/954.

@jgehrig , please review :)

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -191,12 +191,26 @@ void MainWindow::neovimMaximized(bool set)
 
 void MainWindow::neovimFrameless(bool isFrameless)
 {
-	setWindowFlag(Qt::FramelessWindowHint, isFrameless);
-	// DD: Need call show o make the widget visible again.
-	// https://doc.qt.io/qt-5/qwidget.html#windowFlags-prop
+	// When minimum Qt 5.9 is supported, use this instead:
+	//  setWindowFlag(Qt::FramelessWindowHint, isFrameless);
+	if (isFrameless)
+	{
+		setWindowFlags(windowFlags() | Qt::FramelessWindowHint);
+	}
+	else
+	{
+		setWindowFlags(windowFlags() & ~Qt::FramelessWindowHint);
+	}
+
+	// Calling setWindowFlags can cause the widget to be hidden, show() must
+	// be called to make the widget visible again. On some platforms, this may
+	// cause focus issue. See Issue#971.
+	//
+	// Details: https://doc.qt.io/qt-5/qwidget.html#windowFlags-prop
 	show();
-	// DD: It seems there is no event representing the change of flags
+
 	m_nvim->api0()->vim_set_var("GuiWindowFrameless", isFrameless ? 1 : 0);
+
 }
 
 void MainWindow::neovimForeground()
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -91,6 +91,8 @@ void MainWindow::init(NeovimConnector *c)
 			this, &MainWindow::neovimSuspend);
 	connect(m_shell, &Shell::neovimFullScreen,
 			this, &MainWindow::neovimFullScreen);
+	connect(m_shell, &Shell::neovimFrameless,
+			this, &MainWindow::neovimFrameless);
 	connect(m_shell, &Shell::neovimGuiCloseRequest,
 			this, &MainWindow::neovimGuiCloseRequest);
 	connect(m_shell, &Shell::neovimOpacity,
@@ -185,6 +187,16 @@ void MainWindow::neovimMaximized(bool set)
 	} else {
 		setWindowState(windowState() & ~Qt::WindowMaximized);
 	}
+}
+
+void MainWindow::neovimFrameless(bool set)
+{
+	setWindowFlag(Qt::FramelessWindowHint, set);
+	// DD: Need call show o make the widget visible again.
+	// https://doc.qt.io/qt-5/qwidget.html#windowFlags-prop
+	show();
+	// DD: It seems there is no event representing the change of flags
+	m_nvim->api0()->vim_set_var("GuiWindowFrameless", set ? 1 : 0);
 }
 
 void MainWindow::neovimForeground()
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -189,14 +189,14 @@ void MainWindow::neovimMaximized(bool set)
 	}
 }
 
-void MainWindow::neovimFrameless(bool set)
+void MainWindow::neovimFrameless(bool isFrameless)
 {
-	setWindowFlag(Qt::FramelessWindowHint, set);
+	setWindowFlag(Qt::FramelessWindowHint, isFrameless);
 	// DD: Need call show o make the widget visible again.
 	// https://doc.qt.io/qt-5/qwidget.html#windowFlags-prop
 	show();
 	// DD: It seems there is no event representing the change of flags
-	m_nvim->api0()->vim_set_var("GuiWindowFrameless", set ? 1 : 0);
+	m_nvim->api0()->vim_set_var("GuiWindowFrameless", isFrameless ? 1 : 0);
 }
 
 void MainWindow::neovimForeground()
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -342,6 +342,9 @@ void Shell::init()
 
 	// Subscribe to GUI events
 	m_nvim->api0()->vim_subscribe("Gui");
+
+	// Set initial value
+	m_nvim->api0()->vim_set_var("GuiWindowFrameless", (windowFlags() & Qt::FramelessWindowHint) ? 1: 0);
 }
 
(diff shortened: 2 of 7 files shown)


src/gui/shell.cpp
```diff
@@ -835,6 +838,18 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			} else {
 				emit neovimFullScreen(variant_not_zero(args.at(1)));
 			}
+		} else if (guiEvName == "WindowFrameless" && args.size() == 2) {
+			if (isWindow()) {
+				auto isOn = variant_not_zero(args.at(1));
```
jgehrig, line 843: nitpick: Use the real type + a descriptive name.

```suggestion
				const bool isWindowFrameOn{ variant_not_zero(args.at(1)) };
```

src/gui/shell.cpp
```diff
@@ -835,6 +838,18 @@ void Shell::handleNeovimNotification(const QByteArray &name, const QVariantList&
 			} else {
 				emit neovimFullScreen(variant_not_zero(args.at(1)));
 			}
+		} else if (guiEvName == "WindowFrameless" && args.size() == 2) {
```
jgehrig, line 841: Please move this block into a small function such as `handleWindowFrameless`.

You should be able to follow a similar pattern to `handleLineSpace` below.

src/gui/mainwindow.cpp
```diff
@@ -187,6 +189,16 @@ void MainWindow::neovimMaximized(bool set)
...
 }
 
+void MainWindow::neovimFrameless(bool set)
+{
+	setWindowFlag(Qt::FramelessWindowHint, set);
+	// DD: Need call show o make the widget visible again.
+	// https://doc.qt.io/qt-5/qwidget.html#windowFlags-prop
+	show();
+	// DD: It seems there is no event representing the change of flags
+	m_nvim->api0()->vim_set_var("GuiWindowFrameless", set ? 1 : 0);
```
jgehrig, line 199: Let's avoid the `foo ? 1 :0` ternary (if it is not too much effort).

You might be able to leverage `variant_not_zero(set)`

jgehrig: Here and elsewhere

src/gui/mainwindow.cpp
```diff
@@ -187,6 +189,16 @@ void MainWindow::neovimMaximized(bool set)
 	}
 }
 
+void MainWindow::neovimFrameless(bool set)
```
jgehrig, line 192:
```suggestion
void MainWindow::neovimFrameless(bool isFrameless)
```

nitpick: The name `set` is a little ambiguous.

jgehrig:
@f-squirrel 

This is awesome! Thank you for putting this Pull Request together :clap:

I've added some small insignificant comments. The code looks good. I will do some testing and then merge the changes. We should test on all platforms: Linux, Windows, and MacOS.

src/gui/mainwindow.cpp
f-squirrel: done

src/gui/shell.cpp
f-squirrel: done

src/gui/mainwindow.cpp
f-squirrel: `vim_set_var(str, bool)` generates `v:bool` in vimscript while the variable and function accept integers (I did it the same way as `GuiWindowMazimized` and others).

f-squirrel:
> @f-squirrel
> 
> This is awesome! Thank you for putting this Pull Request together clap
> 
> I've added some small insignificant comments. The code looks good. I will do some testing and then merge the changes. We should test on all platforms: Linux, Windows, and macOS.

Thank for the review, please see my answers and fixes in the new commit.
The feature has been tested on Ubuntu 20.04; unfortunately, I don't have macOS or Windows.

P.S. Clazy Github action failed on downloading some dependency.

f-squirrel:
hey folks, @jgehrig and @equalsraf , would you mind reviewing my changes?

jgehrig:
@f-squirrel

Your changes have my approval! :+1: 

The final hold-up is testing on MacOS/Windows. I will merge once someone has tested on the other platforms.

It is on my TODO list, but life is getting the better of me right now :)

src/gui/mainwindow.cpp
```diff
@@ -187,6 +189,16 @@ void MainWindow::neovimMaximized(bool set)
 	}
 }
 
+void MainWindow::neovimFrameless(bool isFrameless)
+{
+	setWindowFlag(Qt::FramelessWindowHint, isFrameless);
+	// DD: Need call show o make the widget visible again.
```
jgehrig, line 195: Pure curiosity... What does "DD:" stand for?

I've never run across this acronym  before. Totally fine to keep as-is!

jgehrig (approved):

src/gui/mainwindow.cpp
f-squirrel: DD is my initials (Dmitry Danilov), sorry for the confusion.

f-squirrel:
@jgehrig 

I have tested the PR on macOS with a little help from my friend @little-zaets.

Performed tests:
1. The initial value of `g:GuiWindowFrameless` is `0`
1. Call `GuiWindowFrameless(1)` -> The window becomes frameless
1. The value of `Gg:uiWindowFrameless` is `1`
1. Call `GuiWindowFrameless(0)` -> The window frame is shown
1. The value of `g:GuiWindowFrameless` is `0`

Please find the screen recording below:
nvim

macOS info:
photo_2022-02-11_19-35-52

P.S. Windows is more problematic because I haven't worked with it for ages.

jgehrig:
@f-squirrel

Sorry for the delay.

I quickly tested the feature on MacOS and Windows. Both look good.

However, I do see one issue on Linux:
Linux-Issue-Cursor-Focus

I'm running KDE/Plasma on the machine displaying the focus issues. The focus issue goes away after cycling the focus.

1. Do you see these issues?
2. What Desktop Environment are you using?

We should investigate and fix if possible.

I don't think this issue is a check-in blocker... If you cannot reproduce the issue or we cannot find an easy solution, we can file an Issue to track this; the problem is minor.

f-squirrel:
@jgehrig ,
No problem!

I do not see these issues, to be honest.
Maybe the application loses focus once we change the style?
Not sure how to debug it.

I am using Ubuntu 20.04, please find more info below:
screen-fetch

jgehrig:
> I do not see these issues, to be honest.

No worries. This is actually what I had hoped to hear!

I'm guessing this issue is specific to my Desktop Environment (Plasma/KDE), since you're running GNOME.

>  Maybe the application loses focus once we change the style?

On my machine, the window has focus but calls to `hasFocus()` inside `ShellWidget` return `false`.

It looks like we receive a `QFocusEvent(FocusOut, ActiveWindowFocusReason)` when the `Qt::FramelessWindowHint` flag is set. The Window regains focus on `show()`, but there is no corresponding `FocusIn` event for ShellWidget.

I tried quickly calling some functions to explicitly force focus into `QMainWindow` and `ShellWidget`, but this has no effect. This smells like a Qt or Plasma bug to me.

> Would you mind trying to reproduce it with the following:
>
> NVIM_QT_RUNTIME_PATH=/usr/local/share/nvim-qt/runtime nvim-qt -- -i NONE -u NONE

The issue still reproduces.

src/gui/mainwindow.cpp
```diff
@@ -187,6 +189,16 @@ void MainWindow::neovimMaximized(bool set)
...
+	// https://doc.qt.io/qt-5/qwidget.html#windowFlags-prop
+	show();
+	// DD: It seems there is no event representing the change of flags
+	m_nvim->api0()->vim_set_var("GuiWindowFrameless", isFrameless ? 1 : 0);
```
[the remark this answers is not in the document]
jgehrig: A quick update to this block...

- `setWindowFlag` is not supported in the minimum Qt version, 5.8.
- A reference to the focus issue I observed.
- More details on how `setWindowFlags(...)` and `show()` interact.

f-squirrel:
hey @jgehrig,
I have applied the changes you have proposed!

khalidchawtany:
On Mac OS, the frameless window cannot be dragged nor resized. It cannot be resized in any way, not using the cursor, nor using a window manager.

khalidchawtany:
Now reading my comment I sound like a d..k. Sorry about that.

jgehrig:
@khalidchawtany

No worries! (I did not think your comment was obnoxious)

Agreed... Your patch is functionally better. However, it has cross-platform maintainability drawbacks: How do we implement equivalent behavior for Windows/Linux?

1. It is awesome to have a link to your commit here! I am certain it will be useful to MacOS users in the future. :+1: 
2. The commit is not check-in ready for general audiences
  a. No Windows/Linux equivalent implementation yet.
  b. I prefer to avoid adding new languages and platform specific code whenever possible.

jgehrig:
@f-squirrel 

Thanks for the ping... I'm just slow :)

PR #973, opened 2022-03-03
Bypass showtabline if GuiTabline is set 

NoahTheDuke:
~Displaying tabs in neovim is controlled by the option showtabline:~
* ~0 -> never display~
* ~1 -> display if more than 1 tab (default)~
* ~2 -> always display~

~Ostensibly, #739 (Improved GuiTabline show buffers) changed the behavior of neovim such that buffers will be displayed in the tabline when GuiTabline is set, but showtabline's default makes GuiTabline a no-op until a second neovim tab is opened (with :tabnew, etc). The fix is to save the showtabline value and then when calling GuiTabline, if enabled set showtabline to 2, otherwise reset to the saved value.~

~This enables the (somewhat silly) configuration of setting showtabline to 0 and then toggling GuiTabline as the only means of seeing the tabline.~

Instead of settings or unsetting showtabline, assume that if GuiTabline is set, the user wants the gui tabline to be shown unconditionally.

### DIFF
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim
@@ -96,14 +96,7 @@ function s:GuiLinespaceCommand(height) abort
 endfunction
 command! -nargs=? GuiLinespace call s:GuiLinespaceCommand("<args>")
 
-let s:stl=&showtabline
 function! s:GuiTabline(enable) abort
-  if a:enable
-    let s:stl=&showtabline
-    execute 'set showtabline=2'
-  else
-    execute 'set showtabline=' . s:stl
-  endif
   call rpcnotify(0, 'Gui', 'Option', 'Tabline', a:enable)
 endfunction
 command! -nargs=1 GuiTabline call s:GuiTabline(<args>)
--- a/src/gui/runtime/plugin/nvim_gui_shim.vim
+++ b/src/gui/runtime/plugin/nvim_gui_shim.vim
@@ -96,8 +96,15 @@ function s:GuiLinespaceCommand(height) abort
 endfunction
 command! -nargs=? GuiLinespace call s:GuiLinespaceCommand("<args>")
 
+let s:stl=&showtabline
 function! s:GuiTabline(enable) abort
-	call rpcnotify(0, 'Gui', 'Option', 'Tabline', a:enable)
+  if a:enable
+    let s:stl=&showtabline
+    execute 'set showtabline=2'
+  else
+    execute 'set showtabline=' . s:stl
+  endif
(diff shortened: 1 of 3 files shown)


jgehrig:
@NoahTheDuke 

Thank you for the Pull Request! You make a good point, the current behavior is odd.

I think the better way to fix this is within `src/gui/tabline.cpp` `Tabline::updateTablineVisibility()`.

We should be able to modify the visibility logic so that the value of `showtabline` does not impact `GuiTabline` visibility for the multiple-buffers and no-tabs case.

I think the current fix as written has a (very minor) bug?

Consider the following:
1. Apply your fix.
2. Open `nvim-qt`
3. `:set showtabline=1`
4. `:GuiTabline 1`
5. `:set showtabline=2`
6. `:GuiTabline 0`
7. Now `:set showtabline` reports 1, overriding the user-defined configuration.

If we fix this in the C++, we can avoid poking vim user options like `showtabline`.

---

Would you like to try writing the C++ fix?

If not, I can put something together when I have some spare time.

NoahTheDuke:
That's a good point, hadn't considered that edge case. I think there is a vimscript solution but handling this in the rendering code is best.

I don't know a lick of c++ but have dabbled in rust, so I'll take a look and see if I can hack something together! If not, I'll holler.

NoahTheDuke:
I can change `updateTablineVisibility` such that it ignores `m_optionShowTabline`, either showing unconditionally or not (as I have it in this PR). Do we want to keep the old functionality around in some form? Otherwise, I can rip out all of this other code.

jgehrig:
@NoahTheDuke 

I don't think we want to remove this option entirely. It provides a useful mechanism for controlling the `tabline` behavior. Instead, let's make `showtabline` look at the buffer or tab count. We will display the `GuiTabline` when the conditions for either buffers or tabs is met.

You can accomplish the same effect as this PR by adding `set showtabline=2` to your vimrc.

Let me know if f00865d19cf91008a8cde1239bae3f8392f7d79a works for you.

Note:
I also discovered a bug in how Neovim handles bufferline updates.

1.`nvim-qt -- -u NONE`
2. `:GuiTabline`
3. `:set showtabline=1`
4. `:badd foo`
5. The tab count is two. The GuiTabline should be displayed, but it is not.
6. `:bnext` causes the tabline event to fire, and the `GuiTabline` to display.

It looks like Neovim never sends the event. We may want to modify Neovim to emit an event when new buffers are added.

NoahTheDuke:
My wife just gave birth so I am away from my computer for a while but I will definitely try this out! Thanks so much for the review

jgehrig:
Congratulations!

No rush, take your time.

NoahTheDuke:
I have returned, and so far this is working great! Haven't run into any issues yet, so I appreciate the work.

Issue #981, opened 2022-04-28
Can't set negative linespace


damanis:
The commit 2c3c7c29 made impossible to set negative linespace.
Seems, there is no reason to prevent negative value, 'GuiLinespace -1' works properly (after disable checking in code).

jgehrig:
@damanis

Good catch! I (incorrectly) assumed a negative linespace value was invalid. Based on the docs, it is an expected case that we should handle.

From `:help linespace`:
```
'linespace' 'lsp'	number	(default 0)
			global
			{only in the GUI}
	Number of pixel lines inserted between characters.  Useful if the font
	uses the full character cell height, making lines touch each other.
	When non-zero there is room for underlining.
	With some fonts there can be too much room between lines (to have
	space for ascents and descents).  Then it makes sense to set
	'linespace' to a negative value.  This may cause display problems
	though!
```

I'll put together a fix.

jgehrig:
Merged. This issue should be fixed, marking as Closed.

PR #985, opened 2022-05-19
Issue 981: Allow negative linespace

jgehrig:
**Issue #981:** We should allow negative linespace, this was accidentally regressed in
commit 2c3c7c2.

Removes the invalid < 0 check.


damanis:
@jgehrig 
I'm not sure, but it's not clear for me what is changed: Shell::handleLineSpace() check '!isValid || linespace < 0' as before.

PR #1001, opened 2022-07-15
Use QFont::exactMatch to check whether font is an exact match

charmander:
Fixes GuiFont being unnecessarily picky, e.g. when the foundry isn’t included ("Roboto Mono" != "Roboto Mono [GOOG]").

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -59,14 +59,15 @@ bool ShellWidget::setShellFont(const QFont& font, bool force) noexcept
 		return false;
 	}
 
-	QFontInfo fi(font);
-	if (fi.family().compare(font.family(), Qt::CaseInsensitive) != 0 &&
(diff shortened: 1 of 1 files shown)


jgehrig:
What scenario does this fix?

I assume something like `:GuiFont Roboto Mono` currently reports an unknown font error, and with this patch the font reports correctly.

We should do some testing to make sure this patch does not cause font description string mismatch issues downstream. 

After testing, this is probably fine to merge! :thumbsup:

jgehrig:
@charmander 

Friendly ping :)

I would be happy to merge these changes if you can quickly explain the broken scenario.

charmander:
> I assume something like `:GuiFont Roboto Mono` currently reports an unknown font error, and with this patch the font reports correctly.

Sorry for the wait! Yes, that’s the scenario. I haven’t had time to look at the tests yet.

charmander:
It looks like `exactMatch` is too picky for this purpose in a different way (font size).

Issue #1005, opened 2022-07-30
cmake test fail for 0.2.17


xyproto:
Hi,

I'm packaging 0.2.17 for Arch Linux, but one of the tests seems to fail now:

```ini
==> Starting check()...
ninja: Entering directory `build'
[0/1] cd /build/neovim-qt/src/build && /usr/bin/ctest --force-new-ctest-process
Test project /build/neovim-qt/src/build
      Start  1: tst_neovimobject
 1/20 Test  #1: tst_neovimobject .................   Passed    1.53 sec
      Start  2: tst_neovimconnector
 2/20 Test  #2: tst_neovimconnector ..............   Passed   10.82 sec
      Start  3: tst_callallmethods
 3/20 Test  #3: tst_callallmethods ...............   Passed    2.01 sec
      Start  4: tst_encoding
 4/20 Test  #4: tst_encoding .....................   Passed    0.04 sec
      Start  5: tst_msgpackiodevice
 5/20 Test  #5: tst_msgpackiodevice ..............   Passed    0.04 sec
      Start  6: tst_shell
 6/20 Test  #6: tst_shell ........................   Passed    3.85 sec
      Start  7: tst_main
 7/20 Test  #7: tst_main .........................   Passed    0.07 sec
      Start  8: tst_qsettings
 8/20 Test  #8: tst_qsettings ....................***Failed    0.34 sec
      Start  9: tst_input_mac
 9/20 Test  #9: tst_input_mac ....................   Passed    0.05 sec
      Start 10: tst_input_unix
10/20 Test #10: tst_input_unix ...................   Passed    0.04 sec
      Start 11: tst_input_win32
11/20 Test #11: tst_input_win32 ..................   Passed    0.04 sec
      Start 12: tst_input_common_mac
12/20 Test #12: tst_input_common_mac .............   Passed    0.03 sec
      Start 13: tst_input_common_unix
13/20 Test #13: tst_input_common_unix ............   Passed    0.03 sec
      Start 14: tst_input_common_win32
14/20 Test #14: tst_input_common_win32 ...........   Passed    0.04 sec
      Start 15: test_cell
15/20 Test #15: test_cell ........................   Passed    1.02 sec
      Start 16: test_shellcontents
16/20 Test #16: test_shellcontents ...............   Passed    0.91 sec
      Start 17: test_shellwidget
17/20 Test #17: test_shellwidget .................   Passed    0.06 sec
      Start 18: bench_scroll
[...]

PR #1007, opened 2022-08-09
only require Qt5Test if tests are enabled

tastytea:
Compiles fine without Qt5Test when `ENABLE_TESTS` is off. 😊


jgehrig:
@tastytea 

Looks good! Nice catch, I'll merge these changes shortly.

It is also on my TODO list to set `ENABLE_TESTS` off-by-default. Tests increase the build time and can cause build issues such as #1005.

PR #1027, opened 2022-11-12
Use modern CMake and use FetchContent for msgpack

gikari:
Using fetch content cmake module is less verbose,
than custom download and extract logic with setting of additional variables.

CMake version 3.16 is in the Ubuntu 20.04 repository, so everyone should have at least it by now.


.gitignore
```diff
@@ -4,3 +4,6 @@ src/gui/runtime/doc/tags
 third-party/msgpack*
 compile_commands.json
 !.github
+
+# Qt Creator project files
+CMakeLists.txt.user
```
jgehrig, line 9: Let's remove this, better to have IDE-specific files in the user's `.gitignore`.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.16)
 if (POLICY CMP0048)
```
jgehrig, line 2: Can you get rid of this if (and keep in policy set) while bumping the CMake version?

It is no longer necessary, CMP0048 is present in CMake >= 3.0.

jgehrig: Same in the other CMakeLists.txt files:
- [src/gui/shellwidget/CMakeLists.txt](https://github.com/equalsraf/neovim-qt/blob/edd4b9f5d570343aee8016b54abd1f82754dccee/src/gui/shellwidget/CMakeLists.txt)
- [third-party/CMakeLists.txt](https://github.com/equalsraf/neovim-qt/blob/9dbd99dfaccd52a8b24b2c914831e5c7b9de6e28/third-party/CMakeLists.txt)
- [CMakeLists.txt](https://github.com/equalsraf/neovim-qt/blob/409fad81323f44e83ebcdbd4e8fa13bfff3f844c/CMakeLists.txt)

jgehrig:
Looks good, thanks for the contribution!

Some minor comments regarding the CMake version update + .gitignore. I will merge as soon as the updates are made.

.gitignore
gikari: I think, given that the project is using C++ stack and has a Qt GUI application component, it is reasonable to exclude this file, because there is a high chance a developer might be using Qt Creator.

jgehrig:
@gikari 

Have you validated these changes on Windows?

It looks like your changes could have introduced a Windows link-time build issue:
https://github.com/equalsraf/neovim-qt/actions/runs/3584393296/jobs/6031004832#step:10:235

I'm not sure if your change is responsible or not, but we do need to validate to make sure there isn't a regression here.

Sorry for the noise in our CI Pipeline. We've been broken by some recent VM updates, and fixing this is on my TODO list :/

jgehrig (changes requested):
I pushed some changes and we need to validate on Windows quickly.

gikari:
Yes, I built the neovim-qt on windows just fine with MSVC.

jgehrig:
> Yes, I built the neovim-qt on windows just fine with MSVC.

Awesome, the issue is likely with our CI then, merging the changes.

Thanks!

jgehrig (approved):

jgehrig:
@gikari 

Re-opening this issue. Can you please investigate #1036?

It looks like this Pull Request may cause some issues on Windows.

PR #1030, opened 2022-11-18
Fix small issues when using as an external project

gikari:
When used as an external project, neovim-qt had some issues:
1. Main target's include directories are not propagated to the targets, that link to neovim-qt.
2. When installing, paths to neovim icons are incorrect.

This PR fixes those issues.


jgehrig:
@gikari 

Thanks for the contribution and cool Qt Creator plugin!

Looks good, merging your changes :+1:

PR #1035, opened 2022-12-02
Use UTF8 everywhere to decode/encode

gikari:
A couple of reasons:
- Neovim uses only UTF8 for RPC
- QTextCodec is removed in Qt6

This is one of the points in #839


equalsraf:
Thanks @gikari I've merged the other PR.

PR #1038, opened 2022-12-09
Assume utf8 for nvim strings, remove QTextCodec

equalsraf:
In the (very) early days of neovim, it was possible to have non utf8 strings, but this was removed so we don't need to support it.

Picked from #840.

This is the same as #1035 but keeps the encode/decode methods to avoid changing the auto generated code.

### DIFF
--- a/src/msgpackiodevice.cpp
+++ b/src/msgpackiodevice.cpp
@@ -1,5 +1,4 @@
 #include <QAbstractSocket>
-#include <QTextCodec>
 #include <QSocketNotifier>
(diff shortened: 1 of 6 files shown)


gikari:
Maybe the methods encode/decode should be marked deprecated, so that the users would prefer using built-in Qt methods?

equalsraf:
I think they still make sense, to make it clear that there is an expected encoding for the Neovim api.

I've added a commit to make them inline functions.

PR #1039, opened 2022-12-09
Fix CI Breaks

equalsraf:
Basically #1033 with some more changes.

The highlights so far

- had to disable some lints. For some of them another approach could be to ignore files in the auto/ folder
- linuxdeployqt does not run in ubuntu 22.04 this is intentional on their part, to be able to link against older glibc so the build works in more systems. I rolled back the image to 18.
- with the absent qt5-default had to pass the qmake path to linuxdeployqt manually

The remaining failed jobs are test failures.


equalsraf:
I'm unable to troubleshoot the windows test failures in github workflows, the tests in appveyor look ok.

PR #1042, opened 2022-12-18
Save test output in files files

equalsraf:
Remove the CTEST_EXE_ARGS, and instead default to log into two locations

1. stdout using the txt format
2. a file named after the text binary

**Rationale:** one more attempt to reduce the CI pain

Notes:
- Not all cmake versions in CI support all output formats, e.g. some runners lack junit support
- fooling around a bit with other test options to help CI
- mingw test failures are some kind of exception (maybe missing dll) Exit code 0xc0000139

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -148,6 +148,9 @@ jobs:
           run: ctest -VV --output-on-failure -C ${{ matrix.flavor }}
       - name: Show windows logs
         if: ${{ failure() && startsWith(matrix.runner, 'windows') }}
+        run: Get-EventLog -List
+      - name: Show windows logs
+        if: ${{ failure() && startsWith(matrix.runner, 'windows') }}
         run: Get-EventLog -LogName Application -ET ERROR
       - uses: actions/upload-artifact@v3
         name: Publish test results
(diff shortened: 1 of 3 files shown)


jgehrig:
The CI test output is a strange one... I couldn't get stdout to work no matter what I tried, even text files and piping the output. Even the uploaded text file logs were empty if I remember correctly.

Locally running the executable worked just fine, so it is some specific interaction between GitHub Actions + ctest.

My vote: I think `CTEST_EXE_ARGS` is a pretty simple and reliable workaround. Simply adding the `-tap` argument causes the output to reliably display. The argument is through ctest, so it should be supported everyone.

With that said, if you can find a simpler or more reliable solution I'm all for it! :)

> mingw test failures are some kind of exception (maybe missing dll) Exit code 0xc0000139

Yes, I think that is the Windows error code for a DLL that is not found at runtime.

I think we need to modify the PATH to include for missing DLLs. I think we need the Qt + MinGW compiler bin/lib directories.

equalsraf:
> My vote: I think CTEST_EXE_ARGS is a pretty simple and reliable workaround. Simply adding the -tap argument causes the output to reliably display. The argument is through ctest, so it should be supported everyone.

yeah my original intent was not to remove it. i was just trying to fiddle with the call to ctest and the qt test binaries to get them working, but alas no luck. Seems like for now github actions is dead for msvc builds.

Issue #999, opened 2022-07-07
<S-BS> and <S-Space> no longer reporting "Shift"


ltratt:
On 7b230767dd04d921ca5c2c31ad46ffd00111aed2:

```
$ cat /tmp/conf.lua
vim.api.nvim_set_keymap('i', '<BS>', 'hello', {noremap=true})
vim.api.nvim_set_keymap('i', '<S-BS>', 'goodbye', {noremap=true})
$ nvim-qt -- -u /tmp/conf.lua
```

Pressing either "backspace" or "shift-backspace" in insert mode inserts the text "hello" but never the text "goodbye". These key bindings worked as expected on neovim-qt-0.2.16.1.

equalsraf:
Based on bisection, it seems to be due to 4352846be7f7f42738c568eafaba965a20bb7a45 https://github.com/equalsraf/neovim-qt/issues/259 - where the shift modifier was removed from space/backspace because it was causing another issue.

ltratt:
Ah so it seems that in `:terminal` mode Shift-Backspace was a problem but, AFAICS, no-one had an issue with it in non-`:terminal` mode? Does that mean the `input.cpp` needs a check along the lines of `if (api2->nvim_get_mode() == "terminal")` around the `if (key == Qt::Key_Backspace)` bit?

jgehrig:
@ltratt 

I think the key sequence we sent in previous versions was a bug. As I understand, `<S-Backspace>` should not be sent by the input layer.

Do your `<S-BS>` mappings work in `nvim`?

From a quick test with my machine, `nvim` also triggers a `<BS>` event when Shift + Backspace keys are pressed.

ltratt:
Yes, <S-BS> works in plain nvim (assuming the terminal supports it: xterm supports it but xfce4-terminal doesn't, for example).

ltratt:
Oops I forgot to explicitly mention you @jgehrig

jgehrig:
@ltratt Thanks for the reminder ping!

I'm a little hesitant to add the `:terminal` mode check to the input layer. I've tried to keep the input layer stateless, so it is easier to validate and test.

I will do some investigation and report back.

jgehrig:
@ltratt 

I don't think there is a practical way for us to fix this... More details added to the upstream Neovim bug.

The check you suggested requires and RPC request/response before every key event. This would be a big performance hit.

[...]

Issue #728, opened 2020-07-29
shift+space prints ;2u in terminal


fuelen:
```
$ nvim-qt --version
NVIM-QT v0.2.16.1
Build type: Release
Compilation:-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -Wall -Wextra -Wno-unused-parameter -std=c++11 -Wfatal-errors
Qt Version: 5.15.0
Environment: 
  nvim: nvim
  args: --cmd let &rtp.=',/usr/share/nvim-qt/runtime' --cmd set termguicolors
  runtime: /usr/share/nvim-qt/runtime

NVIM v0.4.3
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/build/neovim/src/build/config -I/build/neovim/src/neovim-0.4.3/src -I/usr/include -I/build/neovim/src/build/src/nvim/auto -I/build/neovim/src/build/include
Скомпілював  builduser

Характеристики: +acl +iconv +tui
See ":help feature-compile"

   системний vimrc: "$VIM/sysinit.vim"
  заміна для $VIM: "/usr/share/nvim"

Run :checkhealth for more info

```
steps to reproduce:
open neovim-qt
run :terminal
go to insert mode
press Shift+Space

It is annoying, because while writing `iex -S mix` I always get `iex -S;2umix` and I do not experience this in other terminal emulators :)

jgehrig:
Sorry for the troubles!

I cannot reproduce, but I would guess there is an environment difference.

A few questions:

1. What platform are you using? Windows, Linux, or MacOS?
2. Does the issue still reproduce without your `vimrc`, use `nvim-qt -- -u NONE`
3. Are you using an non-english keyboard layout/localization?
4. Did this start with `v0.2.16`?
5. Can you easily build `neovim-qt` from source?

[...]

PR #1049, opened 2023-02-16
Issue 999: Replace Issue 728 with key remap

jgehrig:
Many users find the shift-space and shift-backspace behavior in terminal frustrating. However, some users still want to map these sequences.

Applying the runtime re-map workaround, since Neovim is unable to fix this issue upstream. Neovim still does not report `:terminal` mode.

Related: #728 #999

### DIFF
--- a/src/gui/input.cpp
+++ b/src/gui/input.cpp
@@ -191,13 +191,6 @@ QString convertKey(const QKeyEvent& ev) noexcept
 	const QMap<int, QString>& specialKeys { GetSpecialKeysMap() };
 
 	if (specialKeys.contains(key)) {
-		// Issue#728: Shift + Space inserts ;2u in `:terminal`. Incorrectly sent as <S-Space>.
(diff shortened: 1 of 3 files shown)


ltratt:
This works well for me -- thanks for doing this @jgehrig!

I did encounter a minor build error since due to this line (I assume it's accidentally left in) causing a linking error:

```
void ShiftSpaceWellFormed() noexcept;
```

Once I removed that, everything built, and I can confirm that Shift-Backspace works in my config!

jgehrig:
@ltratt Excellent! :+1: 

Yes -- I missed that on check-in.

Thanks for the feedback and sorry for the delay in making these changes!

I had hoped Neovim would address this issue, but I don't that is likely. This fix is hacky, but it appears to be the only solution that works for everyone.

ltratt:
Thanks again -- I'm grateful to you for doing this, as I know it's not your preferred option (for reasons that I totally understand)!

Issue #1063, opened 2023-04-05
Cant disable cursor styling


merpath:
Platform:
Endeavour OS Linux (Arch based)

I have a line on my configuration file to disable cursor styling:

```
vim.o.guicursor = nil
-- or
vim.o.guicursor = ""
-- either way works on regular Neovim on terminal
```

But it doesn't get sourced when I open nvim-qt. The other configuration option works, so I think it may be a problem on nvim-qt's guicursor implementation. Anyone having the same problem?

jgehrig:
We should support the `guicursor option`, however I'm not familiar with setting it via `vim.o.guicursor`.

I'm only familiar with configuring the cursor via `set guicursor=...` in `init.vim` or `vimrc`.

Can you provide some additional details so I can reproduce (and hopefully fix) the issue?

What do you expect to happen, and what happens instead?

---

You could also try setting the option to a style-less default, such as:
`guicursor=n-v-c-sm:block,i-ci-ve:ver25,r-cr-o:hor20`

manungsa:
These configuration line is basically the same as
```vim
set guicursor=
```
So I expect the cursor to go full block style, but after restarting nvim-qt the cursor still change according to the mode where I'm in, e.g. in Insert mode the cursor changed to thin style, on Replace  mode the cursor go underscore style.

jgehrig:
Okay, I see what is going on here. We do support cursor styling and the lack thereof.

However, when no cursor styling is applied, we use some defaults:
https://github.com/equalsraf/neovim-qt/blob/b4ca557cd133eb049cf4c977a81437a23a1fcd06/src/gui/shell.cpp#L695-L714

To get the behavior you're looking for, we would need to set the cursor into block mode whenever styling is disabled.

I don't think what we're doing is strictly wrong, but there is a behavior difference from the TUI.

Generally, we try to match the TUI behavior when possible.

---

In the meantime, you could explicitly specify how you want the cursor to behave:

`guicursor=n-v-c-sm-i-ci-ve-r-cr-o:block`

manungsa:
The `guicursor=n-v-c-sm-i-ci-ve-r-cr-o:block` worked tbh. I will use it for now. Thanks!.

PR #1067, opened 2023-05-01
Issue 1063: change default guicursor behavior

jgehrig:
**Issue #1063:** can't disable cursor styling

Currently, when no guicursor value is provided (styling is disabled), we use some default values: block, ver25, or hor25 based on mode.

The TUI does not behave the same way, it uses block mode for all cases when guicursor is empty.

This change removes the cursor defaults, and applies block mode to all cases where guicursor is empty.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -588,6 +588,8 @@ void Shell::handleRedraw(const QByteArray& name, const QVariantList& opargs)
 	} else if (name == "popupmenu_hide") {
 		m_pum.hide();
 	} else if (name == "mode_info_set") {
+		qDebug() << "mode_info_set";
+		qDebug() << "  " << opargs;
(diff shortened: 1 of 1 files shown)


jgehrig:
I have some slight hesitancy on check-in for this fix, some users could complain of the exact opposite.

However, the TUI behaves as called out in the bug, so perhaps this will not be an issue?

justinmk:
> The TUI does not behave the same way, it uses block mode for all cases when guicursor is empty.

I think that also matches Vim. If the OS provides an API for "cursor style preferences", that would be ideal. Otherwise, yes, if the user explicitly unsets 'guicursor' they are asking for "dumb mode".

equalsraf:
Merged in https://github.com/equalsraf/neovim-qt/pull/1197

PR #1074, opened 2023-06-30
WIP: Windows CI fixes

equalsraf:
Windows CI (except Azure) has been failing for a while now. This manifests as tests getting stuck and timing out. Strangely the tests suites that fail seem to print the results for all tests, suggesting there is something wrong with the teardown of the tests.

This proved difficult to troubleshoot since I was never able to reproduce it locally. But it happens almost every time for two of the test suites in appveyor, and occasionally on a third one.

The only hint I was able to observe was by opening an RDP connection to appveyor and observing the state of the system when the tests are stuck. ctest execution continues and a number of nvim child processes are still running.

---

Hypothesis: this seems to be related to the life cycle of the QProcess instances - where a leftover QProcess prevents the tests from terminating.

The only supporting evidence I have are the (still running) nvim instances and the output that suggests test execution is complete. I tried to validate this by fiddling with the test code by adding manual calls to deleteLater for the NeovimConnector (which wraps QProcess) in the hopes of terminating the child processes.

This actually seemed to fix two of the three test suites. However the last one is still failing with stack trace (meaning I broke it :D).

- [x] I think there is a bug with our use of neovimconnector in that the instance parent is not set when it should - I have included one commit for this.
- [ ] failing to delete neovimconnector seems to lead to a situation where a test process does not terminate, presumably because the connector (i.e. QProcess) is not terminated

The root cause is still unknown. The qt docs point out that the QProcess destructor could block until it kills the process - but I have no  reason to believe that is the cause, and a manual call to deleteLater() works.

### DIFF
--- a/contrib/appveyor.yml
+++ b/contrib/appveyor.yml
@@ -10,16 +10,18 @@ environment:
     GENERATOR: MinGW Makefiles
     RELEASE_ARTIFACT: true
     CMAKE_PATH_PREFIX: C:\Qt\5.15\mingw81_64\lib\cmake
+    CMAKE_BUILD_TYPE: Release
   - PlatformToolset: VisualStudio2019
     QTPATH: C:\Qt\5.15\msvc2019_64
     GENERATOR: Visual Studio 16
     CMAKE_PATH_PREFIX: C:\Qt\5.15\msvc2019_64\lib\cmake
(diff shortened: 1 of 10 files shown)


equalsraf:
Replaced with https://github.com/equalsraf/neovim-qt/pull/1076

Issue #624, opened 2020-01-20
DPI/font size changes lead to lots of space between characters


youngmit:
I am running into a similar issue as in #391, although a bit stranger and more steps to reproduce. I have a laptop with a high-DPI screen, which is usually plugged into a docking station with two standard-resolution displays. When un-docking, the font gets all squished together; the characters are not too small, but the boxes that they occupy appear to be too small to contain them and they overlap each other like this:
![image](https://user-images.githubusercontent.com/1179265/72762620-254a1780-3b95-11ea-8e0d-1e3bfe46cf63.png)

Setting the font with something like `:GuiFont Consolas:h9` brings everything back to normal. However, when I re-dock with the two external monitors, all of the text gets spread out, where the glyphs are all the correct size, but the boxes that they occupy are too large, like this:
![image](https://user-images.githubusercontent.com/1179265/72762724-88d44500-3b95-11ea-8c20-462807ae84cd.png)

Any ideas?

equalsraf:
One thing we have not tried on this is a [DPI awareness flag](https://doc.qt.io/qt-5/highdpi.html) for windows.

From the docs (I assume the intended value is 0 for DPI Unaware but we should test with 1 and 2 as well)

```
nvim-qt -platform windows:dpiawareness=0
```

PR #1080, opened 2023-08-11
Fix setGuiFont force flag

RyoSusami:
Fixed an issue where fonts were not displayed properly in a per-monitor dpi environment on Win.

Before
image

After
image

Maybe #624 will also be resolved.

Moving windows between monitors still causes problems, but at least this fix makes them look right on startup.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -185,8 +185,7 @@ void Shell::screenChanged()
 {
 	// When the screen changes due to dpi scaling we have to
 	// re-set the current font
-	auto r = setGuiFont(fontDesc(), true, true);
-	qWarning() << __func__ << r;
+	setGuiFont(fontDesc(), true, true);
 }
 
 
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -87,6 +87,16 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 		setGuiFont(fontDescription, true /*force*/);
 	}
 
+#if (QT_VERSION > QT_VERSION_CHECK(5, 13, 0))
+	auto screen = this->screen();
+	if (screen) {
+		connect(screen, &QScreen::logicalDotsPerInchChanged,
+				this, &Shell::screenChanged);
+		connect(screen, &QScreen::physicalDotsPerInchChanged,
+				this, &Shell::screenChanged);
+	}
+#endif
+
 	if (!m_nvim) {
 		qWarning() << "Received NULL as Neovim Connector";
 		return;
@@ -167,6 +177,15 @@ bool Shell::setGuiFont(const QString& fdesc, bool force) noexcept
 
 	return true;
 }
+
+void Shell::screenChanged()
+{
+	// When the screen changes due to dpi scaling we have to re-set the current font
+	auto r = setShellFont(font(), true);
+	qWarning() << __func__ << r;
+	resizeNeovim(size());
+}
+
 
 bool Shell::setGuiFontWide(const QString& fdesc) noexcept
 {
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -181,12 +181,6 @@ bool Shell::setGuiFont(const QString& fdesc, bool force, bool reset) noexcept
 	return true;
 }
 
-void Shell::connectScreenChanged()
-{
-	screenChanged();
-	connect(window()->windowHandle(), &QWindow::screenChanged, this, &Shell::screenChanged);
-}
-
 void Shell::screenChanged()
 {
 	// When the screen changes due to dpi scaling we have to
@@ -1349,6 +1343,9 @@ void Shell::showEvent(QShowEvent* ev)
 	if (m_nvim->isReady()) {
 		init();
 	}
+
+	screenChanged();
+	connect(window()->windowHandle(), &QWindow::screenChanged, this, &Shell::screenChanged);
 }
 
 void Shell::paintEvent(QPaintEvent *ev)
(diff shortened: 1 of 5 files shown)


jgehrig:
@RyoSusami 

Thanks for you contribution!

I wouldn't expect the force `:GuiFont!` flag or the exit-early font comparison to have an impact on multi-monitor dpi scenarios. For my understanding, can you provide some details on how this fix resolves the issues above?

I'd like to understand:
1. What scenario are you fixing here. Provide an example setup with steps.
2. What about the font-description comparison causes issues for the scenario above.
3. What impact does the force `:GuiFont!` flag have?

Once I understand the issue and fix, I look forward to merging this PR! :)

src/gui/shell.cpp
```diff
@@ -118,7 +118,7 @@ void Shell::handleFontError(const QString& msg)
 bool Shell::setGuiFont(const QString& fdesc, bool force) noexcept
 {
 	// Exit early if the font description has not changed
-	if (fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0) {
+	if (!force && (fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0)) {
```
jgehrig, line 121: nitpick: no need for the extra parenthesis here

```suggestion
	if (!force && fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0) {
```

RyoSusami:
Thank you for considering this PR.

Normally, if I write one font setting in ginit.vim as follows and start it, the display will be strange as "Before".
`Guifont! HackGen35:h11`
After this, even if I specify the font again from the command line as follows, nothing will change.
`:Guifont! HackGen35:h11`

However, if I change the height or some font select once from command line, it will be drawn normally like "After".

So, as a workaround, there is also a method of writing two font settings in ginit.vim.
`Guifont! HackGen35:h10`
`Guifont! HackGen35:h11`
By forcibly adding a change, it becomes a normal display like "After" at startup.

This fix still doesn't work without the "!".
I feel like the exit-early font comparison itself should be removed for multi-monitor dpi.(But I can't decide if I should do that.)

I'm not a native English speaker, sorry if my explanation is hard to understand.
I've had this problem myself for a few years, and I've seen some people with the same problem, so it would be nice to have this problem resolved.

equalsraf:
My understanding is that in some hidpi scenarios the fonts will look weird after a display change (resolution change, not sure which).

When this happens users found out the problem gets "fixed" if they change the font via GuiFont. But it is annoying to do it twice because of the if statement.

```c++
if (fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0) {
```

equalsraf:
I don't have reliable way to reproduce the problem though.

@RyoSusami  do you know when the fonts start looking strange? Does it happen after moving the window between screens, or at startup? or something else.

RyoSusami:
@equalsraf

Thank you for your consideration of this issue.
I had a shallow understanding of the current code and thought it would be difficult to respond correctly.

As far as I know, when I move between screens and at startup, as you point.

RyoSusami:
@equalsraf 
Pushed the fix in #1081.
The connection with screenChanged has been moved to showEvent().
Thank you!

src/gui/shell.cpp
```diff
@@ -87,6 +87,16 @@ Shell::Shell(NeovimConnector *nvim, QWidget *parent)
 		setGuiFont(fontDescription, true /*force*/);
 	}
 
+#if (QT_VERSION > QT_VERSION_CHECK(5, 13, 0))
+	auto screen = this->screen();
+	if (screen) {
+		connect(screen, &QScreen::logicalDotsPerInchChanged,
```
equalsraf, line 93: Do we still need these two here now that we have the one in showEvent?

src/gui/shell.cpp
```diff
@@ -168,6 +181,15 @@ bool Shell::setGuiFont(const QString& fdesc, bool force) noexcept
 	return true;
 }
 
+void Shell::screenChanged()
+{
+	// When the screen changes due to dpi scaling we have to
+	// re-set the current font
+	auto r = setGuiFont(fontDesc(), true, true);
+	qWarning() << __func__ << r;
```
equalsraf, line 189: remove qwarning

src/gui/shell.cpp
```diff
@@ -1321,6 +1343,9 @@ void Shell::showEvent(QShowEvent* ev)
 	if (m_nvim->isReady()) {
 		init();
 	}
+
+	screenChanged();
+	connect(window()->windowHandle(), &QWindow::screenChanged, this, &Shell::screenChanged);
```
equalsraf, line 1348: Because showEvent can happen for other reasons besides show() we can

- call connect with Qt::UniqueConnection as third argument
- wrap this in a check for QShowEvent::spontaneous().

Just being pedantic here, but check window() for NULL.

src/gui/shell.cpp
RyoSusami: It fires only on the primary monitor and the font is not adjusted well, so I think it's good to delete it once.

src/gui/shell.cpp
RyoSusami: Deleted qWarning & r.

src/gui/shell.cpp
RyoSusami: Added NULL check for window.

RyoSusami:
Shell::screenChanged() receiver has the same name as QWindow::screenChanged() signal, so I think it would be better to change it to a receiver-like name.
I tried to point it out, but this is my first time using Qt, so I don't know the naming trends in Qt. If it's what you intended, I think it's fine as is.

equalsraf:
I have manually merged the changes into the master branch.

@RyoSusami thanks for this, well done on spotting the issue and the fix.

equalsraf:
One final fix in #1082

RyoSusami:
@equalsraf 
Thank you for considering with me. This was a good experience for me.

“Otsukaresama desita!”

PR #1081, opened 2023-08-12
Investigating #1080

equalsraf:
ref #1080

I was unable to get QWindow in my system, so I settled for QScreen logical dpi signal for now. This API is only available in Qt 5.14.

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -114,11 +114,12 @@ void Shell::handleFontError(const QString& msg)
 ///
 /// @param fdesc Neovim font description string, "Fira Code:h11".
 /// @param force used to indicate :GuiFont!, overrides mono space checks.
+/// @param reset we reseting the font, ignore optimizations
 /// @returns `true` if the font was successfully set.
-bool Shell::setGuiFont(const QString& fdesc, bool force) noexcept
+bool Shell::setGuiFont(const QString& fdesc, bool force, bool reset) noexcept
 {
 	// Exit early if the font description has not changed
-	if (fdesc.compare(fontDesc(), Qt::CaseInsensitive) == 0) {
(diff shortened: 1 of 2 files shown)


equalsraf:
Another way to trigger is by changing the text scaling option in the windows Display settings. e.g. from 100% to 125.

equalsraf:
I confirmed the signal seems to be working (for a change in text scaling), but the call to setCellSize does not have the intended effect.

RyoSusami:
@equalsraf Thank you for experimenting.
I was trying something that I could do myself, so I will report what I did.

I was checking the signal of logicalDotsPerInchChanged, but it seemed to respond only when the scale rate setting of the primary monitor was changed (just only primary monitor).

I tried adding the following code to detect that the window moved between monitors, but it did not respond.
`connect(this->window()->windowHandle(), &QWindow::screenChanged, this, &Shell::screenChanged);`

equalsraf:
Thanks. I'm out of time today but i'll have a look later. I see some examples linked in the ticket too.

In my case I was testing with a single screen, just be switching scalling but even then it was not working well. It did get triggered, but even setting the font was not enough.

RyoSusami:
@equalsraf 

I now have working code derived from your tb-screen-change and pushed it to my repository.

The point is
- screenChanged had to be connected after window show().
- ScreenChanged only responds when windows are moved between monitors, so it was necessary to forcefully call Shell::screenChanged() at startup.

Should I make a another PR?

equalsraf:
Fantastic, great job.

Feel free to open a new pull request or just push on top of the other one (#1080).

If screenChanged has to be connected after window is show() maybe we should be doing it after receiving a showEvent() but i'm ok with the current approach too.

PR #1082, opened 2023-08-17
Disconnect old QSCreen on show event

equalsraf:
One additional fix to the earlier merge on the DPI issue.

I wont be able to test this for a few days, so if anyone can find a windows machine and confirm that this did not change the intended behaviour that would be great/


RyoSusami:
I tried. It seems to be no change in behavior with this fix.

equalsraf:
Thanks.

PR #1093, opened 2024-02-06
cmake: set the minimum version to 3.9.0

davvid:
Newer versions of cmake warn that support for cmake < 3.5 will be removed.

cmake 3.9.0 was released in 2017 and most distros have much newer versions available at this point.

Bump the minimum required version to 3.9.0 to silence deprecation warnings.


equalsraf:
Merged, thanks @davvid

Issue #997, opened 2022-07-03
Calculate window size by lines and columns (feature request)


damanis:
The QT's saveGeometry() works properly, but it saves both dimensions and position. The result restoreGeomentry() always opens window at same position.
The option to calculate dimensions by lines and columns, configured in nvim, and give to Window Manager to set window position should be useful.

damanis:
Seems, --geometry can be used, but still need the option to disable save/restore geometry.

jgehrig:
It is not clear to me, what is the ask or feature request here?

Are you asking for a `--start-position` argument?

The `--geometry` option allows the user to specify the windows size. That can be used to indirectly set the number of lines/columns. Why would you need to set lines/columns explicitly?

Adding a lines/coluns option seems like additional complexity without much benefit.

Currently, there is no way to disable `saveGeometry`. That does seem like a good feature to add...

damanis:
@jgehrig 
It is feature request.
I didn't asked ```--start-position``` option, but vise versa, I requested a parameter to make save/load geometry optional.

The neovim-qt uses ```--geometry``` parameter to set windows size. But loadGeometry also places the window on saved position - as a result all neovim-qt windows open one on top of the other.
In my opinion, window position should be defined by window manager, not by application itself (or be optionally).

P.S. Vim and neovim provide way to set lines and columns in init file. Classic gvim uses them to calculate window size; ```--geometry``` is not needed. Seems, neovim GUI can get these parameters from neovim during start.

damanis:
@jgehrig 
This small patch prevent save and restore main window geometry via nvim-qt.conf. Could you add it to neovim-qt?
```
[General]
save_window_position=false
```
```
diff --git a/src/gui/mainwindow.cpp b/src/gui/mainwindow.cpp
index a9a7178..4618978 100644
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -299,7 +299,10 @@ Shell* MainWindow::shell()
 void MainWindow::saveWindowGeometry()
 {
 	QSettings settings{ "window-geometry" };
-	settings.setValue("window_geometry", saveGeometry());
+	QSettings winpos;
+
+        if (winpos.value("save_window_position").toBool())
+		settings.setValue("window_geometry", saveGeometry());
 	settings.setValue("window_state", saveState());
 }
 
@@ -310,7 +313,9 @@ void MainWindow::restoreWindowGeometry()
[...]

PR #1096, opened 2024-02-11
mainwindow: add an option to avoid restoring window geometry

davvid:
Avoid restoring window geometry when `restore_window_geometry=false` in `~/.config/nvim-qt/window-geometry.conf`. Defaults to `true`.

Suggested-by: @damanis on github
Related-to: #997
Related-to: #1094

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -312,7 +312,7 @@ Shell* MainWindow::shell()
 void MainWindow::saveWindowGeometry()
 {
 	QSettings settings("nvim-qt", "window-geometry");
-	bool restore_window_geometry = settings.value("restore_window_geometry", true).toBool();
+	const bool restore_window_geometry{ settings.value("restore_window_geometry", true).toBool() };
 	if (!restore_window_geometry) {
 		return;
 	}
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -312,6 +312,11 @@ Shell* MainWindow::shell()
 void MainWindow::saveWindowGeometry()
 {
 	QSettings settings("nvim-qt", "window-geometry");
+	bool restore_window_geometry = settings.value("restore_window_geometry", true).toBool();
+	if (!restore_window_geometry) {
+		return;
+	}
+	settings.setValue("restore_window_geometry", restore_window_geometry);
 	settings.setValue("window_geometry", saveGeometry());
 	settings.setValue("window_state", saveState());
 }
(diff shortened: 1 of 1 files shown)


src/gui/mainwindow.cpp
```diff
@@ -312,6 +312,8 @@ Shell* MainWindow::shell()
 void MainWindow::saveWindowGeometry()
 {
 	QSettings settings("nvim-qt", "window-geometry");
+	bool restore_window_geometry = settings.value("restore_window_geometry", true).toBool();
+	settings.setValue("restore_window_geometry", restore_window_geometry);
 	settings.setValue("window_geometry", saveGeometry());
 	settings.setValue("window_state", saveState());
```
jgehrig, line 318: I think we should avoid writing `window_geometry` and `window_state`  values if `restore_window_geometry` is set to false.

I've seen some users who track their config files in VCS complain about the constant changing of these values every open/close. I'll try to find the specific issue and reference it in the main discussion.

src/gui/mainwindow.cpp
```diff
@@ -325,7 +327,9 @@ void MainWindow::restoreWindowGeometry()
 #endif
 
 	QSettings settings("nvim-qt", "window-geometry");
-	restoreGeometry(settings.value("window_geometry").toByteArray());
+	if (settings.value("restore_window_geometry", true).toBool()) {
+		restoreGeometry(settings.value("window_geometry").toByteArray());
+	}
 	restoreState(settings.value("window_state").toByteArray());
```
jgehrig, line 333: I think we should also avoid running `restoreState` in this case too.

Something like:
```C++
	if (!settings.value("restore_window_geometry", true).toBool()) {
		return;
	}
	
	restoreGeometry(settings.value("window_geometry").toByteArray());
	restoreState(settings.value("window_state").toByteArray());
```

jgehrig:
Great idea!

Waiting for some changes/comments from #1100, but we will get this part of the feature merged.

I've added some minor comments. I think we want to optimize out some QSettings reads/writes and tweak a few things.

davvid:
I just rebased and applied the sugs above. This should be good to go now. Thanks for the review.

src/gui/mainwindow.cpp
```diff
@@ -312,6 +312,11 @@ Shell* MainWindow::shell()
 void MainWindow::saveWindowGeometry()
 {
 	QSettings settings("nvim-qt", "window-geometry");
+	bool restore_window_geometry = settings.value("restore_window_geometry", true).toBool();
```
jgehrig, line 315:
```suggestion
	const bool restore_window_geometry{ settings.value("restore_window_geometry", true).toBool() };
```

jgehrig:
Merged! Thanks for the contribution :+1:

PR #1099, opened 2024-02-28
cmake: remove redundant slashes when installing files

davvid:
Using a trailing slash results in install commands that contain two slashes between the base directory and the path being installed.


jgehrig:
Nice catch! Thanks for the contribution

Issue #679, opened 2020-04-13
Don't save window position on wayland


WhyNotHugo:
I noticed that `neovim-qt` keeps creating a config file every time I run it, in `.config/nvim-qt/window-geometry.conf`. I found this curious because I didn't configure anything; all configuration is done via `.config/nvim/config`.

I looked at it and noticed it only keeps the window size, position, etc.

Is there a way to configure it not to do this? It just annoys me that this file keeps popping up (I git-track my settings, and this keeps showing up as a new file).

Moreover, on Wayland, the client (application) has no control over window placement. That is: the compositor completely controls where the windows is positioned. So on Wayland, I see little use for the app even storing this. Could you skip saving it entirely by default on Wayland?

jgehrig:
> I noticed that neovim-qt keeps creating a config file every time I run it, in .config/nvim-qt/window-geometry.conf. I found this curious because I didn't configure anything; all configuration is done via .config/nvim/config.

Sorry for the annoyance. There isn't a way to disable those config files.

Programs creating files within `.config` is common. For example, `google-chrome` caches 1000's of files into this directory. How do you usually handle this scenario?

The storage mechanism for window size is `QSettings`. You can find the code responsible for this behavior in Pull #551.

I think this bug should be "Won't Fix", the behavior is by design.

> Moreover, on Wayland, the client (application) has no control over window placement. That is: the compositor completely controls where the windows is positioned. So on Wayland, I see little use for the app even storing this. Could you skip saving it entirely by default on Wayland?

Several platforms are supported by `neovim-qt` Wayland is the odd-man-out in this scenario. We should avoid writing platform specific logic if possible. It quickly makes testing hard.

Wayland must be give applications control of their size, right? If so, at least some of the information in this config file is still necessary.

WhyNotHugo:
> There isn't a way to disable those config files.

Would it be considerable to add one? I really dislike neovim-qt creating a second config file, just to store information that (a) it cannot use on my platform (b) I would prefer be discarded anyway.

> Programs creating files within .config is common. For example, google-chrome caches 1000's of files into this directory. How do you usually handle this scenario?

I don't use `google-chrome`. Its Linux support is very poor, and doesn't even really run properly on Wayland.

> the behavior is by design.

It does not work though; the window position is save and never used for anything (since you can't control where new windows are created anyway).

> Wayland must be give applications control of their size, right? If so, at least some of the information in this config file is still necessary.

[...]

PR #1100, opened 2024-02-28
Set a default size on first launch

davvid:
When `~/.config/nvim-qt/window-geometry.conf` does not exist then
restoreGeometry() will initialize a window that is tiny and must be
resized in order to be usable in some environments.
    
Resize the window to half the screen when launched for the first time.
Do not restore window state when it is invalid or missing.

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -1,8 +1,10 @@
 #include "mainwindow.h"
 
+#include <QApplication>
 #include <QCloseEvent>
 #include <QEventLoop>
 #include <QLayout>
+#include <QScreen>
 #include <QSettings>
 #include <QStyleFactory>
 #include <QToolBar>
@@ -12,6 +14,23 @@ namespace NeovimQt {
 static QString DefaultWindowTitle() noexcept
 {
 	return "Neovim";
+}
+
+static void SetDefaultWindowSize(QWidget& widget) noexcept
+{
+	const QPoint local_position{ widget.width() / 2, 0 };
+	const QPoint global_position{ widget.mapToGlobal(local_position) };
+	const QScreen* screen{ qApp->screenAt(global_position) };
+	if (!screen) {
+		return;
+	}
+	const QRect geometry{ screen->availableGeometry() };
+	if (screen->orientation() == Qt::LandscapeOrientation) {
+		widget.resize(geometry.width() / 2, geometry.height());
+	}
+	else {
+		widget.resize(geometry.width(), geometry.height() / 2);
+	}
 }
 
 MainWindow::MainWindow(NeovimConnector* c, QWidget* parent) noexcept
@@ -333,8 +352,17 @@ void MainWindow::restoreWindowGeometry()
 	if (!settings.value("restore_window_geometry", true).toBool()) {
 		return;
 	}
-	restoreGeometry(settings.value("window_geometry").toByteArray());
-	restoreState(settings.value("window_state").toByteArray());
+	const QVariant geometry{ settings.value("window_geometry") };
+	if (geometry.isValid()) {
+		restoreGeometry(geometry.toByteArray());
+	}
+	else {
+		SetDefaultWindowSize(*this);
+	}
+	const QVariant state{ settings.value("window_state") };
+	if (state.isValid()) {
+		restoreState(state.toByteArray());
+	}
 }
 
 void MainWindow::setGuiAdaptiveColorEnabled(bool isEnabled)


jgehrig:
This looks closely related to #1096. Let's apply these changes in a single PR over there.

Also the hard-coded size concerns me, it might not be the best size for all platforms.

A few comments:
1. Does the annoying behavior go away when 1096 is applied?
I suspect when the geometry is not explicitly restored, the window manager may set a reasonable default size. This is worth testing and discussing quickly on all major platforms (Windows, Linux, MacOS).
2. We should avoid the hard-coded value of `QSize{ 1280, 1440 }`. Maybe a QSetting can be used?

If we really do need this feature think this feature is necessary, it should likely be an optional setting in `window-geometry`. Maybe something like `default_window_size`?

davvid:
Sounds good. I'll revisit this after working through #1096 since it's touching the same part of the code.

> 1. Does the annoying behavior go away when 1096 is applied?

When I first wrote this patch I was in a temp branch where I had the 1096 changes in effect and it didn't make a difference. I'm going to re-test just to be sure.

> 2. We should avoid the hard-coded value of QSize{ 1280, 1440 }

I can understand that. There's also a part of me that thinks that it's 2024 and the vast majority of users will be just fine with that value, but I agree that we can do better.

This will typically only come into effect on the very first run (e.g. if you `rm -r ~/.config/nvim-qt` you can see this behavior) so a QSetting option to control this doesn't seem that helpful. It's a one-time annoyance, but it's also one of those "try to make a good first impression" kind of polish that's worth improving.

Here's an idea.. we could query the screen size using QScreen https://doc.qt.io/qt-5/qscreen.html (I would suggest QDesktopWidget but it's deprecated and removed in Qt6) and then clamp the initial size to the screen size.

Alternatively, instead of clamping, we could just set the size to be the same as the screen size (effectively making the UI take the full screen) or some portion of it (e.g. half width of the screen).

Let me know what you think. I'll get back to #1096 now and get that into shape for merging.

davvid:
Here's what I landed on.

* When no configuration is present (e.g. on the very first run) then we will apply a default size.
* When configuration is present and users have asked us to *not* restore geometry then we will also apply a default size. I suspect that users that are asking for not restoring geometry do not expect the window to appear tiny on every run, hence this behavior.
* We query for the screen size to determine the default size. For screens in a Landscape orientation we will use half the width of the screeen and full height. For screens in Portrait mode we will use the full width of the screen and half height.

I'm testing on Gnome/Wayland. I suspect that the "tiny initial screen size" issue is seen on other platforms as well, so this should be sensible behavior on all platforms.

Most users will not change defaults. What they'll get is the initial size that we're setting here and then whatever size and position they set manually will be preserved, so there's probably few users who won't want the default size.

If users don't like the default size they can always use the (enabled-by-default) behavior where we restore window geometry.

I hope we don't need another QSettings entry because the `window_geometry` field already serves that purpose.

Let me know what you think. Hopefully this is a nice approach.

davvid:
This PR extends #1096 so I'll rebase this after that topic is resolved. Only a single commit will remain in this branch post-rebase.

src/gui/mainwindow.cpp
```diff
@@ -14,6 +16,19 @@ static QString DefaultWindowTitle() noexcept
 	return "Neovim";
 }
 
+static void SetDefaultWindowSize(QWidget* widget)
```
jgehrig, line 19: Add `noexcept` to the end.

jgehrig: Is this const-correct? I think `widget` can/should be made `const`.

src/gui/mainwindow.cpp
```diff
@@ -14,6 +16,19 @@ static QString DefaultWindowTitle() noexcept
...
+{
+	QPoint widget_position = widget->mapToGlobal(QPoint{ widget->width() / 2, 0 });
+	QScreen* screen = qApp->screenAt(widget_position);
+	QRect geometry{ screen->availableGeometry() };
```
jgehrig, line 23: Needs a nullptr check, since `screenAt` can return null.

> Returns the screen at point, or nullptr if outside of any screen.

https://doc.qt.io/qt-6/qguiapplication.html#screenAt

src/gui/mainwindow.cpp
```diff
@@ -14,6 +16,19 @@ static QString DefaultWindowTitle() noexcept
 	return "Neovim";
 }
 
+static void SetDefaultWindowSize(QWidget* widget)
+{
+	QPoint widget_position = widget->mapToGlobal(QPoint{ widget->width() / 2, 0 });
+	QScreen* screen = qApp->screenAt(widget_position);
```
jgehrig, line 22: Use uniform init, `{}`.

Sorry for the nitpick :)

src/gui/mainwindow.cpp
```diff
@@ -331,10 +346,20 @@ void MainWindow::restoreWindowGeometry()
 
 	QSettings settings("nvim-qt", "window-geometry");
 	if (!settings.value("restore_window_geometry", true).toBool()) {
+		SetDefaultWindowSize(this);
 		return;
 	}
-	restoreGeometry(settings.value("window_geometry").toByteArray());
-	restoreState(settings.value("window_state").toByteArray());
+	QVariant geometry{ settings.value("window_geometry") };
```
jgehrig, line 352: `const QVariant`?

jgehrig:
Everything looks good! A few small comments added. Sorry to be nit-picky :)

I like the high-level approach and this looks Qt6 compatible (if I'm reading the Qt docs correctly).

However, I think we should make one behavior change regarding this:

> When configuration is present and users have asked us to not restore geometry then we will also apply a default size. I suspect that users that are asking for not restoring geometry do not expect the window to appear tiny on every run, hence this behavior.

I suspect the users who want the geometry restore off, will also NOT want us poking the window size. Let's make it so when `restore_window_geometry == false` we don't write QSettings values or modify the window size on startup. 

Some window managers provide a reasonable default size or do their own size restoration. Also,tiling window managers force a window size. Here is an example Issue: #679

davvid:
Thanks for the explanation and review ~ not nitpicky at all. I've consted this up and re-pushed.

While we can't make `SetDefaultSize()` take `const QWidget&` (`QWidget::resize()` is non-const), there's no need for a separate function now that we'll only be using it in a single place. I've inlined the guts of `SetDefaultSize()` directly into `restoreWindowGeometry()` instead.

src/gui/mainwindow.cpp
```diff
@@ -329,12 +331,32 @@ void MainWindow::restoreWindowGeometry()
 	qRegisterMetaTypeStreamOperators<QList<int>>("QList<int>");
 #endif
 
-	QSettings settings("nvim-qt", "window-geometry");
+	const QSettings settings("nvim-qt", "window-geometry");
```
jgehrig, line 334: nitpick: No need to change this line

src/gui/mainwindow.cpp
```diff
@@ -329,12 +331,32 @@ void MainWindow::restoreWindowGeometry()
...
+		const QScreen* screen{ qApp->screenAt(global_position) };
+		if (screen) {
+			const QRect geometry{ screen->availableGeometry() };
+			if (screen->orientation() == Qt::LandscapeOrientation) {
+				resize(geometry.width() / 2, geometry.height());
+			}
+			else {
+				resize(geometry.width(), geometry.height() / 2);
+			}
+		}
```
jgehrig, line 354: Last comment, I promise :)

Let's keep this block as a function, so everything is modular and easy to read.

jgehrig:
@davvid Sorry for the long delay -- I was away from my computer.

Two small comments -- The approach looks great to me, let's get this merged!

davvid:
Good stuff, thanks for the re-review. This should be good to go now.

PR #1102, opened 2024-02-28
workflows: fix macOS CI builds

davvid:
qt5 was renamed to qt@5. Avoid breaking cmake's Qt5 modules by replacing the "brew link qt5" step with a custom PATH on macOS.


jgehrig:
Looks great!

Thanks for figuring this one out :+1:

PR #1104, opened 2024-03-21
Fix azure-pipelines build of neovim-qt on macOS/Qt5

davvid:
The qt@5 package is keg-only these days and not linked into the cmake search path by default.
Set Qt5_DIR so that cmake knows how to find Qt5.

### DIFF
--- a/contrib/azure-pipelines.yml
+++ b/contrib/azure-pipelines.yml
@@ -45,12 +45,15 @@ jobs:
     pool:
       vmImage: 'macOS-latest'
     steps:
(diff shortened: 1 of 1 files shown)


davvid:
Please ignore this PR for now since it'll be noisy while I keep force-pushing to get the azure pipelines fixed. I can't really test this locally so I have to push stuff here to trigger an azure run.

This message will be deleted once it's all passing.

davvid:
I'd recommend prioritizing the merge of this PR and #1105 so that we can rebase the rest of the open PRs to pickup these CI fixes.

PR #1124, opened 2024-08-16
Fix font weight on Qt6

pastalian:
Qt6 has different weight mapping from Qt5. It scales from 1 to 1000.
https://doc.qt.io/qt-6/qfont.html#Weight-enum

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -11,6 +11,14 @@
 #include "compat_shellwidget.h"
 #include "helpers.h"
 
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
+constexpr int c_qtWeightMin{ 0 };
+constexpr int c_qtWeightMax{ 99 };
+#else
+constexpr int c_qtWeightMin{ 1 };
+constexpr int c_qtWeightMax{ 1000 };
+#endif
+
 ShellWidget::ShellWidget(QWidget* parent)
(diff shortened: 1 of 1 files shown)


jgehrig:
Thanks for finding this bug!

Let's avoid `#ifdef`ing blocks of code. We should be able to easily parameterize the min/max values instead.

Something like:
```
#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
    constexpr int c_qtWeightMin { 0 };
    constexpr int c_qtWeightMax{ 100 };
#else
    constexpr int c_qtWeightMin { 1 };
    constexpr int c_qtWeightMax{ 1000 };
#endif
...
if (weight < c_qtWeightMin || weight > (c_qtWeightMax - 1))
{
...
```

pastalian:
I'm not very familiar with modern C++ coding style, so sorry if I'm missing the point.

jgehrig:
Thanks for applying the changes and sorry for the confusion.

I wasn't clear, the `#if` should be at the top of the file and min/max as global constants. The idea is to keep the preprocesor separated from code as much as possible. This is good practice in C or C++. 

The Modern C++ `constexpr` stuff doesn't matter too much, a `#define` would work too. Either way is fine with me for defining the constant values.

pastalian:
I see. I was wondering why your variable names were so verbose, but you intended them to be global variables.

jgehrig:
Merged, thanks for finding and fixing this bug 👍

PR #1129, opened 2024-11-03
refactor: replace non-empty QString constructors with QStringLiteral()

Integral-Tech:
- Qt provides a macro [QStringLiteral()](https://doc.qt.io/qt-6/qstring.html#QStringLiteral), which makes constructing QString objects from string literals more efficient.


jgehrig:
@Integral-Tech 

Thanks for the PR!

The changes you're making to `src/auto` are all auto-generated. We should revert them, and re-upload the results of the CI run: https://github.com/equalsraf/neovim-qt/actions/workflows/neovim-api.yml

It will probably yield the same code -- But we should not make manual changes to those files.

It looks like the CI is broken due to a package deprecation. I'll quickly do a version bump, so the run succeeds again.

jgehrig:
You should be all-set to pull in the changes from:
https://github.com/equalsraf/neovim-qt/pull/1130

Once those changes are present in your branch, you can download the updated and auto-generated versions of the neovim-api* files from the Artifacts section of the CI run that will trigger here:
https://github.com/equalsraf/neovim-qt/actions/workflows/neovim-api.yml

After the auto-generated files are replaced, this PR looks good to me :+1:

Integral-Tech:
@jgehrig I have pulled from master, but "ci/circleci: build" still fails

jgehrig:
Yeah, the CI needs some attention and cleaning…

You can ignore the CircleCI failures and anything else that isn’t directly related to your change.

The only one that matters here is the GitHub Actions run that generates the neovimapi code. You’ll need to download and commit those build artifacts.

Integral-Tech:
> Yeah, the CI needs some attention and cleaning…
> 
> You can ignore the CircleCI failures and anything else that isn’t directly related to your change.
> 
> The only one that matters here is the GitHub Actions run that generates the neovimapi code. You’ll need to download and commit those build artifacts.

Done :)

PR #1139, opened 2025-08-08
Fix CloseEvent

wolfschaf:
### Problem
In Linux, when trying to log out or shut down, the session manager tries to exit the Nvim-qt application.
Nvim-qt ignores that request, but closes anyway and will stop the system from log out / shut down.
Also, if a file is unsaved in Nvim, Nvim-qt will break the functionality of the system user notification, by ignoring the closing request immediately. So the user never sees this system notification:

image

### Fix
Fix unnecessary calls that trigger CloseEvents of the Main Window. If there is unsaved data, the CloseEvent will now stall until the user decides to Save/NotSave/Cancel. This stall will tell the Session Manager that the Nvim process is blocking the close request.

### Disclaimer
This is my first time working with QT, I have no idea what I'm doing, so I may have missed something.
I'm using Linux Mint, but this should also apply to other distros and hopefully other operating systems.

## Some printf debug info. 
Code can be found in Debug branch.

### Current version: ([Debug Branch](https://github.com/wolfschaf/neovim-qt/tree/debug))
**Normal Exit** (Gui Close/Alt-F4/Session Manager Close)
1.   MainWindow: closeEvent start
2.   Shell:      closeEvent ignore
3.   MainWindow: closeEvent ignore
4.   Shell:      handleCloseEvent
5.   MainWindow: neovimGuiCloseRequest
6.   Shell:      neovimExited close
7.   Shell:      closeEvent close
8.   MainWindow: neovimExited close
9.   MainWindow: closeEvent start
10.   MainWindow: closeEvent requested close
11.   Waited for neovim close, timed out: false
12.   MainWindow: closeEvent start
13.   MainWindow: closeEvent requested close

**Nvim Close** (:q)
1.   Shell:      handleCloseEvent
2.   MainWindow: neovimGuiCloseRequest
3.   Shell:      neovimExited close
4.   Shell:      closeEvent close
5.   MainWindow: neovimExited close
6.   MainWindow: closeEvent start
7.   MainWindow: closeEvent requested close
8.   Waited for neovim close, timed out: false
9.   MainWindow: closeEvent start
10.   MainWindow: closeEvent requested close

### Fixed Version: ([Fixed Debug Branch](https://github.com/wolfschaf/neovim-qt/tree/fixCloseEventDebug))
**Normal Exit**
1.   MainWindow: closeEvent start
2.   Shell:      closeEvent start
3.   Shell:      handleCloseEvent
4.   Shell:      neovimExited
5.   Shell:      closeEvent close
6.   MainWindow: closeEvent shell close
7.   MainWindow: handleClosing

**Nvim Close** (:q)
1.   Shell:      handleCloseEvent
2.   Shell:      neovimExited
3.   MainWindow: neovimExited close
4.   MainWindow: closeEvent start
5.   Shell:      closeEvent start
6.   Shell:      closeEvent close
7.   MainWindow: closeEvent shell close
8.   MainWindow: handleClosing


wolfschaf:
> Hi, I did not get a chance to have a proper look yet - but it seems the tests are failing - at a glance the exit status code is not being passed out

Yeah, I saw that, should be an easy fix, I was just unsure if that is the right thing to do? 

```c++
void MainWindow::neovimExited(int status)
{
	if (m_nvim->errorCause() != NeovimConnector::NoError) {
		m_errorWidget->setText(m_nvim->errorString());
		m_errorWidget->showReconnect(m_nvim->canReconnect());
		m_stack.setCurrentIndex(0);
	} else if (status != 0) {
		m_errorWidget->setText(QStringLiteral("Neovim exited with status code (%1)").arg(status));
		m_errorWidget->showReconnect(m_nvim->canReconnect());
		m_stack.setCurrentIndex(0);
	} else if (!m_inCloseEvent) {
		//deleting the else and setting m_exitStatus fixes most of it,
		//but then what's the point of the code above?
		qDebug("MainWindow: neovimExited close");
		close();
	}
}```

**Comment by wolfschaf** [comment]
I'm also investigating if the warnings in the tests are coming from my changes or not
> QWARN  : NeovimQt::TestShell::benchStart() QProcess: Destroyed while process ("nvim") is still running.

File: src/gui/mainwindow.cpp [hunk 1]
```diff
@@ -254,22 +255,7 @@ void MainWindow::neovimFullScreen(bool set)
 
 void MainWindow::neovimGuiCloseRequest(int status)
 {
-	m_neovim_requested_close = true;
 	m_exitStatus = status;
```
equalsraf, line 258: If i understand the idea correctly neovimguicloserequest now only sets status code (triggered from shell::neovimguicloserequest). The actual shutdown is now handled from MainWindow::handleClosing by handling process shutdown and using the status code as appropriate.

There is one problematic case for this - which is when nvim is not spawned but are attaching to a running nvim. Something like

```
nvim --listen /tmp/mynvim.socket
nvim-qt --server /tmp/mynvim.socket
```

and then in nvim `call rpcnotify(0, 'Gui', 'Close', 1)` would set the status but not lead to a shutdown.
But in fairness this seems to broken at the moment anyway, so lets ignore it for now.

Can we set `m_exitStatus = 0` when `handleNeovimAttachment(true)`? That way we avoid issues with reattachment.

src/gui/shell.cpp
```diff
@@ -1709,12 +1706,21 @@ void Shell::closeEvent(QCloseEvent *ev)
...
+
+		// Try to wait for neovim to quit
+		QEventLoop loop;
+		connect(m_nvim, &NeovimConnector::processExited, &loop, &QEventLoop::quit);
+		connect(m_nvim, &NeovimConnector::aboutToClose,  &loop, &QEventLoop::quit);
+		connect(m_nvim->api0(), &NeovimApi0::on_vim_command, &loop, [&loop, ev](){
+			//This will fire if we cancel the closing
+			ev->ignore();
+			loop.quit();
+		});
```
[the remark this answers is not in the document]
equalsraf: You could check specifically for that command by connecting to MsgpackRequest::finished

```c++
		auto req = m_nvim->api0()->vim_command("confirm qa");
		connect(req, &MsgpackRequest::finished, &loop, [&loop, ev](){
			// This will fire if we cancel the closing
			qDebug() << "confirm qa fini";
			ev->ignore();
			loop.quit();
		});
```

equalsraf (changes requested):
Thanks for working on this. LGTM with a couple changes.

It seems we have a few more issues in this code path when dealing w/ remote attachments, buts leave that outside of this review.

equalsraf:
On a another note, I have pushed some fixes to CI for the clang format and tidy steps. So if your rebase you get those for free :)

src/gui/mainwindow.cpp
wolfschaf: Oh thanks, haven't even thought about that. But I'm not sure I follow what difference it makes when setting `m_exitStatus=0`? (If we open QT --server and then exit, m_exitStatus could be undefined, but have to look at what actually happens, but it shouldn't affect reattachment in any way?)


But to clarify, are these all of the close request cases?
- System/User requests close (Pressing X -> Gui Close Event)
- System/User closes Nvim (:q/crash ->Nvim process exits)
- rpcnotify()
- (POSIX Signals)
And when Nvim is spawned, every case should close Nvim & QT, right?


What case should close Nvim and/or QT when it is attached? Master version always closes one or the other, never both, but maybe fixing that could be a separate Pull Request.

equalsraf: The first two cases you listed are the main ones.

The rpcnotify should already be called by nvim during a regular close (:q or :cq) this is part of the vimscript plugin in src/gui/runtime.

> (POSIX Signals)

I don't think you have to do anything special with regards to posix signals - you would get a different exit status - but it is the same as what happens when the nvim process crashes.

> But I'm not sure I follow what difference it makes when setting m_exitStatus=0? (If we open QT --server and then exit,

Very long corner case:

1. Start nvim --listen /tmp/wat.sock
1. Lets say you open with nvim-qt --server --nofork ....
2. nvim exits by calling `call rpcnotify(0, 'Gui', 'Close', 4)` and then quiting
3. nvim-qt will not close, instead it stays open with the error msg and retry button (internally m_exitStatus is **4**)
4. restart nvim --listen /tmp/wat.sock
5. reattach by pressing the `Retry` button
6. close the nvim-qt window (but not nvim) - the exit status for nvim-qt will be `4`

Feel free to leave it, I'm probably one of the few people doing this kind of thing.

> And when Nvim is spawned, every case should close Nvim & QT, right?

yes. spawned here means nvim-qt started nvim. So we need to close it.

> What case should close Nvim and/or QT when it is attached? Master version always closes one or the other, never both, but maybe fixing that could be a separate Pull Request.

Yes lets leave this for future PRs.

There is probably an argument to be made here - should we just close blindly when e.g. nvim crashes? but this applies to both cases.

wolfschaf: > Very long corner case

Oh yeah, that's legit

> There is probably an argument to be made here - should we just close blindly when e.g. nvim crashes? but this applies to both cases.

Yeah, for now I think it's better to leave the behavior the same as before the fix, so I will readjust it to show that again and later one could have an argument about that.

wolfschaf:
> What are you seeing with log out/shutdown?

Normally I expect that a session manager tries to close the application, and if we ignore and continue the app, it should close it forcible.
Or when the user tries to close the app normally multiple times, the first time it will ask (if there is an unsaved buffer) and the second time it will just close everything.

Locally the normal case works now (a small fix for the forceful shutdown is still missing in this PR), but I still have to figure out why the Session Manager 'ignore and continue' does not work (After a timeout the SM terminates the application, so at least it's better than before, where nvim-qt just ignored the request but closed anyway thus hindering the log out).

I looked at QTSessionManager, which may be the solution, but haven't experimented long enough with it.


> And btw is there a way to test this without actually running a full session manager?

Currently one could just test by sending "wmctrl -c" or a normal close event and test the save/cancel/2xCloseEvent cases. But if it requires the SM, then I'm not sure yet.

wolfschaf:
This commit fixes the forceful termination of the app when the user/Window Manager tries to close the app a second time, by intercepting the Quit Event and closing everything cleanly. Previously the main thread would just exit, but the main process and the nvim subprocess would continue to run. The master version would retry to show the Save/No Save/Cancel prompt, but it would sometimes bug out a little.



### New Behavior in this PR
 - rpcnotify is fixed now. (In --server mode it only closes GUI, :q to close both)
 - CloseEvent blocks further execution if there is unsaved progress which allows a Session Manager to notice that the app may have unsaved work.
 - When seeing the Save/No/Cancel prompt, if the user/WM closes the app a second time, the app will close now.

Also tested a bit with rpcnotify(..., 'NewWindow').
In --server mode when closing with `:q`, there may be a race condition where it only closes the GUI. But I only saw this once during testing. (Unsure if these changes are even the cause of that)


### Unfortunately I learned
 - It is recommended that the CloseEvent should not be blocking, but apparently other apps do it and everything seems to work fine. Instead one should stall inside of the the QSessionManager commitDataRequest signal handler. I've got it mostly working, but it will take a lot more work since QT recommends: "You should not exit the application within this signal". (I also have to clean up things since I have a lot of code duplication going on)
- Immediately closing the app when the user clicks "Ignore and Continue" in the Session Manager prompt is impossible to implement in QT. There may be ways to work around this, but I would have to look deeper into that.



I the last few days I didn't have the time to complete the recommended option yet and in the next two weeks I'm on vacation, so the experimentation has to wait unfortunately.
So if you want, you can merge this PR, or we can wait until I fully implemented the QSessionManager approach and compare the two against each other. (Both won't have to Ignore and Continue feature for now)
For the later option I can also write some tests, since it should be possible to mimic the Session Manager.


Keywords: XSMP, DBus

My conversation with Chat-GPT, if anybody is interested:
https://chatgpt.com/share/68a07de8-0084-800a-ba3a-f9775c6d3c99

equalsraf:
merged, thanks @wolfschaf - make sure to enjoy your vacation time

PR #1144, opened 2025-11-08
Fix building the tests with clang

heirecka:
Clang (I only tested with versions >= 18) complains about narrowing from long long int (qsizetype) to int on my platform. gcc accepts it, but issues a warning. So just use qsizetype instead of int as type of the variable as well.


heirecka:
The CI errors our because of newer cmake, this [MR](https://github.com/equalsraf/neovim-qt/pull/1145) should address it.

PR #1145, opened 2025-11-08
Update msgpack(-c) to 6.1.0 and simplify fetching it, fix CI

heirecka:
msgpack split its mixed C/C++ tarball into two separate ones starting with 4.0 [1][2]. Since 6.0.0 the C part uses msgpack-c as a name for everything. This is where the requirement for 6.0.0 comes from, but it has been released over two years ago and I'd expect it to be available for everybody who wants to use -DUSE_SYSTEM_MSGPACK=TRUE. It allows to just use the imported target from msgpack-c's cmake config file and to simplify the cmake code a bit.
This fixes the build with CMake >= 4.0, which has removed compatibility with CMake < 3.5 while msgpack 3.2.0 sets "CMAKE_MINIMUM_REQUIRED (VERSION 2.8.12)".

[1] https://github.com/msgpack/msgpack-c/pull/876
[2] https://github.com/msgpack/msgpack-c/pull/878
[3] https://github.com/msgpack/msgpack-c/pull/1053

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,11 +105,7 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR
 			"-bugprone-unhandled-exception-at-new,"
 			"-readability-suspicious-call-argument,"
 			)
-		if(${CMAKE_VERSION} VERSION_LESS "3.6.0")
-			message("ENABLE_TIDY requires CMake >= 3.6.0")
-		else()
-			set(CMAKE_CXX_CLANG_TIDY clang-tidy "-checks=${TIDY_CHECKS};-warnings-as-errors=*")
-		endif()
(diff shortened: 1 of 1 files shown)


equalsraf:
I'll have a quick look at appveyor, but it does not seem to be caused by this so no worries. I Reran a build for master and that failed too https://ci.appveyor.com/project/equalsraf/neovim-qt/build/job/kdkt12rwqh3ke6nx

equalsraf (changes requested):
I think FetchContent requires cmake 3.11. Can we bump cmake_minimum_required?

equalsraf:
Concerning appveyor it seems that images were updated and the qt versions/paths don't match anymore. I'm not worried about the change since we also build windows in other builders (just annoying because appveyor builds our releases). 

I'm ok with merging this, after that one liner update

PR #1150, opened 2026-03-05
Add support for variable font axes

iteratee:
Requires QT 6.7+, but when compiled with QT 6.7+, A user can now select a variable font on any custom axis, not just `wght` and `slnt`

For QT 6.9+, it will verify that the axes are valid by comparing against the QFontInfo. Versions prior just blindly apply the axes.

### DIFF
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -433,7 +433,7 @@ void ShellWidget::paintForegroundCellText(
 	// Draw chars at the baseline
 	const int cellTextOffset{ m_ascent + (m_lineSpace / 2) };
 	const QPoint pos{ cellRect.left(), cellRect.top() + cellTextOffset};
-	const uint character{ cell.GetCharacter() };
+	const char32_t character{ cell.GetCharacter() };
 	const QString text{ QString::fromUcs4(&character, 1) };
 
 	p.drawText(pos, text);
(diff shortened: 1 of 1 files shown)


iteratee:
The windows failures appeared transient, and not related to the code changes.

jgehrig:
@iteratee Thanks for the Pull Request!

Sorry for the slow response -- I'm happy to get this merged, but I have a few questions/comments first :)

1. I'm not very familiar with "font axes". Can you give me a quick description/example of how this is used for my understanding and testing? Is this feature already supported in the Neovim Terminal UI or other GUIs?

2. Are we adding support for options in `:set guifont` and `:GuiFont`. If so, we should update the [documentation here](https://github.com/equalsraf/neovim-qt/blob/99d26d355545e7241a0b74586cccd636c88629fc/src/gui/runtime/doc/nvim_gui_shim.txt#L38) with a brief description.

3. We should avoid adding `#ifdefs` into the code whenever possible, especially mid-function. It creates a maintainability problem. However, I think we can cleanly remove these ifdef-s with some minor refactoring.

Let's introduce functions into something like `fontaxissupport.cpp` and `fontaxissupport_null.cpp`. See [this directory](https://github.com/equalsraf/neovim-qt/tree/master/src/compat) for an example. Everything you want to call that requires a qt-version-ifdef should be placed into a function that can be conditionally linked. Everything should be structured so we can return a nullish value for the not-supported case.

You can control linkage with the CMake logic: Qt >= 6.7 -> `fontaxissupport.cpp` or Qt < 6.7 -> `fontaxissupport_null.cpp`.

Let me know if you're able to make these modifications.

PR #1153, opened 2026-03-29
Fix Qt6 deprecation warnings

twelveeighty:
These changes fix the deprecation warnings generated during a build against Qt6. Most of these are due to the deprecation of QVariant::type(), but I also addressed the deprecation of QString::fromUcs4(uint*,) and qAsConst().

This is my first pull request for this project, please let me know if any changes need to be made.

### DIFF
--- a/src/gui/shellwidget/helpers.cpp
+++ b/src/gui/shellwidget/helpers.cpp
@@ -54,7 +54,11 @@ bool saveShellContents(const ShellContents& s, const QString& filename)
 			if (cell.GetBackgroundColor().isValid()) {
 				p.fillRect(r, cell.GetBackgroundColor());
 			}
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 			const uint character{ cell.GetCharacter() };
+#else
+			const char32_t character{ static_cast<char32_t>(cell.GetCharacter()) };
+#endif
 			p.drawText(r, QString::fromUcs4(&character, 1));
 		}
 	}
--- a/src/gui/shellwidget/shellwidget.cpp
+++ b/src/gui/shellwidget/shellwidget.cpp
@@ -432,7 +432,11 @@ void ShellWidget::paintForegroundCellText(
 	// Draw chars at the baseline
 	const int cellTextOffset{ m_ascent + (m_lineSpace / 2) };
 	const QPoint pos{ cellRect.left(), cellRect.top() + cellTextOffset};
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 	const uint character{ cell.GetCharacter() };
+#else
+	const char32_t character{ static_cast<char32_t>(cell.GetCharacter()) };
+#endif
 	const QString text{ QString::fromUcs4(&character, 1) };
 
 	p.drawText(pos, text);
@@ -712,7 +716,11 @@ void ShellWidget::paintRectLigatures(QPainter& p, const QRect rect) noexcept
 					blockCursorPos = blockText.size();
 				}
 
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 				const uint cellCharacter{ checkCell.GetCharacter() };
+#else
+				const char32_t cellCharacter{ static_cast<char32_t>(checkCell.GetCharacter()) };
+#endif
 				blockText += QString::fromUcs4(&cellCharacter, 1);
 
 				if (checkCell.IsDoubleWidth()) {
@@ -1035,7 +1043,11 @@ QVariant ShellWidget::TryGetQFontFromDescription(const QString& fdesc) const noe
 	qreal pointSizeF = curFont.pointSizeF();
 	int weight = -1;
 	bool italic = false;
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 	for (const auto& attr : qAsConst(attrs)) {
+#else
+	for (const auto& attr : std::as_const(attrs)) {
+#endif
 		if (attr.size() >= 2 && attr[0] == 'h') {
 			bool ok{ false };
 			qreal height = midString(attr, 1).toFloat(&ok);
@@ -1071,7 +1083,11 @@ QVariant ShellWidget::TryGetQFontFromDescription(const QString& fdesc) const noe
 
 /*static*/ bool ShellWidget::IsValidFont(const QVariant& variant) noexcept
 {
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 	return static_cast<QMetaType::Type>(variant.type()) == QMetaType::QFont;
+#else
+	return variant.metaType().id() == QMetaType::QFont;
+#endif
 }
 
 /*static*/ bool ShellWidget::isBadMonospace(const QFont& f) noexcept
--- a/src/gui/app.cpp
+++ b/src/gui/app.cpp
@@ -473,7 +473,11 @@ void App::openNewWindow(const QVariantList& args) noexcept
 	QString server;
 	QString nvim{ "nvim" };
 	ConnectorInitArgs::Type type{ ConnectorInitArgs::Type::Default };
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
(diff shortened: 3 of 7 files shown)


src/gui/shellwidget/helpers.cpp
```diff
@@ -54,7 +54,11 @@ bool saveShellContents(const ShellContents& s, const QString& filename)
 			if (cell.GetBackgroundColor().isValid()) {
 				p.fillRect(r, cell.GetBackgroundColor());
 			}
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 			const uint character{ cell.GetCharacter() };
+#else
+			const char32_t character{ static_cast<char32_t>(cell.GetCharacter()) };
```
jgehrig, line 60: Something like this probably doesn't need an `ifdef`.

It would be better to write this code so it is compatible with all supported versions of QT, since the new API you're using is part of the STL.

There are a few others changes like this.

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -1071,7 +1083,11 @@ QVariant ShellWidget::TryGetQFontFromDescription(const QString& fdesc) const noe
 
 /*static*/ bool ShellWidget::IsValidFont(const QVariant& variant) noexcept
 {
+#if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
 	return static_cast<QMetaType::Type>(variant.type()) == QMetaType::QFont;
+#else
+	return variant.metaType().id() == QMetaType::QFont;
```
jgehrig, line 1089: It might be better to write a wrapper-function for this, since you're making this change in multiple places. That way you've only introduced a single `#ifdef`.

For example (psuedo-code):
```
QMetaType GetTypeFromVariant(const QVariant& var) noexcept
{
  // Comment describing why this ifdef exists, and when it can be removed
  #if QT_VERSION < 6.0
     return static_cast<QMetaType::Type>(var.type())
  #else
     return var.metaType().id()
  #endif
}
```

Now this snippet can become
```
/*static*/ bool ShellWidget::IsValidFont(const QVariant& variant) noexcept
{
	return GetTypeFromVariant(variant) == QMetaType::QFont;
}
```

jgehrig:
@twelveeighty

Thanks for the Pull Request! It's always good to see new contributions :)

I'm generally hesitant to add `#ifdef`s in the code -- My long-term goal as we refactor is to eliminate preprocessor macros, so I'd like to avoid adding them unless absolutely necessary. They can become difficult to maintain as it generates different code-paths to test for each macro we introduce. For example, we had lots of bugs in the input layer due to if-defing input handling code by-platform.

I'd like to understand two things:
 - What is the consequence of simply ignoring these deprecation warnings?
 - Can we fix some of these by bumping the minimum QT version and updating the code instead?

Let's try to eliminate adding new preprocessor macros here -- Or at least keep the count to an absolute minimum.

These are two good techniques to avoid or contain preprocessor macro usage:
 - Use the build system to provide alternative implementations controlled via CMake.
   [Here is an example from the input layer](https://github.com/equalsraf/neovim-qt/commit/548bb1e09aa339571ca4cc8625529fcdc37ceb10). We compile `_platform.cpp` instead of `#ifdef Q_OS_MAC` individual blocks.
 - If the build system is impractical, isolate the ifdef into a function call. [Example here](https://github.com/equalsraf/neovim-qt/pull/1153/changes/bc9c5f5e4da1257ea3dc61ea7cfd3aa18ebf713c#r3022530159).

src/gui/shellwidget/helpers.cpp
twelveeighty: The better but more drastic alternative is to change the gui/shellwidget/Cell class member `uint character` to `char32_t character`. I wasn't completely sure if that would break supported platforms (see my comment below on bumping Qt version). I'll explore doing that, I agree that would be cleaner.

src/gui/shellwidget/shellwidget.cpp
twelveeighty: In general, I agree to minimize `#ifdef`s, it makes the code hard to read. I should have mentioned in my pull request that I intended these to be only temporary (weeks, months perhaps) until you would bump the minimum Qt version to 6 and then factor out the ifdefs. But, I do not know if that would be a big deal with the other supported platforms. It's easy for me, I'm on Arch Linux, I could do without Qt5 altogether today.

I'll reply to your original comment below with a more general question on how to proceed - probably easier to communicate there.

twelveeighty:
> Thanks for the Pull Request! It's always good to see new contributions :)

Glad to help, I use neovim-qt on a daily basis and I have some free time for the next couple of months, so if you have any priority things that need work, let me know and I can take a look.

> * What is the consequence of simply ignoring these deprecation warnings?

That's a project philosophy question. For me personally, I've always considered warnings as errors, to be either suppressed on a case-by-case basis, or fixed altogether, but I can totally understand if that's not important to the project.

> * Can we fix some of these by bumping the minimum QT version and updating the code instead?

Well, that's the ultimate question. Which platforms that the project supports currently do not have Qt6? If you are OK to bump the Qt requirement to 6 then this pull request becomes a LOT cleaner, of course. Qt6 has been around since 2020 so maybe it's a moot point by now?

> These are two good techniques to avoid or contain preprocessor macro usage:

Let me know what you prefer: keep the support for Qt5, in which case I'll explore the alternatives to `#ifdef` as you suggested, or drop Qt5 support and I'll clean up the code accordingly.

jgehrig:
Thanks for the responses!

>That's a project philosophy question. For me personally, I've always considered warnings as errors, to be either suppressed on a case-by-case basis, or fixed altogether, but I can totally understand if that's not important to the project.

I'm in agreement there -- In an ideal world, I also like to treat warnings as errors.

The Qt deprecation warnings are tricky. I think the best path forward is to keep them around until they're dropped by new versions of the framework, that maintains maximum backwards compatibility when users go to compile this code. Once the minimum QT version supports the new API we can switch over.

For development: treat as many warnings as errors possible. Only silence errors when fixing them would break a required dependency.

For release: silence warnings whenever possible. This prevents bug-reports or build failures from coming in when a compiler or framework decides to introduce a new linting rule.

---

It looks like Ubuntu 26.04 LTS supports QT6, that's usually my bar for setting minimum package versions.

I'm in support of bumping the minimum Qt version to 6. Let's go with that path forward!

Let me know if you have any questions. I look forward to getting a Qt6 update merged! 🎉

jgehrig:
> Glad to help, I use neovim-qt on a daily basis and I have some free time for the next couple of months, so if you have any priority things that need work, let me know and I can take a look.

Awesome! I'll think about this and let you know if I come up with any good intro tasks.

Definitely feel free to work on what you'd like to see neovim-qt support. Open issues or pull requests for anything you'd like to work on.  I'll do my best to help design/plan/implement and review/merge code.

Personally, I've always wanted `ext_messages` support, but it is a tricky one!

twelveeighty:
> I'm in support of bumping the minimum Qt version to 6. Let's go with that path forward!

Perfect, I'll work on that. I'll probably eventually "drop" this pull request in favor of that "Qt6 minimum" pull.

Finally: two questions: 1) is there any debug info we can obtain on why the automated workflows fail? It says 0 errors / 0 warnings but they are in a Failed state and 2) I was thinking of opening a "Make Qt6 the minimum supported Qt version" Issue so we have a place to communicate any issues/decisions that need to be made should they come up, let me know if you are OK with that.

jgehrig:
> 1) is there any debug info we can obtain on why the automated workflows fail? It says 0 errors / 0 warnings but they are in a Failed state and

The CI is a bit messy at the moment -- Eventually, it'd be nice to get everything moved over to GitHub Actions. AppVeyor and PipeLine breaks look like stale images/files unrelated to your change.

I think I needed to approve the run. Until you've made a few commits GitHub asks for approval, as a security feature.

> 2) I was thinking of opening a "Make Qt6 the minimum supported Qt version" Issue so we have a place to communicate any issues/decisions that need to be made should they come up, let me know if you are OK with that.

Perfect! That sounds like a good way to track the work and planning.

twelveeighty:
Closing this PR to replace it with the PR(s) related to issue #1154

Issue #750, opened 2020-09-20
Remove Deprecated Code: QFontMetrics::width


jgehrig:
See `src/gui/shellwidget/helpers.h`, we should remove `GetHorizontalAdvance(...)`.

The minimum required Qt version is 5.11.

jgehrig:
Pull Request: #744

equalsraf:
Removed in 8ddc395978c488cf49b62b4f3b3848def8f6ec13

Issue #1154, opened 2026-04-02
Make Qt6 the minimum supported Qt version


twelveeighty:
This issue is to track the efforts to raise the minimum supported Qt version to Qt6.

twelveeighty:
@jgehrig it looks like the Pipelines are failing because they were configured for Qt5. Should we ignore the Pipelines build failures and focus on GitHub Actions instead? I can work on those Actions in my forked repo, I believe, to keep the noise down. Is there a specific test or coverage that need to be added to Actions that is currently only in Pipelines?

jgehrig:
Yeah, the CI is messy and neglected at the moment.

Long-term, I'd like to see everything operate on GitHub Actions -- It is well integrated and supports all platforms. I'm in full support of validating via GitHub Actions.

Many of the other CIs don't support every platform at our free open-source tier (AppVeyor, CircleCI), and Azure Pipelines does not display error messages well. The blocker for moving over to GitHub Actions entirely was dropped unit test output -- I don't recall the specifics, but there was some sort of output piping bug that blocked us in either Actions or the UT framework. 

You should be able to poke the YAML files for each CI in the repository -- I'm not opposed to deprecating any of the CI runs, if we have a good reason.

Let's go through each failing pipeline and make a decision once the code is ready.

For CI coverage, I'd like to see:
 - Mac/Linux/Windows builds + UTs
 - Ubuntu LTS build
 - Minimum Qt 6.0 build on at least one platform
 - A recent Qt6 release built on at least one platform

Short term we may be able to validate some of those locally, if the CI isn't cooperative.

twelveeighty:
I pushed two more commits to the PR to round out the actual code changes, it should now pass the `clang-format-diff` Action, the other Action/CIs will fail since they are for Qt5. 
The Actions/CI migration to Qt6 will be tricky, so I *think* I should start a new sub-branch for those changes. Can you review the following plan, I'd like to run it by the group for review and any tips:

1. Don't commit the Action/CI changes to the PR branch, but make those changes on a different branch (`qt6-ci`), perhaps a sub-branch of the PR branch
2. This will "pause" the commit chain on the PR
3. In my forked repo, add the `workflow_dispatch` event to every Action so I can trigger them manually and change the `on.push.branches` events to trigger off the `qt6-ci` branch.
4. That should keep my `master` branch clean and able to follow any new commits upstream while be able to build and test the Actions
[...]

PR #1155, opened 2026-04-02
Raise minimum Qt requirement to 6.0

twelveeighty:
See #1154


jgehrig:
Once this work is complete, we should be able to close Issue #750.

Let's go through the issue list and see if anything else is closable. I can take a first pass.

If other issues are found, maybe we should create a list at the top-level-comment either in the PR or this Issue?

twelveeighty:
> Once this work is complete, we should be able to close Issue #750.
> Let's go through the issue list and see if anything else is closable. I can take a first pass.

Yes, there also may be more "hidden" Qt5 workarounds or hacks that can be more gracefully done in Qt6, but I only focused on those files that had the deprecation warnings and those that referenced the gui/shellwidget/Cell class.

> If other issues are found, maybe we should create a list at the top-level-comment either in the PR or this Issue?

Either way works for me, but it might be more clear if we create the list at the Issue level.

I also realized I forgot to take a pass through the contrib/ pipelines/CI configs, I'll have to add a commit with those changes.

jgehrig:
I think you've got some dead code here:
[src/compat](https://github.com/equalsraf/neovim-qt/tree/master/src/compat)

src/gui/shell.cpp
```diff
@@ -1766,8 +1753,10 @@ void Shell::tooltip(const QString& text)
 		m_tooltip->show();
 	}
 
-	m_tooltip->setMinimumWidth(GetHorizontalAdvance(QFontMetrics{ m_tooltip->font() }, text));
-	m_tooltip->setMaximumWidth(GetHorizontalAdvance(QFontMetrics{ m_tooltip->font() }, text));
+	QFontMetrics fm(m_tooltip->font());
+	int width = fm.horizontalAdvance(text);
```
jgehrig: nitpick: Prefer `{}`

```suggestion
	int width{ fm.horizontalAdvance(text) };
```

src/gui/shell.cpp
```diff
@@ -1411,8 +1403,8 @@ void Shell::neovimMouseEvent(QMouseEvent *ev)
 		return;
 	}
 
-	QPoint pos(ev->x()/cellSize().width(),
-			ev->y()/cellSize().height());
+	QPointF eventPos = ev->position();
+	QPoint pos(eventPos.x() / cellSize().width(), eventPos.y() / cellSize().height());
```
jgehrig: nitpick: prefer `{}`

```suggestion
	QPointF eventPos{ ev->position() };
	QPoint pos{ eventPos.x() / cellSize().width(), eventPos.y() / cellSize().height() };
```

src/gui/shell.cpp
```diff
@@ -1476,8 +1468,8 @@ void Shell::mouseMoveEvent(QMouseEvent *ev)
 {
 	setCursorFromBusyState();
 
-	QPoint pos(ev->x()/cellSize().width(),
-			ev->y()/cellSize().height());
+	QPointF eventPos = ev->position();
+	QPoint pos(eventPos.x() / cellSize().width(), eventPos.y() / cellSize().height());
```
jgehrig: nitpick: Prefer `{}`
```suggestion
	QPointF eventPos{ ev->position() };
	QPoint pos{ eventPos.x() / cellSize().width(), eventPos.y() / cellSize().height() };
```

src/gui/shell.cpp
```diff
@@ -344,7 +343,7 @@ void Shell::init()
 	connect(m_nvim->api0(), &NeovimApi0::on_ui_try_resize,
 			this, &Shell::neovimResizeFinished);
 
-	QRect screenRect = screenAvailableGeometry(this);
+	QRect screenRect = this->screen()->availableGeometry();
```
jgehrig: I see the existing code does this, but we really should check `screen()` for `nullptr`, it is technically a possible return value.

```suggestion
	QRect screenRect {};
	if (this->screen())
	{
		screenRect = this->screen()->availableGeometry();
	}
```

src/gui/CMakeLists.txt
```diff
@@ -121,13 +114,7 @@ endif()
 
 if(WIN32 AND NOT CMAKE_CROSSCOMPILING AND NOT DEFINED USE_STATIC_QT)
 	include(WinDeployQt)
-
-	set(__QT_EXCLUDE_DEPLOY_MODULES webkit webkit2)
-	if(${QT_VERSION_MAJOR} VERSION_GREATER_EQUAL 6)
-		set(__QT_EXCLUDE_DEPLOY_MODULES )
-	endif()
-
-	WinDeployQt(TARGET nvim-qt COMPILER_RUNTIME INCLUDE_MODULES ${QTLIBS} EXCLUDE_MODULES ${__QT_EXCLUDE_DEPLOY_MODULES})
```
jgehrig, line 130: What's the reason behind removing this block?

It looks like both were previously excluded for Qt6, and now neither are excluded for Qt6.

.github/workflows/build-test.yml
```diff
@@ -10,45 +10,38 @@ jobs:
...
-          - name: MacOS_GCC
+          - name: MacOS_Debug
             flavor: Debug
-            runner: macos-13
+            runner: macos-latest
             generator: Ninja
-            cc: gcc-12
-            cxx: g++-12
+            cc: clang
+            cxx: clang++
```
jgehrig, line 35: The `gcc` was intentional, so we build against both `clang` and `gcc` on Mac. Let's find some way to cover `gcc` too.

(Comment for Tracking -- No need to address here, I will file an Issue)

.github/workflows/build-test-win.yml
```diff
@@ -0,0 +1,55 @@
+name: Build And Deploy
```
jgehrig, line 1: With the way GitHub Actions is structured, it would be nice to have this in `build-test.yml` so the build is grouped with the other platforms.

Agreed, this has diverged enough to remove from the matrix.

(Comment for Tracking -- No need to address here, I will file an Issue)

jgehrig:
Looks Great! I'll hit the merge button after a few small questions/comments.

We'll continue to iterate on the CI, so these are for my tracking purposes. The comments like this one can be ignored:
https://github.com/equalsraf/neovim-qt/pull/1155/changes#r3086657704

I've made a few comments with trivial syntax changes. Unless you'd like to apply them (or some variant) yourself, I can quickly add the changes via the Pull Request UI before check-in.

I have only one real comment, there seems to be some CMake deployment logic changes:
https://github.com/equalsraf/neovim-qt/pull/1155#discussion_r3086655127

twelveeighty:
Thanks for the feedback!

> We'll continue to iterate on the CI, so these are for my tracking purposes. The comments like this one can be ignored: https://github.com/equalsraf/neovim-qt/pull/1155/changes#r3086657704

I did investigate further into why the Windows tests were failing and discovered that the failures are transient - sometimes they fail, sometimes not, against the same code. That's bad news, of course, very difficult to troubleshoot. I noticed that some prior commit comments mention extending timeouts, so it may be dependent on the system load that GitHub is experiencing when running our CI containers.

> I've made a few comments with trivial syntax changes. Unless you'd like to apply them (or some variant) yourself, I can quickly add the changes via the Pull Request UI before check-in.

Of course, no issues with syntax changes, go ahead. I noticed that the different initializations were mixed throughout the code, so I tried to keep it consistent with whatever was used most in that file, rather than mixing `int x = y` with `int x{ y }`, but if it were up to me we would change all as a cleanup exercise. 

> I have only one real comment, there seems to be some CMake deployment logic changes: [#1155 (comment)](https://github.com/equalsraf/neovim-qt/pull/1155#discussion_r3086655127)

Yes, I'll take another look at that. I *believe* I reverted that code back to its commit from before the Qt6 support was added because I assumed it was introduced solely to support both Qt5 and Qt6 at the same time, but I could be wrong.

Just so I'm not creating merge conflicts: I'll touch the CMake file (if needed) and you'll update all others, correct?

jgehrig:
> I did investigate further into why the Windows tests were failing...

Yes -- Something interesting is going on there. I think it may be related to the way signals are handled on Windows in GitHub Actions. It could be related to the dropped test output; it looks like Actions runs Windows in some sort of headless mode. I think we're not correctly capturing some of the signals in this environment.

I've got a few things I'd like to investigate, but we'll wait until this update is merged.

---

Don't worry about merge conflicts, change anything you'd like.

Yes, I can handle the syntax stuff. I'll do this right before merge: 1 - apply changes via GitHub UI, 2 -run CI, 3 - squash merge PR.

src/gui/CMakeLists.txt
twelveeighty: Both webkit and webkit2 are deprecated in Qt6, so they no longer need to be *excluded*. 

Unless I am mistaken, the original code `if(${QT_VERSION_MAJOR} VERSION_GREATER_EQUAL 6)` **unsets** `__QT_EXCLUDE_DEPLOY_MODULES` so that they are no longer part of the `EXCLUDE_MODULES` array. If Qt6 is the minimum, then we no longer need that exclude array. Let me know if I did that wrong. For reference, here is the [commit](https://github.com/equalsraf/neovim-qt/commit/81a3ca1b1231319e5f03dae7144cf4b822bfa136#diff-2d8b127aed7a6123be24b80481b760ec4e9cc251e19d668d936dc09e87147a0e) that introduced that `if()` statement as part of adding Qt6 support.

src/gui/shell.cpp
twelveeighty: It looks like this doesn't compile:
```
gui/shell.cpp:1411:14: error: type 'qreal' (aka 'double') cannot be narrowed to 'int' in initializer list [-Wc++11-narrowing]
 1411 |         QPoint pos{ eventPos.x() / cellSize().width(), eventPos.y() / cellSize().height() };
```
If you like, I can try cleaning that up. The linter also complains about the placement of braces in the `if` statement in `shell.cpp`.

jgehrig: Thanks. Yes, will happen with `{}`. It can be noisy, but it can also catch bad implicit casting.

You beat me too it! My plan was to spin up a machine on the branch this evening to fix these and then merge.

The build looks clean, so I'll hit the merge button now.

PR #1159, opened 2026-04-17
Fix windows builds

equalsraf:
A few things here

1. fix for azure pipeline windows builds (change image/compiler version to visual studio 2022)
2. removed appveyor since we cannot rely on it to provide qt 5.15 binaries anymore https://github.com/appveyor/ci/issues/3978
3. added one additional task to azure for a release build with artifacts

Todo:
- [ ] publish tagged github release
- [ ] build artifact looks slightly different, it has msgpack lib/headers inside

Open questions here

- since builds include qt binaries, this bundles the aqt installer Qt binary builds with our artifacts. Is this ok? I would prefer if those aqt binaries can be pinned (maybe aqtinstall can accept a hash?)
- I just went with azure since there recipes already worked, should we use something else for build releases for windows? Should we keep appveyor for Qt6 builds?
- testing/testing/testing, if someone has windows and can test the artifacts that would be great. I did not test yet, they do have the qt libs inside

Artifacts are publish under the little related/published link at the top of the build page

- https://dev.azure.com/equalsraf/neovim-qt/_build/results?buildId=1633&view=results
- https://dev.azure.com/equalsraf/neovim-qt/_build/results?buildId=1633&view=artifacts&pathAsName=false&type=publishedArtifacts


equalsraf:
I just saw #1160, so lets just adjust this to github actions :D

jgehrig:
Sorry for the confusion! I didn't notice this soon enough...

Yes, I'd like to move over to GitHub Actions, it's the only free service I've found that can handle all platforms. I've made some progress on the Windows issues there. Dropped test output we can fix with an artifact, and I have some theories on the `tst_qsettings` failures.

I'll merge that PR today.

equalsraf:
no worries I just opened this.

I'll have a look during the weekend to see if I can get windows to build in github actions too. Right now mine is failing the tests in what looks like a dll issue (I probably forgot to add Qt to the PATH).

Issue #1149, opened 2026-03-03
Neovim 0.12 default GUIFONT value causes GUIs to spew an error


uralets:
https://github.com/neovide/neovide/issues/3370

<img width="998" height="646" alt="Image" src="https://github.com/user-attachments/assets/9e8794a3-9734-4838-9a14-93ef0fad1f15" />

tan-wei:
Hmm, seems a upstream change cause this error message.

mikew:
Yep, here's the PR in neovim: https://github.com/neovim/neovim/pull/37175

uralets:
It's now an officially released Neovim version.

jgehrig:
I'll put together a fix in the next day or two. We can make `option guifont` and `:GuiFont` fail silently.

We should also use this as an opportunity to address the 'reports bad fixed pitch metrics` issues. Issues like #732.

uralets:
> the 'reports bad fixed pitch metrics`

I've been hitting those too on switching fonts, but at least these are inconsequential (?) and you don't change the font often anyway

gi1242:
I used to have my font to be `Fira Code:h22`. Running `Guifont Fira\ Code:h22` works. When I look at `~/.config/nvim-qt/nvim-qt.conf` it shows that the font now changed to `Fira Code:h22`. However when I open nvim-qt again, my font has been reset to Source Code Pro. Do you know why, and how I can get my Fira Code font back?

equalsraf:
This is for the case where the font is passed via `set guifont=` right? Not via the vim function GuiFont.

Maybe we can default that case to force=true since there no way to avoid it (i.e. same as GuiFont!)

https://github.com/equalsraf/neovim-qt/blob/master/src/gui/shell.cpp#L980

I'm no longer familiar with that code path, but maybe we need to have a look at what we are doing there wrt the option holding a list of multiple fonts.

equalsraf:
:S i was right, the option is sending a list as a string `"Source Code Pro,DejaVu Sans Mono,Courier New,monospace"` and we are not handling those properly anyway. Here is a PoC that makes things slightly better:

```diff
diff --git a/src/gui/shell.cpp b/src/gui/shell.cpp
index f3d3a6e..6a3d52f 100644
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -978,7 +978,15 @@ void Shell::handleSetOption(const QVariantList& opargs)
[...]

PR #1167, opened 2026-07-24
Minor font fixes

equalsraf:
1. Add a quiet flag to suppress showing font related errors - only used for the case when resetting the font on screenChanged
2. do a split on guifont value to try to address the case when guifont is a list of fonts (default neovim) #1149 - it may help for default value but may be confusing because we ignore other fonts
3. avoid setGuiFont on screenChanged
4. make font update less racy - does not fix anything - just makes racy behavior more consistent

For 4. I settled on a logical clock - I also tried to write some vimscript to do the &guifont update in one step but this was only better in some cases. This may break the current behavior around setting fonts - but the current behavior is already racy.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -157,7 +157,7 @@ jobs:
           -DCTEST_EXE_ARGS=${{ matrix.ctest_exe_args }}
 
       - name: Build
-        run: cmake --build ${{ github.workspace }}/build
+        run: cmake --build ${{ github.workspace }}/build -- -v
 
       - name: Test
         uses: GabrielBB/xvfb-action@v1 # Required by Linux, no X11 $DISPLAY
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,7 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 include_directories(${Qt6SvgWidgets_INCLUDE_DIRS})
 add_definitions(${Qt6SvgWidgets_DEFINITIONS})
+add_compile_definitions(QT_TYPESAFE_FLAGS)
 
 set(NEOVIM_QT_SOURCES
   auto/neovimapi0.cpp
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -181,7 +181,7 @@ void Shell::screenChanged()
 {
 	// When the screen changes due to dpi scaling we have to
 	// re-set the current font
-	setGuiFont(fontDesc(), FontOption::Force & FontOption::Quiet, true);
+	setShellFont(font(), FontOption::Force & FontOption::Quiet);
 }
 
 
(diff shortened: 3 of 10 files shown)


equalsraf:
The failing tst_qsettings test looks like a racy call to updateGuiFontRegisters, that results in the font being changed. The Monaco:h20 font setting is from the test, the "SF Mono,Menlo,Monaco,Courier New,monospace" guifont value should be from neovim, and monospace:h11 is probably our default.

The startup update of 'guifont' seems to be racing with the value from the test (and winning). Not sure why there two updates at the start though (but the second is likely neovim setoption, while the first could be screenChanged).

```
# This should be the startup/ctor call - onAttached
8: QWARN  : NeovimQt::TestQSettings::GuiFont() updateGuiFontRegisters "monospace:h11"
# This is probably neovim doing a set option
8: QWARN  : NeovimQt::TestQSettings::GuiFont() setGuiFont "monospace" NeovimQt::Shell::NvimOption false
8: QWARN  : NeovimQt::TestQSettings::GuiFont() updateGuiFontRegisters "monospace:h11"

# Two calls to handleGuiFontOptions result from the calls above - they will call set option guiftont monospace:h11
8: QWARN  : NeovimQt::TestQSettings::GuiFont() handleGuiFontOption "SF Mono,Menlo,Monaco,Courier New,monospace" "monospace:h11"
8: QWARN  : NeovimQt::TestQSettings::GuiFont() handleGuiFontOption "SF Mono,Menlo,Monaco,Courier New,monospace" "monospace:h11"

# This is our test setting font to monaco
8: QWARN  : NeovimQt::TestQSettings::GuiFont() setGuiFont "Monaco:h20" NeovimQt::Shell::RuntimeCall false
8: QWARN  : NeovimQt::TestQSettings::GuiFont() updateGuiFontRegisters "Monaco:h20"

# Again set guifont - a consequence of handleGuiFontOptions earlier
8: QWARN  : NeovimQt::TestQSettings::GuiFont() setGuiFont "monospace:h11" NeovimQt::Shell::NvimOption false
8: QWARN  : NeovimQt::TestQSettings::GuiFont() updateGuiFontRegisters "monospace:h11"

# We should be stable here - the final 2 callbacks - but Monaco lost the race
8: QWARN  : NeovimQt::TestQSettings::GuiFont() handleGuiFontOption "monospace:h11" "monospace:h11"
8: QWARN  : NeovimQt::TestQSettings::GuiFont() handleGuiFontOption "monospace:h11" "monospace:h11"
8: FAIL!  : NeovimQt::TestQSettings::GuiFont() Compared values are not the same
8:    Actual   (w->shell()->fontDesc()): "monospace:h11"
8:    Expected (fontDesc)              : "Monaco:h20"

```

equalsraf:
I skipped to the close event test for windows - it is unrelated to this PR but it is the most common failing test in windows (it passes in debug builds but systematically fails in release builds)

PR #1175, opened 2026-07-26
Build/CI fixes

equalsraf:
Merge windows into the main gh actions workflow and generate release files.

- [x] check cmake warning: Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR)
- [x] test release workflow by creating a tag
- windows tests are disabled - similar flakiness to others - but harder to debug in release build without debug output
- mac os artifact appears to have issues with rpath
- i sneaked in a fix from the fonts PR to make the tests less flaky


equalsraf:
I've left the windows tests disabled for now.

It would be better if the release workflow did not depend on running the build again - but the only way to do this with separate workflows is by adding logic in the job to find and download the artifacts.

PR #1180, opened 2026-07-31
Support font lists in guifont options

### LINKED ISSUE #1168: guifont option does not support font lists
More discussion in https://github.com/equalsraf/neovim-qt/issues/1149 but we are currently assuming the guifont option is a single font description - when in fact it can be a list of fonts separated by commas with comma backslash escaping.

- The correct handling is probably to try one at a time and defer errors to the end.
- This problem also happens with guifontwide.

Guflly:
Parse `guifont` and `guifontwide` as font lists instead of a single description. This tries each entry in order, handles escaped commas, and only reports an error if none work.

When a fallback is selected, the option is updated to the font that was actually applied.

Tests:
- `test_shellwidget`
- `tst_shell fontDescriptionList`

Closes #1168

### DIFF
--- a/src/gui/shell.cpp
+++ b/src/gui/shell.cpp
@@ -119,6 +119,52 @@ void Shell::handleFontError(const QString& msg)
 	}
 }
 
+QStringList Shell::splitFontList(const QString& value) noexcept
+{
+	QStringList fonts;
+	QString font;
+	bool escaped{ false };
+	bool skipSpaces{ false };
+
+	for (const QChar character : value) {
+		if (skipSpaces && character == ' ') {
+			continue;
+		}
+		skipSpaces = false;
+
+		if (escaped) {
+			if (character != ',' && character != '\\') {
+				font.append('\\');
+			}
+			font.append(character);
+			escaped = false;
+		}
+		else if (character == '\\') {
+			escaped = true;
+		}
+		else if (character == ',') {
+			fonts.append(font);
+			font.clear();
+			skipSpaces = true;
+		}
+		else {
+			font.append(character);
+		}
+	}
+
+	if (escaped) {
+		font.append('\\');
+	}
+	fonts.append(font);
+
+	return fonts;
+}
+
+QString Shell::escapeFontListEntry(QString value) noexcept
+{
+	return value.replace('\\', QStringLiteral("\\\\")).replace(',', QStringLiteral("\\,"));
+}
+
 /// Set the GUI font, or display the current font
 ///
 /// @param fdesc Neovim font description string, "Fira Code:h11".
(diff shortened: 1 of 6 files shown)


src/gui/shell.cpp
```diff
@@ -167,7 +214,7 @@ bool Shell::setGuiFont(
 
 	// Only update the ShellWidget when font changes.
 	if (!m_attached) {
-		return false;
+		return true;
```
equalsraf: What is the reason for this swap here?

src/gui/shell.cpp
```diff
@@ -240,16 +292,20 @@ void Shell::updateGuiFontRegisters() noexcept
 	connect(getOption,
 		&MsgpackRequest::finished,
 		this,
-		[this, timestamp](quint32 msg, quint64 _f, const QVariant& val) noexcept
+		[this, timestamp, src](quint32 msg, quint64 _f, const QVariant& val) noexcept
 		{
+			if (src == FontChangeSource::NvimOption) {
+				return;
+			}
```
equalsraf: what is the intent here?

a src NvimOption may still need to trigger the update of the &guifont option, iff they are different. Otherwise &guifont never reflects the current font, but rather the requested font

src/gui/shellwidget/shellwidget.cpp
```diff
@@ -1056,6 +1060,7 @@ QVariant ShellWidget::TryGetQFontFromDescription(const QString& fdesc) const noe
 	}
 
 	QFont font{ attrs.at(0), -1 /*pointSize*/, weight, italic };
+	font.setFamilies({ attrs.at(0) });
```
equalsraf, line 1063: what was happening here? did the constructor not work?

equalsraf:
Hi @Guflly. Thanks so much for the change and the fixes.

It looks good from my end - I did leave some questions in comments - the only blocker I see is the change in updateGuiFontRegisters.

src/gui/shell.cpp
Guflly: Restored the normal `&guifont` update. The selected fallback now replaces the requested list when they differ.

src/gui/shellwidget/shellwidget.cpp
Guflly: The constructor split a family name containing a comma. `setFamilies({attrs.at(0)})` keeps the escaped comma as one family.

src/gui/shell.cpp
Guflly: I was trying to keep unattached updates accepted, but that changed existing behavior. I reverted the swap and restored the original false return.

equalsraf:
Merged. Thanks @Guflly

